PR #72, opened 2026-06-08
fix: invoke sim2real vlm and eval components

timothy-le7:
Routes Sim2Real VLM and held-out evaluation through configured component commands/images, removes quality-derived VLM/eval signal from the genuine path, and asserts component invocation in tests.

### DIFF
--- a/npa/src/npa/cli/workbench/sim2real.py
+++ b/npa/src/npa/cli/workbench/sim2real.py
@@ -130,6 +130,42 @@ def run_command(
     byo_eval_command: str = typer.Option(
         "", "--byo-eval-command", help="BYO eval command."
     ),
+    k8s_namespace: str = typer.Option(
+        "", "--k8s-namespace", help="Namespace for sibling component Jobs."
+    ),
+    k8s_service_account: str = typer.Option(
+        "agent-sa", "--k8s-service-account", help="Service account for sibling Jobs."
+    ),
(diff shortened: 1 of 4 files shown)


timothy-le7:
Cluster validation for commit `8bde62d`.

- Focused workflow tests passed locally; hosted checks are green.
- Full env generation ran at 10,000 records with an 8,000/2,000 split and uploaded to the configured neutral prefix.
- VLM stage ran as a sibling Kubernetes Job with the configured VLM image, `agent-sa`, env secrets, pull-secret plumbing, and a one-GPU sm_120 allocation.
- Held-out eval ran as a sibling Kubernetes Job with the configured eval image, the same service-account/secret plumbing, and a one-GPU sm_120 allocation.
- The trainer consumed parsed VLM Job JSON. Sample score was `0.646824`, not `0.737737`; parsed tags included `collision`, `minor_alignment`, and `unstable`.
- Held-out eval parsed per-env storage output: 16 env records, 7 successes, 16 unique scores, not a 100 percent canned result.
- Run-owned Jobs/pods were cleaned by the workflow; the Sky cluster was explicitly torn down and verified absent.

Merge call: `SEAM`. The remaining seam is component-contract output from the stage images rather than model-native reasoning, while sibling Kubernetes plumbing, parsing, RBAC, tests, and teardown are proven.

PR #84, opened 2026-06-10
fix(sim2real): genuine VLM->RL signal gradient, image-digest provenance, and pinned-SA creds

timothy-le7:
## Summary

Root-cause fixes for the two open sim2real blockers: the **genuine VLM→RL signal FAIL** (degenerate / low-diversity / incoherent / zero-digest-evidence) and the **creds/RBAC "missing context"** failures. All causes are in-repo code, not just config.

### Signal (pipeline steps 8–10)
- **Held-out flat 1.0 → continuous gradient.** `_heldout_env_score` keeps each env's score continuous in its distance/reward; success/failure occupy separate bands. The pass/fail gate still keys off the `success` boolean, so `success_rate` is unchanged — only the degenerate flat-`1.0` wall is removed. (was `score = 1.0 if env_success else …`)
- **Repeated critiques made visible.** When the VLM returns no `per_step`, the broadcast summary critique is tagged `critique_source=summary_broadcast` instead of masquerading as genuine per-step granularity.
- **Anti-hollow signal-diversity gate.** `run_inner_loop` emits a `signal_diversity` diagnostic (distinct scores/rewards, `coherent` flag) into evidence; opt-in `NPA_SIM2REAL_REQUIRE_SIGNAL_DIVERSITY` fails a degenerate batch. These `distinct`/`coherence` metrics were previously only computed by an external validator.

### Provenance — answers "are the genuine images actually invoked?" (digest evidence=0)
- Capture each pod's resolved `image` + `imageID` (repo digest) into `component_invocation.image_digests`, mirroring SONIC's `repo_digests`. The run now emits digest evidence instead of leaving it to be inferred.
- Force `imagePullPolicy=Always` for `-genuine-` tags so a stale cached image under the same tag can't masquerade as the genuine build (override: `NPA_SIM2REAL_IMAGE_PULL_POLICY`).

### Creds / RBAC
- `_terraform_env` refreshes the IAM token by default, clearing any stale ambient `TF_VAR_iam_token`/`NEBIUS_IAM_TOKEN` that shadows the pinned SA and causes `Forbidden`/`401`; reuse is opt-in via `NPA_REUSE_IAM_TOKEN`.
- `npa skypilot verify` gains `--kubeconfig`/`--cluster` and pins `KUBECONFIG` for `sky check`, so verify no longer reports a misleading `403 anonymous` "missing context" against an ambient/empty context.

## Test plan
- [x] New `tests/workflows/test_sim2real_signal_rca.py` (held-out banding, diversity report, pull policy, pod image-digest capture)
- [x] New verify-context tests in `tests/cli/test_skypilot_bootstrap.py`
- [x] New `_terraform_env` token-refresh/reuse tests
- [x] `tests/workflows/test_sim2real_loop.py` (11 passed) + `tests/cli` + guardrails (729 passed, 2 skipped, 1 xpassed)
- [x] `ruff check` clean on all changed files
- [ ] Operator validation: re-run the loop and confirm `signal_diversity.coherent=true`, non-flat held-out scores, and non-empty `image_digests` in artifacts

### DIFF
--- a/npa/src/npa/cli/cluster/terraform_lifecycle.py
+++ b/npa/src/npa/cli/cluster/terraform_lifecycle.py
@@ -190,10 +190,25 @@ def _require_bin(binary: str) -> str:
 
 def _terraform_env(nebius_bin: str) -> dict[str, str]:
     env = os.environ.copy()
-    if not env.get("TF_VAR_iam_token"):
-        token = _run_capture([nebius_bin, "iam", "get-access-token"], env=env).stdout.strip()
-        env["TF_VAR_iam_token"] = token
-        env["NEBIUS_IAM_TOKEN"] = token
+    # A stale ambient IAM token (e.g. a cloud-env token) silently shadows the
+    # intended Nebius profile and mints kubeconfig/registry credentials for the
+    # wrong principal -- the cause of Forbidden jobs/pods/nodes and 401 image
+    # pulls. Mint a fresh token by default after clearing any stale token; opt
+    # back into reuse only when NPA_REUSE_IAM_TOKEN is explicitly set (e.g. CI
+    # that injects a short-lived token intentionally).
+    reuse = env.get("NPA_REUSE_IAM_TOKEN", "").strip().lower() in {
+        "1",
+        "true",
+        "yes",
+        "on",
+    }
+    if reuse and env.get("TF_VAR_iam_token"):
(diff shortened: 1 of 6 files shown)


timothy-le7:
Superseded by the consolidated sim2real workflow PR #104.

PR #122, opened 2026-06-17
Normalize Token Factory reasoning-model output for Workbench artifacts

stewtong:
## Summary

Token Factory reasoning-model responses can surface thinking text in the same field that Workbench treats as the visible answer. For `nvidia/Cosmos3-Super-Reasoner`, that can show up as a leading `<think>...</think>` block in `choices[0].message.content`; for other reasoning models, visible content can be null while reasoning appears in `reasoning` or `reasoning_content`.

This PR normalizes those response shapes before Workbench writes artifacts, so `npa workbench token-factory reason` stores clean visible output in `scene_reasoning.json` `analysis` and does not turn reasoning-only responses into the literal string `"None"`.

## What changed

- Added a small response normalizer that separates visible text from optional reasoning text in `choices[0].message`.
- Strips leading inline `<think>...</think>` blocks from visible output.
- Supports separate `reasoning` and `reasoning_content` response fields.
- Treats truncated think-only output as no visible answer.
- Makes `chat_completion_text()` raise a clear error for reasoning-only output instead of returning the literal string `"None"`.
- Keeps `scene_reasoning.json` clean by default: `analysis` contains visible output only and no default `reasoning` field is persisted.
- Updates tests for inline think blocks, null content plus reasoning, truncated think-only output, plain visible output, and CLI-written `scene_reasoning.json`.

## Why

Workbench artifacts are often parsed by downstream eval helpers. Keeping `analysis` to the visible answer makes those artifacts easier to trust and less model-specific.

This keeps the default artifact contract simple:

- visible answer in `analysis`
- no thinking trace persisted unless a future explicit opt-in contract is added

## Verification

- `../npa-venv/bin/python -m pytest npa/tests/clients/test_token_factory.py npa/tests/workbench/test_token_factory_tool.py npa/tests/cli/test_workbench_token_factory_cli.py -q`
  - `35 passed`
- `../npa-venv/bin/python -m py_compile npa/src/npa/clients/token_factory.py npa/tests/clients/test_token_factory.py npa/tests/workbench/test_token_factory_tool.py npa/tests/cli/test_workbench_token_factory_cli.py`

## Notes

This is intentionally scoped to Workbench artifact behavior. Token Factory service-side handling for model-specific `thinking=false` or chat-template behavior can be handled separately if needed.


timothy-le7:
Superseded by #127 — branch pushed to `nebius/nebius-physical-ai` so confidentiality scan can access org secrets (fork PRs cannot).

PR #244, opened 2026-08-02
Finish the SkyPilot retirement's tail: dangling pointers, stale docs, and a runner one deletion from passing vacuously

timothy-le7:
## Summary

#233 retired thirty SkyPilot templates. This is the tail it left: the pointers to them, the
docs that still teach the surface they were, and a validation script that would have gone
silently vacuous the moment the directory is deleted.

The first part was pushed to the #233 branch minutes after it was squash-merged and missed the
merge. The rest is the plan's last unblocked item — the documentation rewrites — which are
independent of the two templates still in the directory (those belong to #234/#235 and are
Codex's to finish).

## 1. Pointers to files that no longer exist

Eleven files on `main` right now name something #233 deleted. Three kinds, in increasing order
of how badly they fail someone:

**Dead links.** Four `cosmos3-*` skills send readers to
`skypilot/cosmos3-text-to-image-inference.yaml`, retired in #233.

**Dangling metadata.** Nine specs carry a `skypilotTwin` naming a deleted file. One uses the
**plural** `skypilotTwins:` list — which is exactly why it survived my first pass: a
singular-only regex walks straight past it.

**A self-reference, and a snippet broken more quietly.** Two specs declare *themselves* as their
own `skypilotTwin` — the fingerprint of my own bulk path-repoint in #233 rewriting the field's
*value* along with every other mention of the old path. And
`skills/atomic/cosmos3-env-troubleshoot/SKILL.md` ships a runnable snippet that opens the path
and reads `doc["envs"]`, a key an npa.workflow spec does not have. Repointing the path alone
would have left it broken in a way that is harder to notice than a missing file, so the snippet
body is fixed to the spec's shape.

**Guardrail:** `test_no_dangling_workflow_references.py` — a shipped file may not name a workflow
path that does not exist, across `skills/`, `docs/`, `scripts/` and the specs, in Markdown, YAML,
Python and shell. Python and shell are included precisely because a broken path inside a runnable
snippet is the expensive kind. `EVIDENCE.md`, `CHANGELOG.md` and `DESIGN.md` are exempt:
describing what is gone is their job.

One false positive earned its keep — the check first compared *basenames* and flagged
`cosmos3-generate.yaml`, whose spec correctly names the surviving template of the same name in a
different directory. A spec and the template it replaced usually **share** a filename; only the
full path distinguishes a real self-reference.

## 2. Docs still teaching the retired surface

`docs/workbench-yaml-guide.md` opened with "A Workbench pipeline is a SkyPilot multi-document
YAML file", described `---` task documents with `envs:`/`run:`/`curl`/`jq`, and named
`npa-workflows/bdd100k-pipeline.yaml` as its reference throughout — a file that is a spec with
`states:` and `toolRef:`. Anyone following the guide opened the reference and found something
entirely different.

Its worst section is the one a new contributor reaches last. **"Adding a New Pipeline"** began:
*"Start from a multi-document SkyPilot YAML file."* That is now the one thing a guardrail
rejects.

Rewritten around the spec — `config`/`resources`/`states`, declared fan-out, `num_nodes`, and the
real thirteen state names, which is what `workflow status` reports. Everything still true is
kept: service endpoints, S3 conventions, durable state, and the label-map rationale, which now
also records *why* eval needed its own flag — BDD100K has a category literally called `train`,
the vehicle, so `int('train')` is what a missing label map looks like at runtime.

Two skills went with it:

- `skypilot-workflows` is reframed from authoring to what it now is: how the engine renders and
  submits — JobGroups, task-level `num_nodes`, per-toolRef setup, run preambles, the Isaac EULA
  gate — plus the traps that cost live jobs.
- `workbench-reference-workflows` had a contradiction worth naming: its Three-Tier Contract still
  read *"SkyPilot YAML is the executable source of truth"*, which the retirement directly refutes.
  That tier is the spec plus its toolRef argv now. It also carried an orphaned list fragment from
  one of my own earlier edits.

## 3. A runner one deletion from passing vacuously

The three tmux runners parsed every file in the retiring directory and asserted
`docs[0]["name"]`. With the catalog nearly empty that was already close to meaningless — and the
day the directory goes, `root.glob("*.yaml")` returns nothing, the loop never executes, and the
script prints `skypilot parse OK (0 files)` and **passes**.

They now walk the 54 specs and refuse to run against a corpus smaller than twenty. Verified both
directions: `spec parse OK (54 files)` against the repo, and exit 1 with `found 0 files` against
an empty tree.

## Two things caught by running rather than reading

- That new check **failed on its first execution**: `sim2real-gpu-cross-region-agent.yaml` ships
  `v0.0.1-beta` on purpose and I had written an equality check on `apiVersion`. Relaxed to a
  prefix, which is what was actually meant.
- `Not for authoring: pipelines are…` in a `SKILL.md` description made its YAML front matter
  unparseable. `test_skills_index` said so in seconds. Swept all 55 skills; only that one.

## Testing

- Full offline suite: **5923 passed**, 46 skipped. The single failure,
  `test_golden_eval_tmux`, is pre-existing and reproduces on the merge base.
- `ruff` clean; all 952 guardrails pass.
- Dangling scan returns **NONE** across 360 candidate files.
- Every CLI verb and flag added to the docs was checked against `--help` on a live checkout
  rather than assumed.
- The three runner blocks were executed, not just edited.

## Not in this PR

Deleting `npa/src/npa/workflows/skypilot/` and inverting the guardrail to "this directory must
not exist". It is gated on the two templates #234 and #235 added mid-sweep, and is being handled
separately.

<div><a href="https://cursor.com/agents/bc-bd799333-86e2-48ab-a163-7d9f067197fd?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-bd799333-86e2-48ab-a163-7d9f067197fd&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


timothy-le7:
Superseded by #251, rebuilt directly from `main` after #245 merged.

The successor carries only the remaining SkyPilot-retirement tail, adds the missing spec/matrix/contract guardrails, and includes offline plus live RTX PRO 6000 verification. Keeping this PR closed avoids reviving its pre-#245 branch history.

PR #254, opened 2026-08-03
Harden Cosmos2 Transfer review paths

timothy-le7:
## Summary

- Require hashes when installing the exact Cosmos2 security override URLs, with focused static coverage that binds the flag to that install command.
- Document that the real transfer workflow requires an input video and fails before inference for a missing, empty, or video-free input prefix.
- Document the deliberate global Python wrapper used by SkyPilot while retaining the absolute system interpreter path.
- Add a CLI-level regression test for the missing conditioned-input video error and prove inference is not called.

## Why

This is a stacked follow-up to #252 implementing its four actionable review improvements. It tightens dependency integrity and makes the already-live input and interpreter contracts explicit without redesigning the image or workflow.

## User and developer impact

Operators get a clear mandatory-input contract and a user-visible nonzero CLI failure before GPU inference. Future security overrides without hashes fail during image build. Maintainers get regression coverage for both the Docker install flags and the CLI failure path.

## Verification

- [x] Focused Cosmos2 input and image tests: 22 passed.
- [x] Licensing, packaging, real-component, toolRef, skill-index, and workflow guardrails: 370 passed.
- [x] Targeted Ruff passed.
- [x] Cosmos2 workflow validate-spec and plan-spec passed; the plan retains --condition-on-input and --execute.
- [x] Full non-e2e suite: 5617 passed, 47 skipped, 1 xpassed.
- [x] Public-repo hygiene and changed-file scope checked.

## Limitations

Per the narrow review scope, this PR does not rebuild or republish the container, change the image tag or digests, reclassify licensing, alter model or media boundaries, or dispatch public image publication. No live GPU inference was repeated; #252 already contains the live qualification for the preserved runtime behavior.

## Skill maintenance

- [x] No root skill or skill index update is needed because this follow-up documents and tests behavior already introduced by #252.

### DIFF
--- a/npa/docker/workbench/cosmos2-transfer/Dockerfile
+++ b/npa/docker/workbench/cosmos2-transfer/Dockerfile
@@ -78,7 +78,7 @@ RUN set -eux; \
     printf '%s\n' "${COSMOS_PYTHON_VERSION}" > .python-version; \
     uv sync --locked --no-dev --no-editable --extra=cu128 \
       --python "${COSMOS_PYTHON_VERSION}"; \
-    uv pip install --python .venv/bin/python --no-deps \
+    uv pip install --python .venv/bin/python --no-deps --require-hashes \
       --requirement /tmp/cosmos2-security-overrides.txt; \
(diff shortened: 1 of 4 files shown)


timothy-le7:
Consolidated exact commit c76d58279d772556fd38b4f37585a8b7b7daba1f into draft PR #252 after verifying its remote head. This temporary stacked PR is now closed so all Cosmos Transfer 2.5 work remains in #252.

PR #251, opened 2026-08-03
Finish the SkyPilot retirement tail

timothy-le7:
## Summary

- finish the SkyPilot retirement tail around the real Cosmos2 producer/consumer contract while preserving distinct real `manifest.json` and reference `index.json` schemas
- keep the staged Sim2Real engine best-effort at its documented boundary: only inference/runtime and `FrameExtractionError` failures degrade to the descriptor manifest; unrelated programming errors propagate
- keep executable Cosmos publication fail-closed: conditioned input is mandatory, required zero-frame decodes publish nothing, and extraction process/OS failures have a typed low-level error
- require envgen transfer manifests to prove `schema=npa.cosmos2.transfer.v1`, `mode=cosmos_transfer2.5_gpu`, and `status=executed`, then consume only the exact non-empty `frames[].uri` list
- support `local://` and `file://` manifest inputs consistently, resolve the public envgen CLI reference once, serialize local/S3 manifests identically with a trailing newline, and make declared manifest URIs independent of a trailing slash on `augment_uri`
- preserve the five historical job-339 output guards, one-to-one shipped workflow matrix, dangling-reference protection, public Typer routing, deterministic generated docs, and the conspicuous plan-only fan-out clobber warning

## Review resolutions

1. **Staged-engine fallback:** `_run_real_cosmos_transfer` catches only the typed extraction failure at the staged boundary, emits a specific fallback event, and returns `None` so the caller produces the historical descriptor manifest. Clean zero-frame decode also falls back; unrelated exceptions are not swallowed. The producer used by executable job-339 paths remains fail-closed.
2. **Producer safety tests:** direct subprocess-backed tests cover decode exit failure, OS start failure, successful zero-frame decode, zero-frame publication refusal before any upload, staged fallback, and unrelated-error propagation.
3. **Reference provenance:** reference augmentation retains `index.json` / `npa.sim2real.augmented_frames.v1`; its transfer-shaped compatibility manifest is explicitly `mode=reference_augment`, `status=executed_reference`. Envgen rejects it as real transfer data.
4. **Local URI parity:** transfer manifests load through plain paths, `file://`, and `local://`; unreadable and malformed documents fail with `Sim2RealEnvGenError`.
5. **Consistency:** local reference frame URIs retain their scheme, `frames[].uri` aligns with `index_uri`, public envgen no longer double-derives the scene, local and S3 manifest bytes match, and every affected spec/generated template declares an explicit normalized `augment_manifest_uri` guarded against a slashless prefix.
[...]

PR #257, opened 2026-08-03
Fix Cosmos2 workflow artifact contracts

timothy-le7:
## Summary

- restore the complete Cosmos2 transfer publication contract discovered by live job 339: producer-owned canonical `manifest.json` URI/schema helpers, exact frame publication, fail-closed transfer behavior, and downstream envgen validation
- correct the five historical declarations in `cosmos-synth-fanout-curation.yaml` (`synth-shard-a` and `synth-shard-b`), `cosmos2-transfer.yaml` (`transfer`), `sim2real-vlm-rl.yaml` (`augment`), and `tokenfactory-cosmos-gate.yaml` (`augment-scene`), with a guard that rejects future `index.json`/schema mismatches
- keep the fan-out workflow plan-only and preserve its conspicuous defect note: both apparent shards still share one output URI, so the second overwrites the first until the contract gains a per-shard configuration token
- make `npa.workflow/v0.0.1` the only documented authoring surface; SkyPilot remains the execution engine, and guardrails reject dangling workflow pointers without recreating the retired SkyPilot authoring directory
- retain current main's existing Cosmos live-matrix case and add the other 13 still-needed cases from the 14-case extraction, including safe plan-only coverage where execution is intentionally unavailable
- reconcile generated CLI documentation and its collision/drift guardrails with current main

## Why / impact

The old declarations claimed `manifest.json` while Cosmos2 actually published an incompatible `index.json`. This could make a workflow plan successfully and then fail downstream—or consume the wrong schema—after spending GPU time. The producer, catalog, shipped specifications, and regression tests now share one explicit artifact contract. Authors are directed to the declarative workflow API rather than stale multi-document SkyPilot YAML.

## Local validation

- `npa/.venv/bin/python -m pytest tests/ --ignore=tests/e2e --timeout=120 -q`: **6543 passed, 47 skipped, 1 xpassed**
- focused post-rebase suite: **167 passed**
- all **54** current `npa.workflow/v0.0.1` specifications validated and planned successfully
- Ruff passed for changed Python/tooling paths
- mypy passed for all 10 changed source files (the repository-wide advisory run retains unrelated baseline findings)
- generated documentation drift check passed
- `git diff --check origin/main...HEAD` passed
- gitleaks directory scan and `origin/main..HEAD` history scan passed with no findings and no scanner-rule weakening or new allowlist

## Real GPU validation

Both final runs used an RTX PRO 6000 Blackwell Server Edition and a freshly staged durable source overlay from executable commit `2ffbee656930f9b91a9ad660be886ae38ceff899`. Commit `b278cf33e214e25509fc2bab8984d63bc4761fde` records the resulting evidence; later head `be9248fa89964bb1503695ab3d6e0905616dfbb2` only changes the source-literal representation of the byte-identical live input fixture for confidentiality-scan compatibility.

- job **375**, run `npa-wf-gpu-cosmos2-transfer-b7818c52`: **SUCCEEDED**; exact `gpu:cosmos2-transfer.yaml` case passed in 963.38s. It produced the canonical transfer schema, executed/GPU/conditioned status, eight distinct existing frames, and a non-empty generated video.
- job group **377**, run `npa-wf-multi-sim2real-two-step-b9d5e9dc`: **SUCCEEDED**; exact `multi:sim2real-two-step.yaml` case passed in 1181.79s with zero recoveries. Transfer produced four distinct existing frames, and envgen produced 1,000 canonical rows whose augmented-frame URI set exactly matched those four manifest frames.

The image was the legally redistributable procedural-fixture build merged by #252. The source overlay was re-staged before each run. Managed job resources were cleaned up; no job pod remained. `npa-lancedb` was left deployed as the required shared rotation dependency, and this run created no detection-training resource.

## Limitation

`cosmos-synth-fanout-curation.yaml` remains deliberately plan-only. Its two apparent shards pass the same output URI and therefore yield only one surviving shard; a real fix requires a per-shard configuration token. The underlying Cosmos2 GPU path is covered by jobs 375 and 377 above.


timothy-le7:
Superseded by #251, which now carries the same validated commit and is the canonical merge target.

PR #264, opened 2026-08-07
Add Cosmos 3 access preflight guide and xet/401-403 diagnostics

stewtong:
### What breaks today

Two failure modes hit anyone clearing gated Hugging Face access for Cosmos 3,
and neither is documented anywhere in this repo.

**Two separately-gated guardrail repos, split by serving path.** This repo's
`npa workbench cosmos3 generate` path pulls `nvidia/Cosmos-Guardrail1`.
Serving the same model through vLLM-Omni (outside npa, for example the
`vllm/vllm-omni:cosmos3` container) pulls a different gated repo,
`nvidia/Cosmos-1.0-Guardrail`. Accepting one license does not accept the
other, so an operator who cleared one path still fails the other with no
indication why. Measured on the vLLM-Omni path before the license was
accepted:

    GatedRepoError: 401 Client Error
    Cannot access gated repo for url
    https://huggingface.co/nvidia/Cosmos-1.0-Guardrail/resolve/.../blocklist/custom/branding
    Access to model nvidia/Cosmos-1.0-Guardrail is restricted.
    RuntimeError: Orchestrator initialization failed: 401 Client Error.

**A 401-vs-403 split that identifies the actual problem.** Before that
license was accepted, an authenticated request to the same URL returned 403
while an anonymous request returned 401. That split is what showed the
problem was a missing license acceptance, not a bad token; accepting the
license resolved it immediately. Nothing in the repo's error text or docs
mentions this diagnostic.

**An undocumented xet download failure.** With a valid token and an accepted
license, download can still fail inside `huggingface_hub`'s Xet transfer
client:

    huggingface_hub/file_download.py:1920 _download_to_tmp_and_move
      -> xet_get(...)
        -> huggingface_hub/file_download.py:563  session.new_file_download_group(...)
    RuntimeError: Task error: Unable to parse string as hex hash value

This is huggingface/xet-core#895, closed 2026-07-28, which reproduces on the
exact pin pair `hf-xet 1.5.1` plus `huggingface_hub 1.23.0`. `HF_HUB_DISABLE_XET`
appears nowhere in this repo today. Measured workaround: setting
`HF_HUB_DISABLE_XET=1` downloaded the affected 146-file, 17 GB guardrail repo
in 1m52s and the run proceeded normally.

### What this change does

A new doc, `docs/workbench/cosmos3-access-preflight.md`, walks through Hugging
Face account and token setup, names both gated guardrail repos and which
serving path needs which, gives the 401-vs-403 diagnostic table, and documents
the xet failure signature and workaround. `docs/workbench/cosmos3-generate.md`
gets a short pointer to it from the credential section and a new row in its
troubleshooting table for the xet symptom.

On the code side, `require_model_access`'s missing-token error now also names
the 401-vs-403 split and points at the new doc, for the case where a token is
already set but a download still fails. A new `check_xet_pin` function probes
the installed `huggingface_hub` and `hf-xet` versions in the generation
runtime and prints a warning naming the `HF_HUB_DISABLE_XET=1` workaround only
when it detects the exact affected pair; it never sets the variable itself,
since newer releases of both packages fix the issue and disabling Xet
unconditionally would cost unaffected environments a faster download path for
no reason. The probe fails open: any error reading the versions (missing
interpreter, package not installed, non-zero exit) returns no warning rather
than blocking the run.

### What this deliberately does not do

No new tool, no new CLI command, and no change to `GUARDRAIL_MODEL_ID`; it
stays `nvidia/Cosmos-Guardrail1`, which is correct for the serving path this
repo ships. This is a docs-plus-small-code change on an existing preflight,
not a new access-checking surface.

### Evidence

- 401/403 split, the xet signature, and the `HF_HUB_DISABLE_XET=1` timing:
  measured during Cosmos3-Super bring-up on 8x H200 with vLLM-Omni, logged
  live as it happened.
- Guardrail repo split confirmed against both Hugging Face repos directly.

### Tests

`npa/tests/workbench/test_cosmos3_generate.py` adds coverage for the extended
error text (401/403 hint, doc pointer), `check_xet_pin` on the affected pair,
an unaffected pair, and a failed probe, and the warning path in
`run_cosmos3_generate`. Full existing suite for this module (34 tests)
continues to pass, and the repo's own `ruff check` gate is clean on every
touched file.

A follow-up guide for self-hosted Cosmos3-Super serving on Nebius hardware is
in progress separately.

### DIFF
--- a/docs/workbench/cosmos3-access-preflight.md
+++ b/docs/workbench/cosmos3-access-preflight.md
@@ -1,0 +1,125 @@
+# Cosmos 3 access preflight: accounts, tokens, and gated-repo diagnostics
+
+Clearing gated Hugging Face access is the most common reason a Cosmos 3
+generation run fails before it produces anything. This page is the full
+checklist: create the account, scope the token, accept the right license for
+the serving path you are using, and read the two failure signatures that look
+like credential problems but are not the same one.
+
+See [`cosmos3-generate.md`](cosmos3-generate.md) for the credential table this
+extends and for `--dry-run`, which resolves everything below except the
+network calls themselves, so you can confirm mode and checkpoint on a laptop
+before spending GPU time.
+
+## Hugging Face account and token
+
+1. Create a Hugging Face account at <https://huggingface.co/join> if you do
+   not already have one.
+2. Create a token at <https://huggingface.co/settings/tokens>. A read-scoped
+   token is enough; generation only downloads weights, it never pushes.
+3. Export it as `HF_TOKEN` (or set `NPA_COSMOS3_HF_TOKEN_ENV` to the name of
+   an env var you already use for a different token).
+4. Accept the license for every gated repo your run touches before the run,
+   not during it. `require_model_access` only checks that a token is present;
+   it cannot check that the token's account has accepted a given repo's
+   license, so an unaccepted license surfaces later, mid-download, as the 403
+   case in the diagnostic below.
+
+## Two gated guardrail repos, split by serving path
+
+Cosmos 3's guardrail models are gated separately from the base checkpoint,
+and **which guardrail repo you need depends on how you are serving the
+model**, not on which Cosmos 3 checkpoint you picked. Accepting one license
+does not accept the other.
+
+| Serving path | Guardrail repo | Accept the license at |
+| --- | --- | --- |
+| `npa workbench cosmos3 generate` (this repo, cosmos-framework) | `nvidia/Cosmos-Guardrail1` | <https://huggingface.co/nvidia/Cosmos-Guardrail1> |
+| vLLM-Omni serving (e.g. the `vllm/vllm-omni:cosmos3` container, outside npa) | `nvidia/Cosmos-1.0-Guardrail` | <https://huggingface.co/nvidia/Cosmos-1.0-Guardrail> |
+
+An operator who has cleared the vLLM-Omni guardrail repo for a serving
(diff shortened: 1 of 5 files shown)


stewtong:
Pushed a follow-up commit (`beb038e7`) hardening this after an offline review pass from @timothy-le7:

- Closed a fail-open hole: probe stdout that is valid JSON but not an object now returns `{}` instead of raising through `check_xet_pin`. The docstring promises this check can never fail a run; that now holds for every failure mode (non-zero exit, timeout, malformed JSON, non-dict JSON), and all four are tested.
- The xet warning now fires whenever the run will download from Hugging Face (`access["hf_auth"] == "configured"`), not only with guardrails on. xet-core#895 is a transfer bug, so a guardrails-off run that still pulls its checkpoint from HF hits the identical failure.
- The missing-token error now speaks to the case it actually fires on: with no token set, a gated fetch is anonymous and returns 401; a 403 after setting a token means the account still has to accept the license.
- Test hardening: the version probe now runs against a real interpreter in one test, another asserts which interpreter and package names get probed, and the warning path is exercised through a threaded runner seam (`run_cosmos3_generate(version_probe_runner=...)`) rather than a monkeypatched `check_xet_pin`. Test file goes 34 to 42.
- Docs: the preflight doc's example traceback now names the container's tokenless state (it showed a 401 in the section explaining 403), and the xet signature is added to `known-footguns.md`.

All nine checks green on the new head.

PR #268, opened 2026-08-08
Add a Cosmos3-Super serving container (npa-cosmos3-serving)

stewtong:
Adds a container that serves `nvidia/Cosmos3-Super` as an OpenAI-compatible endpoint on one 8-GPU
node, wrapping NVIDIA's `vllm/vllm-omni` runtime at a pinned digest.

This is the container slice only, built at Timothy's request off the 8x H200 measurement work
behind vllm-project/vllm-omni#5909. The workbench tool surface around it (a CLI verb, the SDK, the
workflow toolRef, tool registration) is deliberately not here; the open questions below are what I
would want settled before building it.

## The gap

npa reaches Cosmos 3 in two shapes today. `cosmos serve` runs the reasoner tower on vLLM, and
`cosmos3 generate` runs single-job batch generation through cosmos-framework in the `npa-cosmos3`
image. Neither loads the generator once and serves many requests against resident weights.

The workload that wants this is synthetic-data production. A batch job pays model load on every
invocation, and for the 64B Super checkpoint that cost is large: on 8x H200, HSDP-sharded configs
reach ready in roughly 280 to 290 s warm, and a cold page cache adds about 320 s on top. This
container pays it once.

## What is in the change

- `npa/docker/workbench/cosmos3-serving/Dockerfile`, from `vllm/vllm-omni` pinned by digest. No
  weights in the image, the same rule `npa-cosmos3` follows: the checkpoint and the gated guardrail
  models download at run time under the operator's own license acceptance, and the build fails if a
  weight file lands in a layer. The image drops to a non-root runtime user.
- `entrypoint.sh`, which preflights three failure modes and then execs the serve command.
- `verify_env.py`, a build gate that runs with no GPU.
- `npa/tests/docker/test_cosmos3_serving_image_contract.py`, 17 tests.
- `docs/workbench/cosmos3-super-serving.md`, plus an index row.

### The three preflight checks

Each one is a way this container has actually failed to start, and each costs minutes of 8 GPUs when
it surfaces as a stack trace instead of a message.

1. **Gated guardrail access.** The serving path pulls `nvidia/Cosmos-1.0-Guardrail`, which is gated
   separately from the batch path's `nvidia/Cosmos-Guardrail1`; accepting one license does not accept
   the other. With guardrails on and no token, the fetch goes out anonymous and dies with a 401
   several minutes into startup, which reads like a bad token rather than a missing one. The
   entrypoint refuses first and carries the 401 versus 403 split in the message.
2. **Cache writability.** The image runs as uid 1000, so a cache mounted from a host directory owned
   by someone else is readable but not writable, and the failure otherwise surfaces as a lock error
   partway through fetching 17 GB of guardrail weights.
3. **GPU count.** The pinned parallel config is an 8-GPU decomposition, and a different count fails
   inside distributed init with a shape error.

### On `HF_HUB_DISABLE_XET`

The image sets it. That is a departure from `npa workbench cosmos3 generate`, which only warns on
the same condition, and the difference is ownership of the pins. That path runs in whatever
environment the operator built, so setting the variable on their behalf would silently disable a
faster download path for environments that do not need it. This Dockerfile chose its base image,
and that base pins `hf-xet` 1.5.1 with `huggingface_hub` 1.23.0, the exact pair
huggingface/xet-core#895 reproduces on. `verify_env.py` re-reads the installed pair at build time
and fails the build once a base bump moves off it, naming the line to delete, so the workaround
cannot outlive the defect.

### On the tests

The behavioral half of the test file runs the real entrypoint with a recording `vllm` and a
configurable `nvidia-smi` on PATH, and reads the argv the script actually built. Asserting against a
copy of the expected command would pass while the script produced something else. The same applies
to the build gate: it executes the entrypoint in dry-run mode rather than restating its expected
output.

## Validation

Built and run on 8x H200 SXM (143,771 MiB each, driver 580.126.09) against base digest
`sha256:6d2630c7d637b699557573f2c3fee8df5d4d0cd718977aa22549ed6a6ef30587`, guardrails on.

| Stage | Result |
| --- | --- |
| Build, no GPU | Succeeded; the build gate passed all four checks inside the image |
| Startup to `Application startup complete` | 592 s, cold page cache |
| One t2v request, 1280x720, 189 frames, 24 fps, 35 steps, seed 17 | HTTP 200, server-side `stage_gen_time_ms` 138,721 |
| Clip | h264 1280x720 yuv420p, 24 fps, 189 frames read, 7.875 s, full decode clean, luma 11 to 238 with mean near 121 across sampled frames, so not blank |
| Memory while serving | 43.8 to 44.0 GiB per GPU on ranks 1 to 7, 50.3 GiB on rank 0, which also hosts the API server and the guardrail models |

The generation figure sits inside the band already measured for this configuration and prompt on the
same hardware without the wrapper, so the wrapper adds no measurable cost.

All three preflight refusals were exercised against real conditions rather than simulated ones,
including a host cache that genuinely belonged to a different uid than the container's runtime user.
A fourth case fired unplanned during the run: an empty token string, caught by the same check as the
first, before a boot that would have died minutes later.

The 592 s cold boot is worth one note on its own. It is the reason the image's `HEALTHCHECK` carries
a 20 minute `start-period`. A probe tuned even for the warm figure would have failed that boot and
restarted it into another one.

## Open questions, and I am happy to take direction on all of them

**Scope: single node.** This assumes 8 GPUs on one node and does not design for multi-node. Say if
multi-node has to be in the shape from the start.

**Config surface.** The parallel config is pinned rather than exposed:
`--cfg-parallel-size 2 --ulysses-degree 4 --use-hsdp --hsdp-shard-size 8`. That is what the model
card recommends for 8x H200, H100, or A100, and it was the fastest of five strategies measured on
8x H200, about 9% faster than plain tensor parallelism at both 35 and 50 steps. An env var appends
extra flags as an escape hatch. The alternative is exposing the parallel flags directly and
inheriting their support burden. Generation parameters are not in this surface at all, since steps,
frames, resolution, and per-request guardrail posture are request fields on `/v1/videos` rather than
server flags.

**Guardrail default.** On, matching the model card. For a shared server the per-request opt-out is a
policy question rather than a convenience one, and guardrails alter the output bytes rather than only
gating requests, so a clip generated with them on is not the same clip generated with them off.

**Naming.** `npa-cosmos3-serving` avoids colliding with the reasoner's `cosmos serve`, but if a tool
verb lands later, `cosmos3 serve` sits awkwardly next to it. Happy to rename.

**Tool registration, deliberately not done here.** The image is not in the packaging contract, the
datacenter Blackwell verdict manifest, `CONTAINER_IMAGE_NAMES`, or the golden-eval manifest, and the
docs page says so. Those four are one chain rather than four edits: a contract entry obliges a
Blackwell verdict, which obliges a `CONTAINER_IMAGE_NAMES` key and a supported-tools version pin,
which obliges a golden-eval entry backed by a real capability smoke that needs an 8-GPU node to run.
That seemed like your call on version scheme and smoke tier rather than mine to assume. I am glad to
do it in this PR if you want it here.

**Who builds the tool surface.** The measured groundwork is done and public in #5909, and the access
path is in #264. The remaining work is npa integration, which touches your architecture more than it
touches Cosmos. I can build it behind whatever direction comes out of this review, or hand over the
evidence and fixtures if you would rather own it.

## One dependency

The docs page links `docs/workbench/cosmos3-access-preflight.md`, which is added by #264 and is not
on main yet. If #264 lands first the link resolves; if this lands first, that link is dead until it
does. Happy to reorder, or to drop the link and inline the two paragraphs it points at.

Full validation logs, the generated clip, and the preflight transcripts are available on request.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,6 +12,7 @@ Nebius Physical AI.
 | [workbench/](workbench/) | Workbench solution docs, including getting started, cookbooks, and troubleshooting |
 | [workbench/kubernetes.md](workbench/kubernetes.md) | User setup and operational guide for running Workbench on managed Kubernetes |
 | [workbench/cosmos3-generate.md](workbench/cosmos3-generate.md) | Cosmos 3 generation (`npa-cosmos3`) — build, run via CLI/SDK/workflow, and the runtime-credential posture that keeps weights out of the image |
+| [workbench/cosmos3-super-serving.md](workbench/cosmos3-super-serving.md) | Cosmos3-Super serving (`npa-cosmos3-serving`), an 8-GPU single-node endpoint: build, run, readiness window, and guardrail posture |
 | [../npa/workflows/workbench/npa-workflows/README.md](../npa/workflows/workbench/npa-workflows/README.md) | **Workflow catalog** — find the right `npa.workflow` spec by what you want to do |
 | [architecture/solutions-model.md](architecture/solutions-model.md) | Platform model for adding and maintaining solutions |
 | [architecture/cli-namespaces.md](architecture/cli-namespaces.md) | CLI namespace conventions |
--- a/docs/workbench/cosmos3-super-serving.md
+++ b/docs/workbench/cosmos3-super-serving.md
@@ -1,0 +1,251 @@
+# Cosmos3-Super serving on the workbench (`npa-cosmos3-serving`)
+
+`npa workbench cosmos3 generate` runs Cosmos 3 generation as a batch job: one
+invocation, one artifact, and a full model load every time. For the 64B
+`Cosmos3-Super` checkpoint that load costs minutes, so a synthetic-data
+workload that wants many clips pays it over and over.
+
+This image is the serving half. It loads `nvidia/Cosmos3-Super` once and serves
+an OpenAI-compatible endpoint against resident weights, on one 8-GPU node.
+
+| Piece | Path |
+| --- | --- |
+| Image | `npa/docker/workbench/cosmos3-serving/Dockerfile` (`npa-cosmos3-serving`) |
+| Entrypoint | `npa/docker/workbench/cosmos3-serving/entrypoint.sh` |
+| Build gate | `npa/docker/workbench/cosmos3-serving/verify_env.py` |
+| Build script | `npa/docker/workbench/cosmos3-serving/build.sh` |
+| Image contract tests | `npa/tests/docker/test_cosmos3_serving_image_contract.py` |
+
+The image wraps NVIDIA's own `vllm/vllm-omni` serving runtime, pinned by digest,
+and adds a preflight entrypoint and a build gate.
+
+## What this is not, yet
+
+The image is **not registered as a deployable workbench tool**. There is no
+`npa workbench cosmos3 serve` CLI verb, no SDK surface, no workflow integration,
+and no entry in the packaging contract, the datacenter Blackwell verdict
+manifest, `CONTAINER_IMAGE_NAMES`, or the golden-eval manifest.
+
+That is a deliberate stopping point rather than an omission, because those
+registrations are one chain rather than four independent edits: a packaging
+contract entry obliges a Blackwell verdict, which obliges a
+`CONTAINER_IMAGE_NAMES` key and a supported-tools version pin, which obliges a
+golden-eval manifest entry backed by a real capability smoke that needs an
+8-GPU node to run. Build it as one deliberate change, with a version scheme and
+a smoke tier chosen on purpose, or not at all.
+
+## Weights are never in the image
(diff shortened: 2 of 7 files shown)


stewtong:
Doc-only update in dbc59caf: folded the measured 8x H200 serving reference into `cosmos3-super-serving.md` (parallel-strategy latency table with the fixed plus per-step decomposition, cache sizing ahead of first start, warm-readiness contrast for non-HSDP configs, the concurrency envelope with practical guidance, determinism band numbers, and a v0.26.0 cross-image recheck). Same measurement series the PR already cites, no code changes. This replaces the separate serving guide we had considered, so the two docs don't overlap.

PR #271, opened 2026-08-09
docs(cosmos3-generate): measured text2video timing for Cosmos3-Super on 8x H200 and 8x B200

stewtong:
## Summary

`docs/workbench/cosmos3-generate.md` currently quotes no latency numbers, and "Validated on real GPUs" covers Nano on RTX PRO 6000 only. This adds a measured-timing section for Cosmos3-Super text2video at the anchor shape (1280x720, 189 frames, 24 fps, seed 17, guardrails on), from the first real runs of the image on both 8x H200 SXM and 8x B200 SXM. The new section covers:

- measured wall and sampling time per invocation on both platforms
- the load-every-time cost structure behind the batch-versus-served comparison
- per-platform determinism findings
- the configuration single-GPU operation needs at 720p (`PYTORCH_ALLOC_CONF`) and the per-run output-prefix requirement

It also extends the stated GPU coverage to B200, which the repo README's GPU-coverage list does not include today. All numbers are traceable to the run records and manifests cited in the PR description's evidence table.

## Verification

- [ ] Ran the relevant unit, smoke, or workflow checks.
- [ ] For workflow/tool changes: ran or updated the matching skill smoke in `npa/tests/guardrails/test_skills_index.py`.
- [x] Checked public-repo hygiene: no customer names, VM IPs, private endpoints, project IDs, tenant IDs, registry IDs, bucket names, or secrets.

Docs-only change; no unit, smoke, or workflow checks apply.

## Skill Maintenance

- [x] I updated the relevant root `skills/` entry and `skills/index.yaml`, or this PR does not change behavior that needs skill guidance.

### DIFF
--- a/docs/workbench/cosmos3-generate.md
+++ b/docs/workbench/cosmos3-generate.md
@@ -228,3 +228,42 @@ before generation starts.
 For access checks before a run (`gh`/HF/NGC reachability) see
 `npa workbench cosmos check`. For the un-baked, clone-at-job-time text-to-image
 smoke, use `npa/workflows/workbench/npa-workflows/cosmos3-text-to-image.yaml`.
+
+## Measured timing: Cosmos3-Super text2video on H200 and B200
+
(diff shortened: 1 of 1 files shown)


stewtong:
@timothy-le7 ready for review when you have a moment: doc-only follow-up to the container work in #268, adding the measured batch-path timing section (H200 and B200) to cosmos3-generate.md. CI 9/9 green.

PR #281, opened 2026-08-13
Complete embedded Foxglove agent workflow

timothy-le7:
## Summary

Continues #267 and completes the Foxglove integration as a real in-agent viewer and export workflow.

- preserves lazy loading of the official `@foxglove/embed` SDK, runtime backend selection, the self-hosted Lichtblick fallback, explicit unavailable/error states, and state-only Describe-this behavior
- adds a canonical per-run MCAP plus provenance contract in S3, with native reuse, conversion from real artifacts, cache invalidation, cross-run isolation, byte-range/CORS transport, and identical bytes for embedded playback, download, and Cloud import
- adds explicit Download MCAP and Open in Foxglove Web actions; Cloud imports are content-idempotent, wait for indexing, derive layouts only from real channel schemas, and use documented bounded deep links
- keeps the Cloud credential server-side in the private NPA credential store and stages exact-project credentials for reproducible fresh-agent workflows
- makes the self-hosted fallback functional on a fresh driverless agent VM by installing/starting Docker during bootstrap when needed
- expands MCAP schemas/timing inspection, focused route/client/browser tests, live verification hardening, and operator documentation

## Impact

Operators can inspect a run in the embedded Foxglove pane without leaving the agent, download its persistent canonical MCAP, or explicitly publish the same recording to Foxglove Web. Run changes cancel stale exports and clear prior viewer state. Failures remain visible and actionable rather than presenting an empty frame.

## Validation

- required Foxglove pytest matrix: 124 passed
- mocked agent Cypress: 72 passed (55 general UI, 17 Foxglove)
- full repository suite: 8,888 passed, 48 skipped, 1 xpassed
- fresh live agent destroy/setup/bootstrap and `npa agent verify-live`: 27 passed, 2 skipped; embedded viewer root and recording routes returned 200
- `ruff check` on changed Python files
- generated CLI docs drift check
- staged and committed credential-value scans

## Limitations

- The official hosted viewer still requires a configured embed application/org and browser sign-in on a Foxglove plan that permits embedding. With no embed source, the pane deliberately uses the self-hosted OSS viewer.
- Describe this cannot capture pixels from the cross-origin Foxglove iframe; it reports grounded viewer/run/source state only.
- If the Foxglove account can upload recordings but cannot create or update shared layouts, the indexed recording opens without a `layoutId` and reports that fallback explicitly.


timothy-le7:
Consolidated the complete Foxglove agent-UI work into draft PR #267. PR #267 now carries the integrated implementation, final local/GitHub validation, and exact-SHA live deployment evidence, so this duplicate PR is retired.

PR #290, opened 2026-08-13
PAIDF: consolidate deterministic LeRobot quality routing

timothy-le7:
## Summary

Repairs Agent artifact viewing for large, paginated workflow inventories and exact source-qualified media access.

- Follows every server cursor before choosing a preferred artifact, merges and deduplicates pages, and makes later-page RRD/MCAP recordings immediately discoverable.
- Keeps pagination cancellation and viewer selection race-safe, including page-boot recording mount handoff.
- Carries the server-issued run reference and validated project, bucket, prefix, key, and source-selected state through cards, thumbnails, previews, downloads, and viewer autoload.
- Rejects unscoped raw object requests and avoids basename rediscovery when exact provenance is already available.
- Serves exact MP4 objects correctly for GET, HEAD, and byte-range requests, and prevents JSON/error responses from being attached to video elements.
- Keeps RRD loading on the same-origin static recording capability and retains authenticated blob/range checks as diagnostics.

## Motivation and impact

Inventories larger than one page could previously select a page-one video before discovering a higher-priority recording on a later page. Some preview/download links also lacked the exact authorization tuple, causing a JSON error body to reach browser media handling. The repaired flow computes preference globally, preserves source identity end to end, and gives operators actionable media errors.

## Validation

- Full Python unit suite: 11,769 passed, 49 skipped, 1 expected pass.
- Agent-focused Python suite: 238 passed.
- Agent mocked browser suite: 66 passed, including multi-page global preference, stale pagination, exact scoped media URLs, JSON rejection, and page-boot mount handoff.
- Full mocked browser gate: 113 passed.
- Smoke suite: 107 passed.
- Guardrail suite: 2,247 passed.
- Ruff, JavaScript syntax, staged secret scan, and diff checks passed.
- The existing Agent deployment was updated in place and its built-in live verification passed.
- Live browser validation confirmed automatic later-page RRD discovery, a visibly nonblank Rerun pane, and real MP4 metadata decode plus advancing playback.
- Live GET/HEAD/Range checks returned actual media bytes with the expected 200/206 semantics and headers.
- The existing MP4 is H.264/yuv420p with faststart, so no publication derivative or transcode was added.
- Read-only before/after inventory comparison found stable object keys, sizes, and modification metadata; the persistent dataset remained available.

## Limitations

The open-source Playwright Chromium binary available on the validation host lacks H.264 codec support. Rerun was validated with Chromium, while MP4 decode and playback were independently verified with the repository's codec-enabled Cypress/Electron browser harness.


timothy-le7:
CI status note: the branch-specific checks pass (ruff, docs drift, gitleaks, confidentiality scan, and the focused PAIDF suites). The remaining Test/guardrails failures reproduce on the current `main` SHA `41aaa13c` and are the pre-existing README-contract failures from main runs 31720886217 and 31720886298; this PR does not modify README. Live GPU validation is still running and will be added here at terminal state.

PR #363, opened 2026-08-31
feat(encord): add fail-closed S3 push and pull transport

stewtong:
## Summary

This adds an Encord integration to NPA Workbench. By default, it registers existing S3 objects with Encord without copying them. Operators can explicitly upload an Encord-managed copy, pull curated media back to S3, and verify each returned artifact against saved identity and integrity records.

The implementation is derived from [PR #339](https://github.com/nebius/nebius-physical-ai/pull/339) at source head `507964d2b6d7412f9195c842ab919baacdb9aee6`. It adds the stricter identity, failure, and evidence handling described below, and the first commit records the derivation directly.

This proposed PR was reviewed at head `902ebbb4fc4789f623d187dbef4227c83f1255bf` and is composed of three commits: `295196e5` (fail-closed transport core), `0a896908` (CLI, SDK, and workflow wiring), and `902ebbb4` (documentation).

## Exact identity replaces filename matching

The source implementation could accept an existing Encord item when its filename occurred once among the requested S3 keys, even if the full object paths differed. A minimal reproduction linked an unrelated item under that condition.

This implementation does not use filenames or basenames as identity. Push receipts record the canonical source URI, exact bucket and object key, submitted object URL, optional record ID, source size and integrity fields, Encord item UUID, and identity and transfer state. Conflicting or incomplete identity stays unresolved and prevents a completed receipt.

## Transfer and evidence contracts

- Register-in-place is the default. A registration failure never changes the operation to upload.
- Upload requires `--transfer upload` and records that the Encord-managed copy has a separate lifecycle and may be duplicated by repeated runs.
- Push writes a provisional receipt before the first change in Encord and checkpoints progress with discovered, attempted, successful, unresolved, failed, and unattempted counts that agree with the recorded items.
- Pull records every discovered media item and label artifact as successful, failed, unresolved, or unattempted, preserving earlier evidence across later failures.
- Label initialization is disabled by default. An explicit project-pull option enables it and warns that the operation changes the Encord project.
- `verify-roundtrip` consumes the final push receipt and pull manifest, checks exact item identity, confirms destination existence, and compares sizes and compatible checksums.

Encord remains remote SaaS. The Workbench integration is a stateless API and transfer client that runs on CPU. It adds no service, container, image, or GPU routing requirement.

## Repository integration

The change includes the shared implementation, validated receipt and manifest schemas, thin CLI and SDK surfaces, workflow tool references, three reference workflows, credential and health preflight support, operator documentation, and the Encord skill. It uses the repository's optional `npa[encord]` installation extra and keeps Encord authentication out of workflow specifications and durable artifacts.

## Verification

On the reviewed head, all 2,355 repository guardrail tests passed, and Ruff reported no issues across `npa/src` and `npa/tests`. All three Encord workflows passed validation and planning, and the generated CLI documentation matched the implemented commands. The selected Encord tests produced 148 passed and 9 skipped. The repository confidentiality scan and a checksum-verified gitleaks 8.28.0 scan returned no findings over the PR range.

## Full suite boundary

The full repository test command completed on the stacked PR 2 head and exited with code 2. Results were 12,578 passed, 98 failed, 49 skipped, 12 excluded by the test selection, and 1 expected-failure test that passed unexpectedly. Tracebacks show that 97 failures occurred in unchanged tests and came from host-specific constraints or missing tools, including missing `kubectl`, `/proc`, `flock`, and `tmux`, macOS Unix-socket path limits, platform-specific path assertions, ffmpeg behavior, and local timeouts. The remaining failure was `ModuleNotFoundError: No module named 'encord'`: the Python 3.14 full-suite environment did not include the optional `npa[encord]` extra. The same SDK contract test passed in the separate Python 3.12 environment with that extra installed. This is not a clean full-suite pass.

## Evidence boundaries

The current evidence is offline. Fixtures and a no-network contract test validate the supported Encord SDK interface and the transport's identity, failure, and evidence rules. No test authenticated to Encord or changed Encord or S3 data.

Live registration, upload, pull, label initialization, recovery after an induced partial failure, and repeating a push for the same item have not been tested. Those tests require separately approved credentials and isolated remote test data.

## Scope

This PR contains the push and pull transport only. Stable-ID reconciliation is reserved for the next stacked PR. Cosmos augmentation, demo seeding, GPU resources, and persistent services are excluded.

One durability limitation remains: if saving the receipt fails immediately after Encord accepts a change, the latest remote change may be missing from the last saved checkpoint. The client then stops, makes no further remote changes, and exits with an error instead of reporting success.


jonlwowski012:
@stewtong I took at a stab at fixing your failing test: https://github.com/stewtong/nebius-physical-ai/pull/1/changes

timothy-le7:
**Independent review at `ff06bbcf5` — the fail-closed claim holds under execution. No findings on the integrity gate.**

The PR title claims fail-closed transport, so I tested that rather than read it. The gate is `compare_checksums` in `npa/src/npa/workbench/encord/integrity.py:77`, and the fail-open mutation is the obvious one:

```python
return _normalize_checksum(expected) == _normalize_checksum(observed)
# mutated to:
return True
```

| | result |
|---|---|
| baseline | 147 passed, 10 skipped |
| mutated | **5 failed**, 142 passed |

Run on a pinned detached copy at this exact hash, with the import control confirming the candidate tree rather than my own.

**Five tests depend on the checksum comparison actually comparing.** An integrity gate that silently accepts is the failure mode that matters here — a corrupted or substituted object transferring successfully is worse than a transfer that fails loudly — and that path is asserted rather than assumed.

**Scope, stated plainly:** this covers the integrity gate only. The PR is +7004 across 20 source files including `clients/credentials.py`, `cli/workbench/health.py`, and workflow catalog/render changes, and **I have not reviewed those.** One verified property is not a clearance, and I would rather name the boundary than let a green mutation table imply more coverage than it has.

PR #380, opened 2026-09-03
docs(sim2real): document existing-cluster adoption, registry creds, Kueue install, source_sha, and resume retry

jeffmac-aix:
## Summary

Found live driving this runbook end-to-end against a real, externally-built mk8s cluster (not one npa provisioned), plus two more gaps a colleague flagged from the same live run after cross-checking against the current repo:

- The runbook's cluster setup assumed `npa provision-if-absent`, which plans a *second* Terraform-managed cluster on a project that already has one. There was no documented path to adopt an existing cluster into NPA's own local tracking. `submit` refuses with two distinct, actionable errors when this step is skipped ("No NPA cluster identity exists...", "No shared controller owner is bound...") but nothing pointed at `npa cluster kubeconfig` / `npa skypilot bind-controller` as the fix.
- Private-registry image preflight needs `SKYPILOT_DOCKER_SERVER`/`NPA_REGISTRY_SERVER` set alongside `USERNAME`/`PASSWORD`, or the resolved credentials are silently empty and every pull check 403s. Also documented the separate `~/.sky/config.yaml` `imagePullSecrets` merge nodes need at run time, since `sim2real.yaml` has no config var for it.
- **§4 creates the Kueue admission objects but never installs Kueue itself** — a fresh mk8s cluster has no Kueue CRDs. Added the pinned `helm install kueue oci://registry.k8s.io/kueue/charts/kueue --version 0.17.3 ...` command (note: no `v` prefix on the chart tag — `v0.17.3` returns "not found").
- **§6's `plan-spec`/`submit` examples never passed `--var source_sha`**, but `require_baked_npa` (the default) refuses to render or submit without a 40-hex `config.source_sha`, and every stage re-verifies it against the image's own baked `NPA_IMAGE_SOURCE_SHA` at runtime.
- `--resume-run` alone does not retry a wave that already reached a genuine terminal `FAILED` status, even after the root cause is fixed — it replays the recorded failure and refuses to resubmit. `--retries 1` is what authorizes the new attempt; this was undocumented.

Cross-checked the last two against the live repo/registry before writing them up: the other two items from the same colleague's report — GHCR image publishing coherence, and the compute-only GPU driver issue — are **already fixed upstream** (all 5 images now carry a matching `NPA_IMAGE_SOURCE_SHA` + bootstrap-contract label on a coherent GHCR tag; the driver issue is fixed by #368), so only Kueue-install and `source_sha` needed picking up here.

## Tests

Docs-only change. Full guardrails suite: 2,397 passed (this also caught and fixed a `cr.*.nebius.cloud` example hostname the legacy-registry-host guardrail now forbids in shipped docs — replaced with a generic placeholder).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jeffmac-aix:
Closing per author request.

PR #390, opened 2026-09-04
feat(workbench): add opt-in Nebius authentication preflight

stewtong:
Adds opt-in Nebius control-plane authentication verification to `npa workbench health preflight`, so operators can verify their CLI profile before provisioning.

- The default remains `hf,ngc,s3,token_factory`. Request `--checks nebius` explicitly; `--checks all` now includes it.
- Profile selection is `NPA_NEBIUS_PROFILE`, then `NEBIUS_PROFILE`, then the CLI active/default profile. Project selection does not select an authentication profile.
- PASS requires both `iam whoami` and IAM token minting. Both probes disable browser launch/update checks, close stdin, discard provider output, and scrub ambient IAM token variables. Each has the existing 30-second subprocess timeout.
- Authentication failures exit nonzero unless `--warn-only` is requested. Offline Nebius checks report SKIP with exit zero and JSON `ok: true`; callers must inspect the row before treating a profile as authenticated. PASS proves authentication, while resource permissions still depend on project access.
- Empty and unknown check selections fail before probing, including typos mixed with `all`; duplicate names run once. JSON output follows the CLI contract. A token-probe exception preserves identity verification already obtained.

The shared verifier now lives in `npa.clients.nebius_auth`; the VM-auth module retains compatible re-exports and interactive recovery. Documentation and operational skills describe the explicit provisioning check. This builds on the generic credential preflight from #197 and VM profile verification from #308.

Credit: **@stewtong (Stew Tong) authored the original PR and implementation.** Their original commits and authorship are preserved. Follow-up work integrates current main, addresses the input/error-handling edge cases, and completes Linux and live validation.

Validation evidence:

- Full repository unit suite: **13,124 passed, 36 skipped, 1 xpassed, zero failures**.
- Focused auth/health tests: 119 passed; onboarding tests: 114 passed.
- Real CLI coverage: 3 passed, covering authenticated PASS with stale ambient token values, missing-profile FAIL, and offline SKIP.
- Real cloud workload: after the authentication gate passed, created an isolated task-owned VPC network using the same selected profile, verified its identity/project, deleted it, and verified NotFound. No task-owned cloud resources remain.
- The pre-existing failing Cosmos denial-path test now isolates its download prerequisite and asserts that the vendor runtime executes once. Its denial and prompt-redaction assertions are preserved; all 69 tests in that file pass.

Repository validation also passed all 2,447 guardrails, Ruff, documentation drift, all 10 clean-install checks, and diff-scoped confidentiality/gitleaks scans. All seven GitHub checks passed on commit `183912c`. The [Python 3.12 CI run](https://github.com/nebius/nebius-physical-ai/actions/runs/33939891465) passed **13,137 tests** (392 skipped, 1 xpassed) with **73.62% coverage**, followed by all 10 clean-install checks. Exact operational identifiers and provider responses remain in private evidence.


stewtong:
PR description clarification after the final pre-review audit: I updated the description to state the changed `--checks all` contract, profile-selection order, offline `SKIP` and exit semantics, the existing VM-auth call-path impact, and the remaining unverified live `PASS`. Follow-up commit `875d8e5f` also makes the opt-in Nebius flag explicit in the four retrieval summaries; no other behavior changed.

PR #449, opened 2026-09-11
Add LIBERO runtime-fetch image and B200 workflow

timothy-le7:
Adds a neutral LIBERO bootstrap image and a headless B200 workflow for eight BC-RNN training steps on the official LIBERO-Spatial demonstration, checkpoint reload, and trajectory-disjoint held-out evaluation. Source, runtime dependencies, data, and weights are fetched at runtime after customer/run authorization.

The trusted public development build passed for `cc733b646f148667d07c475ed7ddea1ba3e60294`: [build and image verification](https://github.com/nebius/nebius-physical-ai/actions/runs/35494745037). The immutable image is `ghcr.io/nebius/nebius-physical-ai/npa-libero@sha256:6c81a42d9eabc8bf1b2adc8cfafa3ef82767cc970c59250007c6a5f853d54e82`. Independent anonymous pull verified its revision and complete OCI graph: 19 blobs, 408,283,271 bytes. Both registry provenance and SPDX 2.3 Sigstore bundles were independently signature-verified against that exact subject, source revision, workflow and build run.

The current operator source supports qualification records for those actual published artifacts, separates caller and customer verification keys, and mounts only provided immutable caller/customer registrations. Customer registration destinations are bound before profile signing; actual Pod verification accepts the checked-in profile and refuses unrelated ConfigMaps or writable mounts. Original image-producer evidence and current operator enforcement are bound separately. Docker build inputs remain byte-identical, so the existing public image is reused.

Validation: 478 focused tests, 114 CLI smoke tests, 3,859 guardrails, lint, confidentiality and secret scans passed. The full local suite recorded 24,350 passed, 154 skipped and one xpass; two unrelated GR00T visualization failures reproduced identically on the frozen base because the host Rerun CLI lacks its Python module. Both final CI runs **passed** at operator source `36481f948aa7cfe892cbd2589ba4b6633d790cad`: [Test](https://github.com/nebius/nebius-physical-ai/actions/runs/35548355935) and [Image Security Scan](https://github.com/nebius/nebius-physical-ai/actions/runs/35548356953).

The checked-in runtime qualification remains unconfigured. Completing it requires the real independent storage verification root and reviewed operator metadata. Real B200 execution also requires independently provided caller/customer trust material and fresh signed customer, caller and storage authorizations for the exact run/profile/image. No GPU run is claimed. Supported release promotion remains gated on real exact-digest B200 evidence.


timothy-le7:
**Review: minor**

This hand-rolls AWS SigV4 signing instead of reusing boto3 (already a repo dependency) — harder to audit, and no test was found exercising the signing path itself. Consider switching to boto3 or adding a signing-path test with known-answer vectors.

timothy-le7:
**Independent review at `cc733b646` — no findings on the payload scanner. Baseline 110 passed.**

I came at this with a specific hypothesis and it was wrong, which is worth stating first.

`npa/scripts/scan_image_libero_payload.py` is **1914 lines and imports nothing from the existing `scan_image_omniverse_payload.py`** — a full parallel implementation of an image payload scanner. My expectation was that hardening present in one would be missing from the other, specifically the BuildKit in-toto attestation handling that `#607` is currently adding to the omniverse scanner.

**It is not missing. This scanner is stricter.**

| property | omniverse (with #607) | libero |
|---|---|---|
| statement type check | prefix `https://in-toto.io/Statement/` | **exact** `== "https://in-toto.io/Statement/v0.1"` |
| predicate type | not inspected | `== "https://slsa.dev/provenance/v0.2"` |
| builder identity | not inspected | checked |
| base subject digest | not inspected | checked against manifest |
| label-vs-content | must parse as in-toto statement | must parse as a JSON object |
| content scanned regardless of label | yes | yes — `SECRET_CONTENT` runs over the blob before the media-type branch |

Exact equality makes the domain-prefix attack that `#607`'s tests cover (`in-toto.io.evil.example`) structurally impossible here.

**The one thing worth raising is structural, not a defect in this PR.** Two independent ~1900-line security scanners now gate redistribution claims, with no shared code and no mechanism keeping them aligned. Both are correct today, and they got there by different routes — one by a prefix constant carefully written to consume the domain boundary, the other by exact equality plus three additional field checks. **A future hardening applied to one will not reach the other, and nothing will report that.** Worth a tracked note about consolidation rather than a change here.

**Scope:** scanner comparison and the test baseline on a pinned detached copy. The rest of +36265 is unreviewed, and nothing here ran against a built image.

timothy-le7:
## Review lane — one durable finding on the new payload scanner

This PR is large (63 files, +36,760) and currently CONFLICTING, so a line-level
review would be invalidated by the rebase it needs. I looked instead at the part
that survives one: the credential coverage of the new
`scan_image_libero_payload.py`.

### The new scanner hand-rolls its rules, and three credential paths fall out

It does not import `image_payload_credentials`. It declares 17 path patterns of
its own. Compared against that shared module:

| path | libero | shared module |
|---|---|---|
| `.docker/config.json` | detects | detects |
| `.aws/credentials` | detects | detects |
| `.git-credentials` | detects | detects |
| `etc/ssh/ssh_host_rsa_key` | detects | detects |
| `kubeconfig` | detects | detects |
| **`.ssh/id_rsa`** | **misses** | detects |
| **`.ssh/id_ed25519`** | **misses** | detects |
| **`.netrc`** | **misses** | detects |

User SSH private keys and `.netrc` are absent from the path rule set.

### The content rule does not fully cover the gap

Fair objection: the scanner also has a private-key *content* rule, which could
catch a key regardless of its path. It catches some formats, not all. Its
pattern accepts only `RSA `, `OPENSSH `, or a bare prefix:

| key type at `~/.ssh/id_*` | libero content | shared content | libero overall |
|---|---|---|---|
| `RSA PRIVATE KEY` | detects | detects | caught |
| `OPENSSH PRIVATE KEY` | detects | detects | caught |
| `PRIVATE KEY` (PKCS#8) | detects | detects | caught |
| `EC PRIVATE KEY` | misses | detects | **undetected** |
| `DSA PRIVATE KEY` | misses | detects | **undetected** |
| `ENCRYPTED PRIVATE KEY` | misses | **misses** | **undetected** |

So a LIBERO image can ship an EC, DSA, or passphrase-protected key at
`~/.ssh/id_ecdsa` and this scanner reports nothing — neither rule fires.

Modern `ssh-keygen` writes OPENSSH format, which is caught, so this is not every
key. It is the openssl-generated, converted, and passphrase-protected ones.
`ENCRYPTED PRIVATE KEY` is the standard PKCS#8 encrypted form and entirely
ordinary.

### Why I think this is worth acting on rather than noting

This is the third independent occurrence of the same gap. I found user SSH keys
missing from the flex-pi scanner in #551 earlier today, the cross-scanner work
in #688 exists because three scanners had diverged the same way, and this PR
adds an eighth scanner that repeats it. Each is written by hand and each forgets
`.ssh/id_*`.

[#688](https://github.com/nebius/nebius-physical-ai/pull/688) is actively
consolidating scanners onto `image_payload_credentials` precisely so a rule
added once applies everywhere. Adopting it here is a smaller change than the
1,914 lines already written, and it removes this PR from the divergence problem
instead of adding to it.

### Method, and its limit

I tested the declared **rule sets**, not the assembled scanner. Its `scan()`
takes rootfs tar, config, build metadata and base provenance bytes, and I would
have had to fabricate those; a harness I invented would not have told you
anything about the real pipeline.

So the honest scope: these are the rules the scanner declares, and on those
rules the paths above are uncovered. If another mechanism in the assembled
scanner catches them — a different walker check, a manifest assertion — then the
end-to-end result may differ, and I would want to see that demonstrated rather
than assume it either way.

### Separately, for #688

`ENCRYPTED PRIVATE KEY` is missed by the shared module too, so this is not only
a libero gap. Raising it there.

### Position

Not a blocker from me on the PR's purpose, and I have not reviewed the other 62
files. One finding on the new scanner's credential coverage, with a concrete
undetected case. No pin at this head yet, since the rebase will move it.

PR #454, opened 2026-09-11
feat(workbench): add Habitat runtime-fetch bootstrap

timothy-le7:
Adds a neutral, non-root Habitat-Sim bootstrap image and the existing one-state RGB/depth, Bullet, NVIDIA EGL, and navigation workflow. The public image delivers the exact corresponding sources for its Ubuntu packages. Pinned Habitat source, scientific dependencies, and the official CC BY 4.0 Skokloster Castle scene are fetched and verified in the operator's runtime, with attribution and exact cache cleanup.

The current PR head is `27828856906152bb40349d156c77cd3d4d7f3cca`. Its final change fixes only the mock fixture's user/group identity. The validated image was built from `c7a04eda062664e1bd44d7ca5665a40fec79312d`; its runtime files are unchanged by that test-only commit.

Validated public image:

`ghcr.io/nebius/nebius-physical-ai/npa-habitat-sim@sha256:a2c77e0bdf2ee5c5f07db85043468150ae459680eda230e43f0096bb7dd2168c`

Validation:

- [Trusted public build](https://github.com/nebius/nebius-physical-ai/actions/runs/35490009559) and [image security](https://github.com/nebius/nebius-physical-ai/actions/runs/35490011240) passed for the image source commit. Anonymous full OCI pull verified 17 blobs and 390,217,857 bytes.
- The standard workflow succeeded on one RTX PRO 6000 Blackwell. Independent readback verified the observed image digest, GPU/node placement, NVIDIA EGL, 19 RGB frames, 19 depth frames, 19 Bullet steps, and 2.2466 metres of agent traversal.
- All 58 published objects, totaling 8,589,430 bytes, passed independent manifest/proof/frame hash readback. This is standard-workflow evidence; the separate legacy direct-pod transport selector is not claimed passed.
- Real Python 3.10 compatibility checks passed 108 tests; Python 3.12 workflow/image-contract checks passed 142. The fixture failure reproduced under actual UID/GID 1001, and all 12 selector tests passed after the fix. Independent read-only review passed each final scoped repair.
- [Test CI for the current fixture-only head](https://github.com/nebius/nebius-physical-ai/actions/runs/35492484283) passed, including all Python shards and coverage.

Exact workload, controller, namespace-resource, and local API cleanup passed, with outputs preserved. The shared validation cluster remains under coordinator ownership for another validation before final teardown.

Supported release selection remains quarantined. This PR does not request release promotion, merge, or auto-merge.


timothy-le7:
**Independent review at `278288569` — no findings on the packaging-contract gate.**

`habitat-sim` is declared `redistribution: public` while both Dockerfiles grant `ubuntu ALL=(ALL) NOPASSWD:ALL`. A public image with unrestricted passwordless root is exactly the combination worth checking, so I tested whether that exemption record is **bound to the artifact** or merely written next to it.

Mutation — make the contract declare a *narrower* grant than the Dockerfile actually contains:

```yaml
sudoers_entry: ubuntu ALL=(ALL) NOPASSWD:ALL
# mutated to:
sudoers_entry: ubuntu ALL=(ALL) NOPASSWD:/usr/bin/apt-get
```

| | result |
|---|---|
| baseline (verifier + license manifest + contract) | 552 passed |
| mutated contract | **1 failed**, 364 passed |

**The declaration is enforced against the Dockerfile, not just documented.** An understated or stale exemption cannot pass, which is the property that makes the `public` classification meaningful rather than aspirational.

**A correction on my own instrument:** I first flagged this PR as a runtime-fetch image carrying *no* payload-scan coverage. That was wrong — I had grepped for filenames containing `payload_scan`, and this PR names its coverage `test_habitat_sim_image_verifier.py`, `test_habitat_sim_license_manifest.py`, and `test_public_publish.py`, alongside a `REDISTRIBUTION.md`. The coverage was there under different names. I checked before publishing rather than after.

**Scope:** packaging-contract binding only, on a pinned detached copy with an import control. The claim is **not** verified against a built image, which `solution-licensing` requires before the `public` classification is fully established; +20092 across the rest of the PR is unreviewed.

PR #471, opened 2026-09-12
fix(agent): harden served APIs and public-pool deploys

timothy-le7:
## Summary

- add an optional public IPv4-pool input for agent deployments, without binding the implementation to any specific pool
- bound agent-side artifact S3 transport and discovery retries
- return an explicit incomplete state immediately on a cold artifact access/discovery path, complete it in the background, and retry once in the browser
- retain synchronous, source-qualified checks for exact artifact requests

## Validation

- focused agent/artifact suite: `436 passed`
- guardrails: `3441 passed`
- confidentiality scan and gitleaks: passed
- live fresh-agent smoke: authenticated UI `200`; anonymous root `401`; models `200`; grounded chat `200` with a nonempty response; cold artifact request `200` in under a second and background discovery subsequently returned five runs

## Live recovery note

The previously recorded VM was provider-confirmed absent, so recovery used a fresh agent-only deployment with restricted ingress. No destructive replacement was performed.

### DIFF
--- a/docs/agent.md
+++ b/docs/agent.md
@@ -66,6 +66,25 @@ npa agent fresh-setup --project "<alias>" \
 `fresh-setup` provisions the VM with Terraform. `npa agent bootstrap` refreshes
 only the UI/backend/nginx layer on an existing VM, without touching infra.
 
+### Optional public IPv4 pool
+
+Most deployments let Nebius select the default public IPv4 pool. If the target
+region requires an operator-provisioned public pool, pass its ID explicitly to
+the new agent deployment. NPA assigns it to the NPA-created network and makes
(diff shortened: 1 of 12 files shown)


timothy-le7:
Consolidated into #477 so the agent UI, shared public-pool option, artifact hardening, and workflow reliability changes are reviewed in one PR.

timothy-le7:
**Review: minor**

The new retry/timeout settings (`total_max_attempts=1`, `read_timeout=8s`) are applied to the *shared* S3 client used by all 14 call sites, not just the "interactive discovery" path the comment claims. A transient blip could become a user-visible failure on legitimate downloads. Please scope the aggressive settings to the discovery path or use a dedicated client.

PR #545, opened 2026-09-15
Add admin-permission warning and preflight pointer to configure

### LINKED ISSUE #519: `npa configure`: surface the admin requirement inline and point to preflight at the end
## Problem
Two gaps in an otherwise strong `npa configure` flow (`npa/src/npa/cli/main.py`, `_run_interactive_configure` :1410):
1. A valid-profile non-admin user discovers the storage-provisioning permission failure at provision time, mid-prompt-flow. The docs state the admin requirement and the `--no-provision` escape hatch exists, but the interactive flow never surfaces either before attempting provisioning.
2. `configure` never verifies the setup itself — no preflight runs at the end, so a typo'd endpoint or bad token is discovered later (the README quickstart covers `npa workbench health preflight --checks nebius`, but the flow doesn't point there).

## Suggestion
Surface the admin requirement in the interactive prompt before attempting provisioning, with the `--no-provision` alternative inline; end the flow with a pointer to `npa workbench health preflight --checks nebius`.

---

timothy-le7:
Fixes #519.

## What
Two additions to `_run_interactive_configure` in `npa/src/npa/cli/main.py`:

1. **Admin warning before the storage prompt.** Creating object storage
   requires Nebius project-admin permissions; non-admins are now told to
   answer `n` (or re-run with `npa configure --no-provision`) instead of
   failing at the provisioning step.
2. **Preflight pointer at end of flow.** After the summary line, prints
   `npa workbench health preflight --checks nebius --project <alias>`
   (alias included when set). Placed before the early returns so it appears
   on every path: provision, project-only, complete, and incomplete.

## Design decisions
- The warning goes *before* the prompt rather than in the failure handler
  because the failure mode is a permissions error mid-provisioning — cheaper
  to prevent than to explain after.
- The pointer uses `--checks nebius` (a valid check name per
  `SUPPORTED_CREDENTIAL_CHECKS`) rather than the full default set, matching
  the issue: configure just set up Nebius-side state, so that's what needs
  validating.

## Verification
- `ruff check`: passes.
- `test_configure_fresh_user.py`: 28 passed.
- `test_onboarding_smoke.py`: 16 passed.
- `test_main.py`: 98 passed.
- Full CI runs on this PR.

Closes #519.


timothy-le7:
## Test evidence (2026-09-15)

**CI:** 22 pass, 1 skipped — all green.

**Local validation:** 254 passed on the config-selection paths; the 2 failures reproduce on main (pre-existing, unrelated).

GPU: not applicable (docs pointer change; no GPU code paths).

Issue #520, opened 2026-09-15
Fix stale `npa workbench retargeting workflow` in the workbench-reference-workflows skill


timothy-le7:
## Problem
`skills/workflows/workbench-reference-workflows/SKILL.md:137` tells agents about *"tool-specific workflow commands such as `npa workbench mjlab workflow` or `npa workbench retargeting workflow`"`.

`npa workbench retargeting workflow` **does not exist**. The retargeting tool was moved under SONIC; the correct command is `npa workbench sonic retargeting workflow` (verified: `npa/src/npa/cli/workbench/sonic/cli.py:46` mounts `retargeting_app` under `sonic`; `docs/cli/retargeting.md` documents `npa workbench sonic retargeting` with `run`/`workflow`/`status`/`list`). This is agent-facing: an agent following the skill runs a nonexistent command. The rename is otherwise fully propagated (docs/cli, tool catalog, solutions-model.md, guardrail tests).

## Suggestion
Change the example to `npa workbench sonic retargeting workflow`.

---
*Filed from an onboarding audit of `main` (snapshot Sep 14, 2026). All findings re-validated against a fresh tree the same day.*

PR #546, opened 2026-09-15
Add guardrail for npa command mentions in docs and skills

### LINKED ISSUE #521: Extend the docs-drift guardrail to command examples in prose and skills
## Problem
The `docs-drift` CI job (`scripts/build_docs.sh --check`) regenerates `docs/cli/` from live `npa --help`, walking the full Typer tree — so all 58 CLI reference pages are provably current. But it cannot catch stale commands in prose guides or skills, which is exactly the class of bug behind the stale `npa workbench retargeting workflow` example in the workbench-reference-workflows skill.

## Suggestion
Add a guardrail test that extracts `npa ...` command stems from onboarding docs/skills and asserts each resolves against the Typer tree (the pattern already exists: `test_typer_command_calls.py`).

---

timothy-le7:
Fixes #521.

## What
New guardrail `npa/tests/guardrails/test_cli_docs_drift.py`: scans
`docs/workbench/`, `workflows/guides/`, and `skills/` for `npa ...`
command mentions and asserts each resolves in the Typer command tree.

## Design decisions
- **Scoped roots, not the whole tree.** The full docs tree holds historical
  design notes that intentionally reference unshipped commands; the
  onboarding-relevant roots are where a stale command actually misleads.
- **Same-line, top-level-anchored extraction.** A mention only counts when
  the token after `npa` is a real top-level command (kills prose false
  positives like "wants npa to drive"), and matching does not span
  newlines (a code block's next line is usually the next command).
- **Reasoned allowlist, not silent skips.** Positional args that read as
  subcommands (spec names, tool names, file names), prose fragments, and the
  deliberate "do not invent this tool" negative example each get an entry
  with a reason; a companion test fails on reason-less entries.
- **The #520 stale command is allowlisted, not fixed here.**
  `npa workbench retargeting workflow` does not resolve, and the entry says
  so with a pointer to #520, which owns fixing the docs. Verified by mutation:
  removing the entry makes the guard fail on exactly that stem.

## Verification
- `ruff check`: passes.
- New tests: 2 passed (resolution + allowlist-documentation).
- Mutation check: guard catches the stale command without its entry.
- Full CI runs on this PR.

Closes #521.


timothy-le7:
## Test evidence (2026-09-15)

**CI:** 22 pass, 1 skipped — all green.

**Local validation:** 113 drift-guardrail tests passed.

GPU: not applicable (docs guardrail; no GPU code paths).

PR #551, opened 2026-09-16
Onboard Flex-Pi inference and reproducible public GPU training

timothy-le7:
Add pinned Flex-Pi inference and real public training through the workbench CLI, SDK, service and workflow catalog. Training now enforces exact sample accounting, complete held-out validation, deterministic execution, and checkpoint byte readback followed by fresh-process continuation. JSON stdout remains one document.

The public workload is `flex-pi/sort_utensils` at revision `0780dd0a0b281df91abcef9434c4b3ac2757448c` (CC-BY-4.0), using the pinned upstream YAM 32D task: 136 training episodes / 115,620 anchors and 16 validation episodes / 12,390 anchors. It retains three RGB/depth views, official transforms and initialization, BF16, peak learning rate `1e-4`, global batch 96, and all 36 final-tail samples. Source revisions, downloaded byte hashes and split membership are recorded. Model weights and media are runtime fetches.

This workload is **non-comparable** to the operator's 5.66 samples/s reference. Exact private dataset identity, split lineage, checkpoint/configuration and use rights remain unverified after the residual investigation. Both unresolved JSONLs were scenario metadata; the related path remained unavailable; no surviving dataset mount or newly authorized matching data established identity. `reference_benchmark_beaten` remains false.

Checkpoint closure includes the original normalization file and all four rank RNG states. Independent jobs can pin the same normalization bytes. Strict deterministic algorithms and fixed DDP buckets support exact fresh-process replay. The adapter fixes eager microbatch one and accumulation 24. Experimental microbatch-three and RMSNorm compilation candidates were rejected by unchanged numerical gates; their selectable product surfaces have been removed. Nondefault optimizers require separate parity qualification.

A real four-B200 `profile-resume` completed 2,880 unique anchors. Its three sustained windows were 5.4709, 5.4366 and 5.4391 samples/s (median 5.4391); the six excluded startup/trace updates took 171.92 seconds. The complete checkpoint was uploaded and read back by byte hash. Fresh-process loaded model/optimizer/scheduler/cursor/RNG state and the exact next 96-anchor update matched uninterrupted execution bit-for-bit. This is early resume evidence, not a complete epoch or validation result. The complete-epoch run is now submitted with no wall-clock cancellation limit; full validation and final epoch-boundary resume remain pending. The earlier 7.17 samples/s profile predates deterministic execution and remains diagnostic only.

The full run preserves all numerical/data settings from the qualified profile. Removing the inactive compilation audit field changes its strict workload hash; the source and that sole configuration delta are separately frozen. The adapter supports one node with four GPUs; it does not represent four one-GPU RTX nodes as a single-node target.

The existing r2 image remains `sha256:e27978b682056339fb332acfdd0df2369af1f180d93ba1e0fd86772d6649efe6`. This continuation uses the standard source overlay and changes no image bytes. Historical one-GPU r2 inference measurements remain separate: B200 0.094180444 seconds and RTX PRO 6000 0.151054024 seconds for four compiled steps with finite 32×14 actions. These are inference results. DINOv3 retains its separate model terms.

Local validation after the normal merge of current main and removal of rejected candidates: 25,431 unit tests passed, 155 skipped and one xpass; 35 focused and 65 contract checks passed. Documentation drift, full Ruff, the security gate (zero blocking findings), staged/aggregate confidentiality, gitleaks and workflow validation passed. An initial unstaged-deletion source-inventory failure was corrected by staging the intended removals and rerunning the complete gate without changing tests. CI is pending for this milestone. Independent aggregate review found no functional or security launch blocker; full GPU acceptance remains pending.

Exact operational evidence remains outside Git in owner-controlled evidence. Run-owned GPU resources remain active for qualification; shared infrastructure is preserved. Trajectory delivery is pending because destination ownership has not been proved. Final cleanup and terminal results will be recorded after training completes.


timothy-le7:
Independent review of the security and licensing surface at head `3036d37cc`. I did not review the training/inference implementation or the GPU acceptance claims (see the last section for what I explicitly did **not** verify).

## The payload scanner is genuinely fail-closed

`scan_image_flex_pi_payload.py` backs the central REDISTRIBUTION claim that weights, caches and credentials "are absent" from image layers, so I tested it directly rather than reading it. I built eight synthetic `docker save` tarballs and ran the script from this PR unmodified:

| control | exit | verdict |
|---|---|---|
| clean image | 0 | `clean` |
| `opt/flex-pi/checkpoints/policy.safetensors` | 1 | `model_weight`, `packaged_checkpoint` |
| `etc/ssh/ssh_host_ed25519_key` | 1 | `credential_file` |
| private key under `opt/flex-pi/` | 1 | `credential_content` |
| `root/.cache/huggingface/hub/...` | 1 | `populated_model_cache` |
| 17 MiB `.safetensors` (over the 16 MiB read cap) | 1 | `model_weight`, `packaged_checkpoint` |
| `root/.ssh/id_rsa` | 0 | `clean` |
| `home/ubuntu/.ssh/id_ed25519` | 0 | `clean` |

The clean control produces no false positive, and the oversize case confirms the 16 MiB payload-read cap is not an exemption bypass: `payload` stays `None`, both exemption predicates return `False`, and the file is still flagged. The exit code is the enforcement, and the `>` redirect in `publish-public-images.yml` preserves it, so wiring the scan into both the pre-push and pushed-image phases does fail the job.

## One coverage gap: user SSH private keys

The last two rows are the gap. A private key at `~/.ssh/id_*` is not detected, for two reasons that compose:

- `FORBIDDEN_PATHS`'s `credential_file` pattern covers `ssh_host_(rsa|ecdsa|ed25519)_key` but not `id_rsa` / `id_ed25519` / `id_ecdsa`.
- `SECRET_CONTENT` does contain the `-----BEGIN ... PRIVATE KEY-----` matcher, but `_scan_archive` only applies it when `_application_content(name)` is true, i.e. under `opt/npa-src/` or `opt/flex-pi/`. The app-path row above proves the matcher works; the home-path rows prove it never runs there.

**This is hardening, not a live leak in this image.** The Dockerfile has no `ssh-keygen` for user keys and copies no `.ssh/` tree, and its host-key handling is already correct: `apt-get install ... openssh-server` and `rm -f /etc/ssh/ssh_host_*_key*` are in a single `RUN` (lines 3-13), so no key survives in an ancestor layer. The `(id -u ubuntu >/dev/null 2>&1 || useradd ...)` parenthesisation is also right, keeping `useradd` from running when the install fails.

Smallest fix is widening the one pattern:

```python
r"(?:^|/)(?:\.aws/credentials|\.git-credentials|\.docker/config\.json|kubeconfig"
r"|ssh_host_(?:rsa|ecdsa|ed25519)_key|id_(?:rsa|ecdsa|ed25519|dsa))$"
```

## Why this gap matters more than usual right now

On `origin/main`, `_scan_trivy_exact_digest` invokes Trivy with `--scanners vuln,secret` **and** `--severity CRITICAL`. Trivy applies `--severity` to every enabled scanner at once, and it reports private keys at HIGH, so secret findings are dropped before the Python parser can reject them. #687 fixes exactly this and is still open at `2f9032465`.

So the Trivy layer does not currently back-stop the gap above. Both are narrow on their own; I am flagging the combination rather than either one.

## The gitleaks allowlist entry is correctly justified

I checked the new entry against the rule its comment blames, `nebius-generic-id` = `\b[eu]00[a-z0-9]{12,}\b`. `e00cc33ab9e0dfdd...` does match it, so the "coincidental e00" justification is accurate, and the entry is a single exact literal rather than a widened pattern. No objection.

Unrelated to this PR, the same check found three pre-existing entries (`e071e85c...`, `e091cc3e...`, `e0262be9...`) whose comments credit the Nebius identifier rule although they begin `e07`/`e09`/`e02` and cannot match that regex. Their real trigger is undocumented. I am raising that separately against main, not here.

## Licensing

The REDISTRIBUTION record is careful in the place that usually goes wrong: it declines to infer a license for `flex-pi/robotwin_3d` from the MIT simulator source because the dataset card declares none, and it keeps sample bytes runtime-only rather than redistributing them. The `packaging-contract.yaml` entry matches that record and names the accepted digest.

## Not verified

- **GPU acceptance.** The contract states the release passed independent B200 and RTX PRO 6000 inference gates. I have not reproduced those and am not treating the statement as verified; per the program rule, author self-certification is not acceptance.
- **Built-image bytes.** All scanner results above are against synthetic archives that exercise the scanner's logic. They do not show what the real published layers contain, which needs the scan bound to the built digest.
- **VLM: not applicable.** This surface produces no visual artifact.

No objection to the gitleaks, publication-gate and licensing changes. The user-key pattern is the one change I would ask for.

PR #556, opened 2026-09-17
Keep merge queue validation under fifteen minutes

timothy-le7:
## Summary

- use five duration-balanced pytest shards for merge-group candidates while retaining four shards per interpreter for the full compatibility audit
- distribute the seven pinned base-image scans dynamically across three worker-private Trivy caches
- run the full Python 3.10/3.12/3.14 audit daily at 05:23 UTC instead of launching 12 long pytest jobs immediately after every merge
- preserve the complete suite, merged coverage floor, browser checks, focused compatibility checks, fixed-CRITICAL base-image policy, and isolated scanner caches

## Measured reason

PR #553 reduced the secure merge-group gate from more than 30 minutes to 16m48s under active runner contention. The remaining execution tails were the base-image inventory and the slowest pytest shard. Its post-merge three-version test audit also competed with the next queued candidates for hosted runners.

This change removes those tails without deleting or making a blocking check advisory. The queue still validates the latest-main synthetic commit, and the full three-version audit remains automatic on a daily schedule.

## Validation

- 3,558 guardrail tests passed
- 177 focused CI, image-security, sharding, and skill-contract tests passed
- full repository Ruff passed
- diff-scoped confidentiality scan passed with 0 unresolved findings
- gitleaks scanned the complete commit range with no leaks


timothy-le7:
Operational settings applied alongside this PR:

- removed the empty classic push-restriction allow-list that caused green merge groups to fail with `branch_protection_failure`; required checks and signed commits remain enabled
- changed merge queue `max_entries_to_build` from 2 to 1 so the head candidate gets the runner pool and later PRs do not create speculative stale runs
- changed `check_response_timeout_minutes` from 30 to 45 as a fail-safe for GitHub-hosted runner backlogs; this does not add waiting to successful runs
- preserved `ALLGREEN`, squash merging, zero minimum wait, and one PR per merge group

The PR-head security pipeline is green. Its deep three-worker base-image inventory completed in 12m19s; all required security, leak, confidentiality, lint, guardrail, smoke, and hostile-input checks passed.

PR #557, opened 2026-09-17
Allow external queues to retry temporary GPU capacity failures

avidor-b:
## Summary

External schedulers cannot currently distinguish an impossible GPU request from a supported request waiting for occupied capacity. This makes it difficult to keep a job queued safely instead of failing it or retrying an uncertain launch.

Return exit code `75` from `workflow submit` only for typed, retryable capacity failures before any provider launch. Preserve the existing pending-GPU-placement exception and release the lifecycle operation after a proven preflight failure; ambiguous or attempted launches still require reconciliation.

- Accept positive integer GPU counts within a node's advertised maximum, including six GPUs on an eight-GPU node.
- Check nominal fit separately from free GPU, CPU, memory, pod-slot and ephemeral-storage capacity. Six GPUs split across two nodes cannot satisfy a one-node request; disjoint six- and two-GPU requests can share one node when their other resources fit.
- Apply storage accounting at both CLI and SDK preflight boundaries, including active containers, init containers and pod overhead.
- Document the external scheduler contract and add regression coverage for fragmentation, impossible requests, pending placement, storage pressure and submission-lock recovery.

This exposes an admission result for an external queue, such as Airflow. It does not add a queue service or reserve observed capacity; Kubernetes remains authoritative for placement.

## Verification

Core candidate `5d3637b5` was validated against upstream `fa9b7590` in an isolated Linux ARM64 checkout using Python 3.12.12. The subsequent `cdff4dc8` adds only the live collective probe and its procedure; it changes no production behavior.

- [x] Ruff: passed.
- [x] CLI smoke: 114 passed.
- [x] Final targeted CLI, SDK, capacity, resource-quantity, lifecycle and existing PAIDF workflow regressions: 510 passed.
- [x] Repository guardrails: 3,625 passed.
- [x] Gitleaks and built-in infrastructure confidentiality scans: passed.
- [x] Security regressions with CPU-only PyTorch 2.13.0: 783 passed.
- [x] Generated CLI documentation: up to date.
- [ ] Full non-live suite is not green on this host: 22,393 passed, 258 skipped, 1 xpassed, 35 failed and 71 setup errors. One failure was the existing PAIDF capacity-report expectation, now corrected and included in the passing 510-test rerun. Rerunning all 106 failed/error cases against untouched upstream reproduced the other 34 failures and all 71 setup errors. These include missing FFmpeg/ffprobe, unsupported ARM64 native tooling, unavailable `os.memfd_create`, and existing simulator/export test failures. No unrelated source changes or test exclusions were added.
- [x] Upstream candidate CI for `5d3637b5`: successful.
- [ ] CI for the added probe/docs commit `cdff4dc8`: the fork workflow requires maintainer approval.
- [x] Live PCIe placement: sequential 8-, 6-, and 8-GPU managed jobs on one existing eight-GPU RTX PRO 6000 Blackwell node all reached `SUCCEEDED`. The six-GPU case passed the real catalog resolver, requested/received six GPUs in one pod, exposed six distinct devices from the full-node set, and passed every all-reduce correctness check. The same digest-pinned image was used throughout (PyTorch 2.8.0+cu128, NCCL 2.27.3, driver 580.173.02).
- [x] Measured transport: all three jobs used NCCL `SHM/direct/direct` data channels on PCIe. At 256 MiB, normalized all-reduce bandwidth was 24.85 GB/s for six GPUs and 25.37 / 24.89 GB/s for the two eight-GPU controls. Results establish the tested PCIe configuration only; they do not guarantee identical performance at every message size.
- [x] Cleanup: test workers and the isolated controller/API are gone; the existing training job and other controllers were preserved.
- [ ] NVLink/NVSwitch validation remains outstanding: the available hardware has no such fabric. This result does not satisfy a strict NVLink/NVSwitch qualification requirement. The reproducible probe and procedure are included for that follow-up.

## Skill Maintenance

- [x] Updated the existing `skypilot-workflows` skill and workflow guide. No new skill or index entry is needed.
- [x] No production identifiers, private endpoints, credentials or training assets are included.


timothy-le7:
Thank you for the PR. Would you please address the following:

Live GPU placement validation needed: the non-power-of-two GPU allocation change (e.g. requesting 6 of 8 GPUs on a node) has only been tested against mocked/stub inventory data, not real hardware. Need at least one real placement run to confirm the interconnect (NVLink/NVSwitch) isn't silently degraded on a non-power-of-two split before this merges.

avidor-b:
Live **PCIe** validation is now complete, with an important limit: the available eight-GPU nodes have RTX PRO 6000 Blackwell PCIe GPUs, not NVLink/NVSwitch. I cannot claim NVLink/NVSwitch qualification from these results.

I ran **8 → 6 → 8 GPUs on the same physical node** through this PR's NPA SDK submission path and SkyPilot 0.12.2. All three benchmark jobs reached `SUCCEEDED`.

- The six-GPU request passed `resolve_kubernetes_accelerator` against the actual SkyPilot catalog, despite its displayed quantities being `1, 2, 4, 8`, and passed the real Kubernetes inventory preflight.
- Kubernetes assigned one pod with GPU requests and limits both equal to six. Its six ranks had six distinct GPU UUIDs, all belonging to the eight-GPU baseline's device set. The pod/node evidence was captured while the worker existed.
- All three jobs passed every collective correctness check: five FP32 message sizes, three repeats each, 20 warm-ups followed by 100 timed all-reduces per repeat. All ranks completed.
- The image digest, CPU/memory/shared-memory allocation and probe were identical. Runtime: PyTorch **2.8.0+cu128**, CUDA **12.8**, NCCL **2.27.3**, driver **580.173.02**. Only NCCL debug logging was enabled; the final control also verified that the image's system/user NCCL config files were absent.

Median normalized all-reduce bus bandwidth, GB/s:

| Message | 8 GPUs before | 6 GPUs | 8 GPUs after |
|---|---:|---:|---:|
| 1 MiB | 12.65 | 14.69 | 12.91 |
| 16 MiB | 27.37 | 25.41 | 27.36 |
| 64 MiB | 26.14 | 25.59 | 25.59 |
| 256 MiB | 25.37 | 24.85 | 24.89 |

The normalization is `algbw × 2 × (N−1)/N`, using the slowest rank's elapsed time. This is the checked-in PyTorch distributed probe, not a run of NVIDIA's `nccl-tests` executable.

Topology showed PCIe `NODE`/`SYS` links. NCCL reported **`SHM/direct/direct` data channels in all three cases** and no NVLS support. There was no observed transport change for six GPUs. The table preserves the message-size differences rather than claiming identical performance: six GPUs were about 7% lower at 16 MiB and effectively matched the repeated eight-GPU control at 64 MiB; at 256 MiB they were 0.14–2.04% lower.

I added the exact executed probe and a repeatable qualification procedure in `cdff4dc8`. Core execution used `5d3637b5`; the added commit changes no production behavior. One earlier setup attempt failed before worker provisioning because raw SkyPilot YAML used an unadvertised accelerator alias; the successful runs use the exact live catalog name.

Cleanup is verified: all test workers and the isolated controller/API were removed. The existing four-GPU training job remained running and the other five controllers were preserved.

Private evidence SHA-256:

- Eight-GPU baseline log: `8c4c904499f4d347341620411311740cf510c61d7c7bed952596e9feb2a333bd`
- Six-GPU log: `54ad1e0fc48716dd7571b399df03d1d833127ae4cba9cfde4a2322ad1d81ef69`
- Eight-GPU repeated-control log: `1c66e67385e6fbc629727fd20cb31dc574d04dd997d6251b8405de905b4f8acb`
- Executed/committed probe: `a22c423ea2698f8d0b3b1624b8fc5839d4b67c17a1db49f8037df09c35612651`

**Remaining before claiming NVLink/NVSwitch safety:** repeat this procedure on hardware with that fabric and verify its NCCL transport and bandwidth. The available PCIe result cannot close that specific requirement. The new fork CI run also needs maintainer approval; the preceding production-code candidate passed upstream CI.

PR #561, opened 2026-09-18
Close the two offline FIXMEs and root-cause the third

timothy-le7:
## Summary

Two `FIXME.md` entries closed, a third root-caused without a cluster, a fourth re-checked and deliberately left open. Plus one test-robustness bug found along the way.

- **SDK_PUBLIC_SURFACE.** `npa.sdk.workbench.<tool>` is the documented Python entrypoint, but `npa.sdk` was missing from the top-level lazy submodule list, so `import npa` followed by `npa.sdk.workbench.<tool>` raised `AttributeError` and `dir(npa)` never mentioned it. Two more defects the audit turned up: both SDK namespaces imported eagerly, so `import npa.sdk` pulled 349 `npa` modules plus pyarrow and boto3 where it now pulls two and no third-party dependency — which is the laziness the top-level package deliberately introduced for the minimal images; and `npa.sdk.workbench.lancedb` resolved to either the implementation module or the SDK client depending on import order. `__dir__` now reports `__all__` instead of the module globals, which were exposing `_importlib`, `_LAZY_SUBMODULES`, and `TYPE_CHECKING` as public. `sim_to_real` was a client nobody could reach by attribute access; it is exported, and a test now fails on any public SDK module left out of `__all__`. Surface and v0 stability documented in `docs/sdk/README.md`.

- **Standalone LeRobot validation.** The 25 conversion tests read the exported parquet and JSON directly, so none of them can see a failure that only happens when `LeRobotDataset` joins meta, data, and videos. Three new tests load a converted dataset with the real library: loaded metadata, sample 0's state and action shapes, and both video-backed cameras decoding to normalized channels-first frames with non-zero variance — the mp4 path the parquet assertions never touch. A third case downgrades the exported `codebase_version` to `v2.1` and asserts the loader raises, pinning that field as load-bearing rather than cosmetic.

- **cosmos-curate bootstrap stall.** Root-caused statically from SkyPilot 0.12.2's `sky/skylet/constants.py` and the published image, with no cluster. No code changed; see below.

- **Remote-env upload extraction.** Still two tools, so the "another tool needs it" trigger has not fired. But they hold *three* helpers with three different payload shapes — local directory, remote directory, remote single file — so the originally proposed `upload_via_remote_env(host, local_path, s3_uri, env_file_path)` would add a fourth shape rather than remove duplication. What actually repeats is the env-sourcing preamble and the boto3-client-from-env construction. Recorded in `FIXME.md` so the next reader does not re-derive it.

## The bootstrap stall, in short

The PATH hypothesis was right but not in the shape it was written, and the corrected version explains the one observation that did not fit. The ray version check is `uv pip list | grep "ray "` inside SkyPilot's own `~/skypilot-runtime` venv, so the image venv's ray is invisible to it and is not what skips the install. What reaches for the ambient PATH is the `|| RAY_STATUS` fallback on the next clause: `SKY_RAY_CMD` resolves bare `which ray` whenever `~/.sky/ray_path` is missing, and that file is only written by the last clause of the same command. On a fresh pod the fallback fires first, and the image's `PATH=/opt/cosmos-curate/venv/bin:...` puts ray 2.57.0 (verified from the image layers) ahead of everything, against SkyPilot's pinned 2.9.3.

So SkyPilot's python runs the cosmos-curate ray script and asks a non-existent GCS on `127.0.0.1:6380` for status before any install has happened. That also accounts for the detail that read as a contradiction: there was "no ray, pip, or uv child process" because the process is `$SKY_PYTHON_CMD` running ray's entry script, so it is named `python`. The evaluator image never stalling fits too — it has no ray on PATH, so the clause fails fast into the install.

This is a static read, so it establishes the mechanism and not the hang. `FIXME.md` records three candidate fixes with their tradeoffs and applies none: the entry already said to confirm before changing anything, and an image change needs a rebuild plus live validation and catalog reconciliation.

## The test-robustness bug

The full suite kept failing `test_streaming_encoder_reports_real_ffmpeg_failure_and_broken_pipe` on this host while passing in isolation. `pytest.raises(match=...)` is `re.search` without `DOTALL`, so `"ffmpeg failed.*npa_missing_codec"` only matches while the encoder error stays on the message's first line. Real ffmpeg stderr is multi-line, and on a host whose dynamic linker warns, those warnings land ahead of it — ffmpeg had reported `Unknown encoder 'npa_missing_codec'` and the adapter had raised exactly the right error, but the regex could not reach past the newline. `(?s)` keeps both halves of the assertion rather than loosening it to the codec name alone.

## Verification

- [x] Ran the relevant unit, smoke, or workflow checks.
- [ ] Checked the PR's Python 3.10, 3.12, and 3.14 test results; recorded any pending or failing checks below.
- [x] For image/security changes: n/a — no image, packaging, or security-policy change.
- [x] For workflow/tool changes: n/a — no workflow, tool, or `toolRef` change. Skill guardrails (`test_develop_skills`, `test_skills_index`) pass.
- [x] Checked public-repo hygiene: confidentiality diff scan over `origin/main..HEAD` passed (`raw=0 dispositioned=0 unresolved=0`).

Numeric results on Python 3.12:

| Check | Result |
| --- | --- |
| Full unit suite, `-n auto`, e2e/GPU markers deselected | **22928 passed, 164 skipped, 1 xpassed, 0 failed** (5m49s) |
| `ruff check npa` | clean |
| `test_documentation_examples` | 624 passed |
| `test_sdk_surface.py` | 20 passed |
| `TestLeRobotLibraryLoad` against `lerobot==0.5.1` | 3 passed in 51s |
| `TestLeRobotLibraryLoad` with no lerobot | 3 skipped in 0.4s |
| Confidentiality diff scan | passed |

Getting to a green suite took three host dependencies that `npa[dev]` does not ship: CPU `torch==2.13.0` (as CI installs separately), `onnxscript` (torch 2.13's ONNX exporter needs it), and `onnxruntime` (the `npa[sonic]` extra; the code says so in its own error message). Every failure before those were installed was `ModuleNotFoundError`, and each reproduced identically on an `origin/main` worktree against the same virtualenv.

Both new-behavior guards were checked against mutation rather than assumed: adding a public module under `npa/src/npa/sdk/workbench/` fails `test_sdk_workbench_exports_every_tool_client_on_disk` while an underscore-prefixed one does not, and the ffmpeg regex fails pre-fix under `LD_LIBRARY_PATH=/exec-daemon/tmux-root/lib` and passes post-fix under both that and a clean environment.

## Skill Maintenance

- [x] `skills/atomic/testing-conventions/SKILL.md` gains an "Optional Third-Party Library Validation" section: how to run the LeRobot test against the real library in a throwaway virtualenv, its `ffmpeg` and `python3-dev`/`linux-libc-dev` prerequisites, and why the `importorskip` belongs in an autouse fixture. A test that only ever skips proves nothing, so it needs to be discoverable.

<div><a href="https://cursor.com/agents/bc-01a0b289-093d-7c4f-b623-453235a22397?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-01a0b289-093d-7c4f-b623-453235a22397&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


timothy-le7:
### Surfaced by a coverage sweep — this needs a disposition decision, not a code review

Review lane, 2026-09-20T09:34:21Z. A coverage sweep over open `cursor/` PRs turned this up as
the one PR with no record in my review tracking. Before spending review effort
on the diff, the state is worth stating:

- Created `2026-09-18T04:02`, last touched `2026-09-18T05:05` — **no
  activity in two days**, and it predates the current improvement program.
- Branch `cursor/close-offline-fixmes-2397` does not match the program's
  naming pattern, which is why it never entered my queue.
- `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY` — **it conflicts with
  `main` as it stands.**

Reviewing +514/-80 across 11 files that cannot currently merge would mean
reviewing a diff that has to be rewritten against a two-day-newer `main`
before it means anything, and several of those files (`npa/src/npa/__init__.py`,
the SDK `__init__` chain) are exactly the kind that collect conflicts.

So I am not reviewing it yet, and I am saying so rather than silently skipping
it — silently skipping is how it stayed invisible for two days in the first
place.

**What I need to proceed:** rebase it onto current `main` and take it out of
draft, or close it. If it is rebased I will review it on the same terms as any
program PR. If the FIXMEs it closes still matter but the branch does not, say
so and I will file them as backlog items so the work is not lost with the
branch.

timothy-le7:
## Review lane, 2026-09-20T10:22:14Z — measured this instead of leaving the disposition open

My earlier triage asked for rebase-or-close without telling you which. That was not much help, so I measured it. **My recommendation is rebase, not close**, and the numbers are cheaper than they look.

### The rebase is small
- **2 conflict hunks** against current `main` (via `git merge-tree`)
- 25 commits on `main` since this branch's base

### It is partially landed, not superseded
Two of the three new files now exist on `main` independently — but in smaller versions:

| file | this PR | on `main` |
| --- | --- | --- |
| `npa/tests/test_adapter.py` | 521 lines | 455 lines |
| `npa/tests/test_sdk_surface.py` | 336 lines | 232 lines |
| `docs/sdk/README.md` | 77 lines | **absent** |

So the test files diverged rather than being replaced, and this branch carries roughly 170 lines of test coverage beyond what landed. Worth a look at whether that surplus is real coverage or drift.

### Two things here exist nowhere else
1. **`docs/sdk/README.md`** — a map of the public SDK surface. Nothing on `main` covers it.
2. **The `FIXME.md` SkyPilot analysis** — a statically confirmed mechanism for why `RAY_INSTALLATION_COMMANDS` version-checks against SkyPilot's own venv (`VIRTUAL_ENV=~/skypilot-runtime`), making the image venv's `ray` invisible to that check. It explicitly corrects an earlier hypothesis and accounts for the one observation that did not fit, derived from reading `sky/skylet/constants.py` and the published image with no cluster.

That second item is diagnostic knowledge that would be genuinely expensive to rediscover, and it should not die with a stale branch. I am filing it as a backlog item so it survives either decision — but if this PR rebases, it lands in the right place instead.

If you would rather close this, say so and I will lift the docs page and the analysis into follow-ups. Either way this should not keep sitting.

PR #566, opened 2026-09-18
Pin remaining unpinned Dockerfile FROMs

### LINKED ISSUE #506: Eliminate remaining unpinned FROMs in workbench Dockerfiles
## Summary
The digest-pinning story the repo otherwise enforces has a few soft exceptions:

- `FROM rayproject/ray:2.46.0-py310-gpu`
- `FROM python:3.12-slim`
- `FROM python:3.11-slim`
- `FROM nvidia/cuda:${CUDA_BASE_TAG}` (parameterized, in `npa/docker/workbench/base/cuda13-b300/Dockerfile`)

## Suggested fix
Resolve digests per `docs/security/image-reproducibility.md` and extend the base-image CVE scan matrix in `image-security-scan.yml` to cover these images.

---
Identified during an external audit of `nebius/nebius-physical-ai` at main `d743f18` (Sep 2026).

timothy-le7:
Closes #506.

## What
- Pins  in  to digest 
- Pins  in  to digest 
- Adds the cuda 13.0.1 base to  scan matrix

## Testing
- : 355 passed
- Digests resolved via Docker Hub registry API per


timothy-le7:
**Review lane — unsolicited, same grounds as #570 and #580. No findings.**

Digest-pinning `FROM` lines is the right control:

```
nvidia/cuda:13.0.1-cudnn-devel-ubuntu22.04@sha256:6ff9753c…
```

A tag is mutable and a digest is not, so this closes the case where an upstream retag silently changes what a build consumes. The catalog entry carrying `purge_linux_libc_dev` and `upgrade_os` flags alongside it means the update path is explicit rather than implicit, which is the part that usually goes missing when people pin by digest.

**One observation, not a finding, and it is the mirror of what I raised on #570.** A digest pin means upstream security rebuilds of that tag no longer reach you automatically. That is the correct trade for reproducibility, and unlike an unattended version ceiling it is visible — a stale digest is inspectable in the file, where a `<6.0` bound silently holds a package back. I mention it only so the two PRs are read as the pair they are: both trade automatic updates for control, and both need someone to own the refresh.

`CLEAN`, undrafted, open since 09-18.

PR #570, opened 2026-09-18
Tighten dependency floors for security-sensitive packages

### LINKED ISSUE #486: Tighten dependency floors for security-sensitive packages
## Problem
`npa/pyproject.toml` uses deliberately wide, floor-only ranges for security-sensitive client deps (`httpx>=0.27`, `paramiko>=5.0`, `bcrypt>=5.0`, `pyyaml>=6.0`, `rich>=13.0`, `boto3>=1.34`, `jinja2>=3.1.6`, `pillow>=12.3.0`), with the comment "deliberately wide above reviewed security fixes". Fresh installs therefore pull the newest matching versions, so newly disclosed CVEs land immediately. `genesis-world` in the `genesis` extra is fully unpinned. The CI dependency-regression gate and weekly image scans only partially mitigate this.

## Suggestion
Adopt a lockfile (or tighter upper bounds) for the security-sensitive client deps, and pin `genesis-world`.

---

timothy-le7:
Closes #486.

## What
Adds upper bounds to security-sensitive dependencies to prevent unreviewed major version jumps that could introduce CVEs:

- httpx>=0.27,<0.29
- paramiko>=5.0,<6.0
- bcrypt>=5.0,<6.0
- pyyaml>=6.0,<7.0
- rich>=13.0,<16.0
- boto3>=1.34,<2.0
- jinja2>=3.1.6,<4.0
- pillow>=12.3.0,<13.0
- genesis-world>=1.4.1,<2.0 (was fully unpinned)

## Testing
- pyproject.toml syntax verified


timothy-le7:
**Review lane — unsolicited; this PR may be outside my charter. I reviewed it because I reviewed #580 on the same grounds and it would be inconsistent to stop here. One finding, non-blocking.**

## The change adds ceilings, not floors

Every specifier keeps its existing lower bound:

```
-"httpx>=0.27",          +"httpx>=0.27,<0.29",
-"paramiko>=5.0",        +"paramiko>=5.0,<6.0",
-"bcrypt>=5.0",          +"bcrypt>=5.0,<6.0",
-"pyyaml>=6.0",          +"pyyaml>=6.0,<7.0",
-"rich>=13.0",           +"rich>=13.0,<16.0",
-"boto3>=1.34",          +"boto3>=1.34,<2.0",
-"jinja2>=3.1.6",        +"jinja2>=3.1.6,<4.0",
-"pillow>=12.3.0",       +"pillow>=12.3.0,<13.0",
-"genesis-world",        +"genesis-world>=1.4.1,<2.0",
```

**Eight of nine raise no floor at all.** `genesis-world` is the only one that gains a lower bound, going from entirely unconstrained — and that one is a clear improvement.

The title and body describe raising floors for security-sensitive packages. What the diff does is cap majors. Those are different changes with different risk profiles, and I would not have known which I was getting from the description.

## The finding: for these packages specifically, a ceiling is a liability

Upper bounds in an application are defensible — they stop an untested major from arriving silently. But the argument inverts for exactly the packages named here:

- **`paramiko<6.0`** — SSH transport
- **`bcrypt<6.0`** — password hashing

**When paramiko 6.0 ships carrying a security fix, this pin blocks it** until someone notices and bumps. A ceiling on a security-sensitive dependency converts an automatic patch into a manual chore, and the failure mode is silence: nothing breaks, you simply stay behind.

Compare #580, which I reviewed an hour ago. That PR existed *because* a dependency sat at a vulnerable version. A ceiling is the mechanism that produces that situation and holds it there.

**Not a blocker.** Capping majors is a legitimate stability choice and `genesis-world` genuinely needed its floor. But I would want either the title corrected to say what it does, or a note on who watches for the majors these bounds will eventually hold back — because an unattended upper bound on `paramiko` is the next #580.

timothy-le7:
**Re-measured at `eb46ba81f`, and correcting my own earlier finding on this PR.**

My pinned note said this PR *"adds ceilings not floors."* That is wrong, and it mischaracterised the change as not doing what its title says. The actual constraints in `npa/pyproject.toml`:

```toml
"paramiko>=5.0,<6.0",
"bcrypt>=5.0,<6.0",
```

It adds **floors and ceilings**. The floor is exactly what the title promises, and my note read as if it were absent.

**The narrower concern survives, and only this part should be weighed.** `<6.0` means that when `paramiko` or `bcrypt` publishes a 6.x carrying a security fix, resolvers will not take it until someone edits this file. For libraries whose entire job is transport security and password hashing, a major-version cap converts an automatic upgrade into a manual one, and the cost of that is paid at the worst possible moment.

That is a judgement call, not a defect: caps also prevent an unreviewed major from breaking SSH at runtime. Worth an explicit decision rather than an inherited default — and if the caps stay, they are worth a comment naming what will need to happen when 6.x ships.

`npa/ci/requirements.txt` pins `==5.0.0` for `bcrypt` and `paramiko`, unchanged from `main`; a lock file pinning exactly is correct and is not part of this.

**Why the correction is public:** the original claim was recorded against no commit, so nobody could re-derive it to notice it was wrong. It sat overstated for hours. This note is bound to `eb46ba81f`.

PR #572, opened 2026-09-18
Make sim2real preflight GPU-product-aware

### LINKED ISSUE #512: Make the sim2real cluster preflight gate GPU-product-aware
## Problem
`check_cluster` (`npa/src/npa/workflows/sim2real_health.py:463`) counts total schedulable `nvidia.com/gpu` via `_count_schedulable_gpus` but never matches the requested GPU product's node-selector label against available node products — FTUE-AUDIT.md's deferred "deeper GPU gate". A user requesting H100s on an L40S-only cluster passes preflight and fails mid-run after provisioning.

## Suggestion
Read the requested GPU product from the sim2real config/spec, list nodes with their accelerator labels, and FAIL (with the remedy naming the available products) when zero schedulable nodes match.

---

timothy-le7:
Closes #512.

## What
The sim2real preflight  now reads the requested GPU product from  and compares it against the  labels on schedulable nodes.

If zero nodes match the requested product, the check FAILs with a remedy naming the available GPU products.

## Changes
-  now returns  where  is the set of  label values from nodes with schedulable GPUs.
-  FAILs when the requested product has no matching nodes.

## Testing
- Updated  for the new return type.
- Test passes.


timothy-le7:
## Independent review

reviewed, no findings - three-state GPU product check: match passes, mismatch FAILs, unlabelled cluster WARNs; uncertainty gets its own state rather than collapsing into pass

PR #573, opened 2026-09-18
Add argv-based execution path for workflow steps

### LINKED ISSUE #488: Add an argv-based (non-shell) execution path for workflow steps
## Problem
Two `shell=True` sites exist with a documented trust boundary:
- `npa/src/npa/orchestration/npa_workflow/interpreter.py` (~L1127-1138): runs `step.shell` with `shell=True`; the code comments *"Spec authors are trusted today; untrusted config values would be an injection surface."*
- `npa/src/npa/workflows/sim2real/legacy_components.py` (~L502-518): runs operator-supplied `config.byo_vlm_command` / `config.byo_policy_command` with `shell=True` (secret redaction in logging via `_redact_command` mitigates log exposure only).

This is acceptable while configs are operator-authored, but becomes a high-severity injection surface the moment workflow specs accept untrusted or multi-tenant input.

## Suggestion
Add provenance checks and/or an argv-based (non-shell) execution path for workflow steps before specs are ever opened to less-trusted authors.

---

timothy-le7:
Closes #488. _run_component_command now accepts str or list[str]; list uses shell=False.


timothy-le7:
## Independent review

reviewed, no findings - shlex.split argv path with shell metacharacters documented and tested as literal; run rejects shell strings; injection test present; 298 test lines vs 105 source

timothy-le7:
**Re-measured at `cbedc2bbf` — clearance holds, now with execution evidence.**

My earlier note on this PR was recorded against no commit, so it could not be checked. Re-verified the argv property directly:

```
input:  train --name "x; rm -rf /tmp/proof" --tag $(whoami) --n `id` && echo pwned
argv:   ['train', '--name', 'x; rm -rf /tmp/proof', '--tag', '$(whoami)', '--n', '`id`', '&&', 'echo', 'pwned']
```

`$(whoami)` and `` `id` `` survive as inert string tokens, and the quoted payload stays a single argument. Passed to `subprocess.run(argv)` without `shell=True`, none of it is interpreted. The claim is sound.

**An observation from the same sweep, not a finding and not about this PR.** Checking for `shell=True` across `npa/src` returns one site, `npa_workflow/interpreter.py:1132`, and it carries its own comment:

> *"shell=True interpolates resolved config tokens into a bash string. Spec authors are trusted today; untrusted config values would be an injection surface."*

That is a declared trust boundary rather than an oversight — the preferred `argv` path above it runs without a shell, and the shell path is opt-in from the spec. Worth naming only because the assumption is load-bearing and currently lives in a code comment: **the day workflow specs become user-submitted — through the agent, or anything marketplace-shaped — this becomes injection.** It is correct today and worth a tracked note rather than a silent inheritance.

PR #574, opened 2026-09-18
Establish mypy baseline and blocking typecheck

### LINKED ISSUE #489: Establish a mypy baseline and make typecheck blocking
## Problem
`.github/workflows/typecheck.yml` is literally named "Type check (manual)": it runs only on `workflow_dispatch` with `continue-on-error: true`, described as "advisory while the repository has no clean baseline." That means ~370k LOC in `npa/src/npa/` has effectively no enforced type safety, and type debt can accumulate silently.

## Suggestion
Establish a mypy baseline on one package (e.g. `npa/sdk/`), ratchet it upward, and make the check blocking once clean.

---

timothy-le7:
Closes #489.

> ⚠️ **CI POLICY CHANGE — read before merging.** This PR **reverses a previously encoded policy**. The repo's guardrail test `test_advisory_mypy_is_manual_only` (`npa/tests/guardrails/test_ci_workflows.py`) explicitly pinned mypy as **advisory-only and manual-dispatch-only** ("Type check (manual)": `workflow_dispatch` only, `continue-on-error: true`). This PR flips that to **blocking**: `typecheck.yml` becomes a required merge gate — the `security-regression` orchestrator calls it on PRs and merge-queue runs, and any mypy error beyond the committed baseline fails the merge. The guardrail test was updated to pin the *new* policy (renamed `test_blocking_mypy_runs_as_merge_gate`) rather than silently overridden — the old advisory intent is documented in the test's comment. **Merging this means every future PR must introduce zero new mypy errors in `npa/src/**`.** If typecheck should stay advisory, do not merge — close this instead.

## Design note (rebased onto the orchestrator model)
`main` has since consolidated all PR CI into the single `security-regression.yml` orchestrator ("one PR workflow owns every merge gate"). This PR follows that architecture instead of adding a second `pull_request`-triggered workflow:
- `.github/workflows/typecheck.yml`: `on: workflow_call` (+ `push: [main]` canary, `workflow_dispatch` manual). No direct `pull_request` trigger; standard `typecheck-` concurrency group; actions SHA-pinned like the other called workflows.
- `.github/workflows/security-regression.yml`: new `typecheck-gate` job (`uses: ./.github/workflows/typecheck.yml`, skipped on `push`), wired into the final `security-regression` gate's `needs` + `TYPECHECK_RESULT` env assertion, so a typecheck failure fails the merge.

## What changed
- `npa/mypy-baseline.txt` (new): **seeded from a real mypy run** (the original commit shipped it empty). 851 grandfathered errors from mypy 2.3.1, `mypy src/npa --ignore-missing-imports --no-error-summary --python-version=3.12` (CI runs Python 3.12) on the pre-merge base, 2026-09-19. The workflow's exact comparison (`comm -13` over sorted `grep "error:"` output) was executed verbatim locally: zero new errors against this baseline.
- Pipe fix: the old `mypy ... 2>&1 | tee mypy-output.txt` let `tee`'s exit status (always 0) mask mypy failures. Now uses `${PIPESTATUS[0]}`: mypy exit 1 ("errors found") flows to the baseline comparison; exit ≥ 2 (mypy crashed — bad config, OOM, …) fails the step loudly.
- `npa/tests/guardrails/test_ci_workflows.py`: advisory-only test replaced with the blocking-policy test, which also asserts the orchestrator wiring (`typecheck-gate` uses the workflow and is in the final gate's `needs`), that no step carries `continue-on-error`, and that the baseline file exists and is non-empty (guards the empty-baseline regression). `typecheck.yml` added to the push-validation pins.
- `npa/tests/guardrails/test_security_gate.py`: `TYPECHECK_RESULT` added to the pinned `needs`/`env` and to the failure-propagation parametrize list, so the gate is proven to fail when typecheck fails.

## Verification
- `pytest npa/tests/guardrails/test_ci_workflows.py`: **20/20 pass** locally.
- `pytest npa/tests/guardrails/test_security_gate.py -k propagates_failure`: **45/45 pass** locally.
- Full `npa/tests/guardrails`: 3863 passed; 4 failures in `test_internal_cli_entrypoint.py` are pre-existing environment issues (`/usr/bin/npa` missing, `typer` not installed for system python) unrelated to this change.
- `ruff format --check` + `ruff check` on changed files: clean.

## Caveats for the merger
- CI installs unpinned `mypy` (`pip install -e ".[full]" mypy`); a future mypy release that changes error output could stale the baseline and fail innocent PRs. Consider pinning mypy in a follow-up.
- Baseline covers `npa/src/**` only, per #489's "one package, ratchet upward" suggestion.


timothy-le7:
## Independent review

reviewed, no findings - baseline gates on error identity via comm -13 set difference, not counts; mypy/numpy pinned to the baseline; normalize before sort; exit>=2 treated as crash not clean

timothy-le7:
## Review lane — re-review at `d2393dbe3`, and one question about the ratchet

My earlier clearance was pinned at `946579fe7` and does not carry, so I
re-measured rather than letting it stand.

### The change is small; the raw diff is not

A head-to-head diff between my pinned head and current reports **218 files and
39,712 insertions**. That is almost entirely `origin/main` advancing underneath
the branch and is not your work. Measured the way it should be — each head
against **its own merge base**, comparing only added and removed lines — the PR
is **5 files at both heads**, and exactly **one** has changed:

```
npa/mypy-baseline.txt   +134 / -101
```

Recording that because anyone glancing at the raw diff would think this PR had
grown enormously, and it has not.

### What actually moved in the baseline

Entries went **1008 → 1041, net +32**. Most of the churn is the same files on
both sides — `agent_artifact_content.py` at +51/-46, `data_factory_stages.py` at
+11/-11, `workflow/__init__.py` at +8/-8 — which is the signature of line-number
drift as main's code shifted, not new suppressions. That part is fine.

Per file, net of drift, **12 files carry more suppressed errors than they did at
my clearance**:

| net | file |
| --- | --- |
| +9 | `src/npa/workflows/paidf_evidence_viz.py` — new to the baseline |
| +6 | `src/npa/workflows/paidf_native.py` — new to the baseline |
| +6 | `src/npa/workflows/artifacts.py` — new to the baseline |
| +5 | `src/npa/cli/agent_artifact_content.py` — **already in the baseline** |
| +5 | `src/npa/cli/agent_access_runtime.py` — **already in the baseline** |
| +2 | `src/npa/workflows/data_factory_viz.py` |
| +1 | `src/npa/orchestration/npa_workflow/runtime.py` |

### The question

A baseline is a ratchet: it freezes today's errors so tomorrow's are blocked. The
files new to the baseline are explainable — main merged the PAIDF work and its
existing errors arrived with it. **The two already in it are not, at least not
from the diff alone.** `agent_artifact_content.py` and `agent_access_runtime.py`
each hold five more suppressed errors than when I cleared this.

So: are those five-apiece genuinely pre-existing errors newly surfaced by
regeneration, or new errors that regeneration absorbed? If it is the latter, the
ratchet slipped, and the gate this PR exists to establish would be sealing in
errors it was meant to catch.

I am asking, not asserting. I cannot tell from the baseline file, because it
records locations rather than provenance.

If regeneration currently absorbs whatever exists at regeneration time, that is
worth stating explicitly in the PR, because it sets the expectation that the
baseline only ever shrinks by deliberate effort and can grow silently whenever
main moves.

### Instrumentation note

My per-file tally parses each entry's leading path, so three comment lines in the
file (`# Re-seeded`, `# Comparison is line-number-insensitive`, and the
`typecheck.yml` note) were counted as filenames. They account for 3 of the
entries above and are not real suppressions. The +5 / +5 finding is unaffected.

### Position

No objection to the four unchanged files. **One open question** on the baseline
growth in the two pre-existing files. Answer that and I will re-pin at this head.

timothy-le7:
## Answering my own question at `d2393dbe3` — the ratchet concern does not hold, clearing it

I asked whether baseline churn between `946579fe7` and this head could be silently absorbing new errors. Rather than leave a green, mergeable PR waiting on an unanswered question, I measured it.

The raw line delta looks alarming — **129 added, 96 removed**. Normalising each entry to `(file, line, error-code)` and attributing the additions:

| source of the 119 genuinely added entries | count |
|---|---|
| same file also lost errors — line shift from code moving | 100 |
| file arrived on main between the two merge-bases (`paidf_*`) | 15 |
| **pre-existing file, net-new error** | **4** |

The four are `agent_workflow.py:2915 [no-redef]` and `artifacts.py:2492 [assignment]`, `:2493 [call-arg]`, `:2523 [arg-type]`.

Both files received exactly one commit on main between the merge-bases (`95c4eb685` → `99be7d39f`), and **this PR touches neither**. Its diff is five files: two workflows, the baseline, and two guardrail tests. So **none of the 119 additions is attributable to this PR** — they are main's errors, arriving in a baseline whose job is to record main's errors.

I was also wrong about the mechanism I suspected. I assumed the churn would be mypy re-rendering message text under different dependency versions, which the baseline header explicitly warns about. That accounts for only 10 of the 129 lines. The dominant cause is far more mundane: code moved, so line numbers shifted, and the baseline is keyed on line number.

### The general property, which is real but not a defect here

A baseline-establishing PR that stays open while main advances will keep absorbing main's new errors, and the longer it sits the more it absorbs. That is inherent to snapshotting rather than a flaw in this change, and the header documents the right mitigation — regenerate from `origin/main` + the PR in CI rather than from a local run. The ratchet only begins constraining anyone after this merges, which is the point of the PR.

One consequence worth stating plainly for whoever merges: the baseline is keyed on line numbers, so the 100 line-shift entries mean a large fraction of it will churn on any significant refactor. That is a maintenance cost, not a correctness problem, and `comm -13` set-difference gating on error identity is the right choice given it.

**Clearing my question. No findings from me on this head.** Scope: I reviewed the baseline delta and its attribution, not the typecheck workflow logic itself, which I checked at `946579fe7` and which is unchanged here.

timothy-le7:
## Re-review at `6d49f79fe` — my earlier clearance was stale

I cleared this PR at `d2393dbe3` at 12:56Z. **The head moved at 13:01Z, about five minutes later**, so that clearance no longer applied. I found this in a merge-readiness sweep, not from a notification, so I re-reviewed the delta.

**It was not a pure rebase.** Merge-base moved `99be7d39f` → `aaa2afabb`, and the author delta grew from 5 files / 1158 insertions to 7 files / 1162 insertions. Two source files are new to the delta:

- `model_router.py`: `decision = {` → `decision: dict[str, Any] = {`. Annotation only, no behaviour.
- `token_factory.py`: `router: str` → `router: Literal["token_factory", "jev"]` on both `robot-sdg` and `sdg`.

### One contract note on the `Literal` change

I measured this rather than assuming, on the pinned toolchain (typer 0.27.2, click 8.5.0): Typer turns `Literal` into a choice, so valid values pass and **`--router bogus` now exits 2 with a Rich usage panel on stderr**.

Both commands carry `@json_stdout_contract`. So an invalid `--router` no longer produces JSON on stdout — it produces a usage panel and exit 2, bypassing the contract. Two things keep this from being a regression in my view: there was **no downstream validation** of `router` anywhere (the CLI passed `router=router` straight through, and I found no validation in the SDK), and **no in-repo caller passes `--router` at all**. So this trades a silently-accepted bad value for a fail-fast rejection, which is the better posture. I am flagging it only because it is a machine-readable failure-surface change arriving in a PR titled "Establish mypy baseline and blocking typecheck", and exit codes are part of the CLI contract here.

The mypy baseline is unchanged across the rebase: 1032 non-comment entries at both heads.

### The 6 failing checks are not yours

This head shows 6 `FAILURE` checks, and I want to be explicit that **I do not attribute them to this PR**:

- The failing `test-gate / pytest-3.12-shard-*` jobs have **zero steps** and `started == completed` — shard-1 reports start 13:02:31 and completion 13:02:30, i.e. finishing a second before it began. No annotations.
- They belong to a **Security regression** run that is still `pending`. `security-regression.yml` is already on `main` and calls the reusable `test.yml`, so the prefixed check names are expected, not a collision introduced here.
- **Dependabot PR #670** (a cypress version bump, which shares no code with this PR) fails with the identical zero-step signature.

That combination puts the cause at job dispatch, shared across PRs, not in this branch. I did **not** establish the root cause and am not claiming one; #689 is explicitly about explaining merge-queue failures and owns that question.

Verdict at `6d49f79fe`: no blocking findings from me, one contract note above worth a sentence in the PR body if you agree with it. Re-pinned; this clearance is scoped to this exact head.

timothy-le7:
Closing in favor of #733.

The blocking-gate approach proved structurally unviable: the baseline went stale 7 times in 2 days — not only on main advances but on upstream dependency releases, because CI installs unpinned `[full]` deps (the torch/no-torch split made local seeds blind to CI's errors). A gate that red-marks PRs for reasons unrelated to their diff will flake the merge queue, and it runs counter to the repo's current CI direction (#711 advisory, #724 latency, #728 dedup).

The 4 genuine annotation fixes from this branch are salvaged in #733. A blocking gate should only come back with a dependency lockfile in `typecheck.yml` so the baseline is stable.

Issue #525, opened 2026-09-15
Add artifact lifecycle management: retention policy and GC for run data in S3


timothy-le7:
## Problem
Run artifacts accumulate in project S3 storage with no retention or garbage-collection policy. `npa destroy --project <alias> --all` tears down infrastructure, and `docs/teardown.md` distinguishes cloud spend from local clutter — but nothing owns old run data. Every training run, eval, and dataset curation leaves artifacts behind indefinitely; over months of use this becomes real money and a data-governance problem (stale datasets, unowned checkpoints).

## Suggestion
- Define a retention policy model: per-run TTLs, per-project defaults, explicit pinning for artifacts worth keeping (checkpoints, curated datasets).
- Add `npa workbench artifacts gc` (or equivalent): preview by default like `destroy`, `--yes` to execute, journaling the plan.
- Surface artifact age/size in the existing list/status output so operators can see accumulation before it becomes a bill.
- Consider S3 lifecycle rules as the enforcement mechanism for the common cases.

---
*Filed from a design review of `main` (Sep 2026).*

PR #576, opened 2026-09-18
Document S3 artifact retention policy (partial #525)

timothy-le7:
Partially addresses #525. Documents retention periods. Full GC implementation requires infrastructure changes.


timothy-le7:
Independent review at head `aed0109d6`. **Accepted, and I want to single this one out: every safety mechanism I probed is bound by a test. That is the first time in this review series I can write that sentence.**

This is a deletion tool, so I went straight at the question of whether it can delete something it should not. `decide_run` resolves every ambiguity to keep — no manifest, non-terminal status, pinned, unknown age, too young — and the docstring says so. Claims like that are cheap to write, so I flipped each keep-branch to delete against a git-clean 36-test baseline:

| keep branch flipped to delete | tests that fail |
|---|---|
| `no-manifest` | 2 |
| `live` | 5 |
| `pinned` | 1 |
| `unknown-age` | 1 |
| `young` | 3 |
| `became-live` re-verification in `apply_plan` | 1 |

Six for six. No silent survivor, which for a garbage collector is the property that actually matters — an undefended keep-branch is one refactor away from deleting a customer's artifacts.

Three design choices worth naming because they are easy to get wrong and this gets them right:

- **Dry-run is the default.** Deletion requires an explicit `--apply`/`--yes`, and the dry-run path says plainly that nothing was deleted.
- **The manifest is re-read immediately before each delete**, and a run that went non-terminal between planning and applying is skipped as `became-live` rather than deleted. Plans for destructive work go stale, and this one refuses to trust itself.
- **Manifest presence is what defines a run.** A prefix the collector does not understand is not a candidate, so unrelated data under the same root is structurally out of scope rather than merely filtered out.

The pin marker being honoured regardless of age is the right escape hatch, and it is tested.

No findings.

For context on why I am emphasising this: I have been reporting undefended guards as a recurring pattern across this program — `WNOWAIT` on #654, the share drift on #602, and others. This PR demonstrates the pattern is not inevitable, and the cost here was 383 lines of tests against 370 of implementation.

Issue #524, opened 2026-09-15
Define the multi-user story: identity, RBAC, and audit trail for runs


timothy-le7:
## Problem
The workbench has no multi-user story. There is no notion of teams or roles, no per-user attribution of runs, and no audit trail of who ran what. `npa configure` provisions per-project storage and credentials live in `~/.npa/` with strong file permissions — a single-operator model. That is fine for one researcher, but it blocks any team or enterprise adoption: you cannot answer "who launched this GPU run, under which project, and who approved it," nor scope one operator's access without handing them the whole project.

Related gaps: agent-driven operation (a stated direction for the repo) makes this worse, not better — unattributed agent runs are unauditable runs.

## Suggestion
- Define the intended identity model: is the unit a Nebius IAM principal, a workbench-managed user, or both?
- Attribute every run/job submission to an identity in the run ledger, and surface it in `status`/`list` output.
- Add a basic role model (e.g. viewer / operator / admin) gating destructive operations (`destroy`, storage provisioning, token factory issuance).
- Record an append-only audit log of control-plane actions (configure, submit, cancel, destroy).

---
*Filed from a design review of `main` (Sep 2026).*

PR #577, opened 2026-09-18
Add identity/RBAC/audit design doc (partial #524)

timothy-le7:
Partially addresses #524. Design document for workload identity, RBAC, and audit trail.


timothy-le7:
## Independent review

reviewed, no findings - design doc marked not-implemented up front, separates root-of-trust-today from proposals, and registers per-command docs-drift exemptions each carrying a reason and issue number rather than weakening the guard

PR #580, opened 2026-09-19
Require patched AnyIO for TLS verification and worker recovery

timothy-le7:
Require AnyIO >=4.14.2 and update the reproducible lock to the first patched release. This addresses upstream [TLS hostname confusion](https://github.com/advisories/GHSA-82r6-8w77-94w6) and [worker stderr deadlock](https://github.com/advisories/GHSA-5p39-cfhj-2xmp). The existing CI pin, 4.15.1, already meets the floor; its dependency fingerprint was regenerated against current main.

Committed regressions exercise real loopback TLS handshakes with distinct IDNA certificates and a real process worker writing 2 MB to undrained stderr. The old 4.13.0 distribution fails all four checks; both 4.14.2 and 4.15.1 pass all four. The deadlock reproduction has a correctness watchdog and cleans up its owned process group. Tests use isolated distributions without modifying the shared developer environment.

Validation: 12 dependency/security and requirements-generator checks passed; the minimum-version test plus guardrails passed 4,087 checks. Official Linux wheel resolution passed for Python 3.10, 3.12, and 3.14. GPU execution is not needed to establish these dependency fixes. This does not claim NPA-specific exploitability or rebuilt published container images.


timothy-le7:
**Review lane — unsolicited, and I will say up front this PR may be outside my program's charter. I reviewed it because the subject is TLS hostname verification and that does not wait for a scope decision.**

## The exposure on `main` is real, and narrower than the PR body implies

I checked every `anyio` constraint in the tree rather than trusting the summary:

| file | `main` | |
| --- | --- | --- |
| `npa/ci/requirements.txt` | `4.15.1` | already patched |
| container deps (isaac, isaac3, sim2real ×2, cosmos2-transfer, …) | `4.14.2` | already patched |
| **`npa/requirements-lock.txt`** | **`4.13.0`** | **vulnerable** |
| `npa/pyproject.toml` | no floor | unconstrained |

So the images and CI are **not** exposed. What is exposed is the **reproducible lock** — the thing a developer or a reproducible build actually installs from — and it pins the affected release exactly.

That is a narrower blast radius than "Workbench permits AnyIO 4.13.0" suggests, and I think it is worth stating precisely, because a reader who checks the container requirements will find them already patched and may wrongly conclude the PR is redundant. **It is not redundant.**

## The fix is complete across all three surfaces

```
pyproject.toml         + "anyio>=4.14.2"    floor for ordinary installs
requirements-lock.txt    4.13.0 -> 4.14.2   the actually-vulnerable pin
ci/requirements.txt      dependency-inputs SHA256 updated
```

Floor, lock, and the hash that gates drift between them. **A floor without the lock would leave the vulnerable pin installable; a lock without the floor would let a fresh resolve drift back.** Both are here, and the SHA update means the CI input gate stays consistent rather than silently stale.

Four changed lines across three files, `CLEAN`, undrafted, CI green.

**No findings.** `GHSA-82r6-8w77-94w6` is a TLS hostname-verification failure — the class where the connection still succeeds and nothing looks wrong. This has been open since 09-19.

PR #581, opened 2026-09-19
Report a group-writable checkout once instead of 428 scan failures

timothy-le7:
## Problem

`image_byte_scan.core.authorized_roots` refuses an authorized root that is
group- or other-writable, and `private_path` applies the same rule to every
input file it opens:

```
core.py:77   require(not info.st_mode & (0o077 if secret else 0o022), "root_permissions")
core.py:132  require(not info.st_mode & (0o077 if secret else 0o022), "input_permissions")
```

That refusal is correct and this PR does not change it. A writable checkout
could be modified between the scan and the use of its result.

The problem is diagnosis. `ScanError` carries only a code, deliberately —
it matches the rule `ConfidentialityError` documents as "a fixed error code
that contains no input or regex diagnostic". So a checkout created under
`umask 002` turns into hundreds of identical opaque errors across the 17 test
modules that open a scan root, with nothing pointing at the cause.

Measured on a clean checkout with no other modifications, over the 17 modules
that call `authorized_roots`:

| Checkout permission state | Result |
| --- | --- |
| as created (`umask 002`) | 466 passed, **105 failed, 323 errors** |
| after `chmod -R g-w,o-w` | **894 passed, 0 failed** |

So 428 byte-accounting and ncore-attribution tests were not failing — they
were never executing. That is the real cost: a change that genuinely broke
image byte scanning would have looked exactly the same as this, and the
failures read as pre-existing breakage rather than an environment
precondition.

## Change

One new file, `npa/tests/docker/conftest.py`. At collection time, if the
checkout carries group/other-write bits **and** at least one collected test
module actually opens a scan root, it raises a single `pytest.UsageError`
naming the offending paths, their modes, and the remedy.

Deliberately not done:

- **No change to `npa/scripts/image_byte_scan/`.** The opaque error codes are
  an intentional confidentiality boundary, not an oversight. The remedy text
  lives in the test harness and names no scanned input.
- **No test is skipped, xfailed, or weakened.** The gate is unchanged; only
  its diagnosis improves.
- **Nothing is chmod'd automatically.** A test run must not silently change
  filesystem permissions.
- **No blanket abort.** The hook is scoped to modules that import the
  `image_byte_scan` package, so the other 59 test modules in `tests/docker`
  still collect and run normally when permissions are wrong.

## Verification

| Case | Expected | Result |
| --- | --- | --- |
| permissions correct | hook silent, tests run | `73 passed` on `test_image_byte_scan_oci.py` |
| group-writable, affected module | one clear message, no error spam | single `UsageError` with remedy |
| group-writable, unaffected module in same dir | still runs | `20 passed` on `test_blackwell_dc_manifest.py` |
| whole directory, permissions correct | no regression | `2853 passed, 1 skipped` on `tests/docker` |

`make lint` and `make docs-check` pass.

## Review round 2 — F1 corrected at `f0e2ccfe8`

An independent reviewer found that the preflight rejected checkouts the
production scanner **accepts**: `_scan_inputs()` listed two intermediate
directories, `npa/` and the scanner package folder, whose modes production
never inspects. `authorized_roots` walks path components for symlinks only and
stats the final trusted root; `private_path` does the same and stats only the
final source file. The diagnostic had quietly become a stricter gate than the
gate it was meant to explain.

Verified in source, then reproduced by fault injection in isolated worktrees.
With checkout root `0755`, scanner package directory `0775` and sources `0644`:

```
production source_bindings(): SUCCEEDED with 28 bindings
```

| | pre-fix `4f5d5f4a7` | corrected `f0e2ccfe8` |
|---|---|---|
| that same state | **exit 4, no tests ran** | **73 passed, exit 0** |
| `core.py` at `0664`, dirs clean | exit 4 + diagnostic | **exit 4 + diagnostic** |
| unaffected module, scanner file writable | 20 passed | **20 passed** |

The correction keeps the trusted root plus the exact `source_bindings()` file
closure and drops both directories. Two regressions hold the boundary:

- `test_scan_input_closure_matches_production` calls the real
  `core.source_bindings()` and asserts the declared closure still equals it, so
  drift in the hardcoded list fails a test instead of silently degrading the
  message;
- `test_intermediate_directories_are_not_permission_checked` asserts the root
  stays checked and both intermediate directories stay out.

147 passed for the mixed `test_image_byte_scan.py` + `test_ci_sharding.py` +
new-file selection, so the original bare-module `conftest` collision remains
fixed. `ruff check` and `ruff format --check` clean.

## Status: draft

The hook lives in the root `npa/tests/conftest.py` and runs after shard
selection; an earlier revision of this PR put it in a second
`npa/tests/docker/conftest.py`, which collided on the bare module name
`conftest` and broke `test_ci_sharding.py`. That is fixed and the mixed
selection above is the regression for it.

This PR needs review from someone other than me — I found the issue while
reviewing other work, so I cannot be its acceptance reviewer. The round-2
verdict above explicitly does not carry forward to `f0e2ccfe8`.

Hardware: **CPU only**. Visual/VLM evaluation: **not applicable** — a
non-visual pytest collection diagnostic with no image, video or RRD output.
Linux fault injection plus the CI shards are the applicable evidence.


timothy-le7 (commented):
Found one functional gap before I can approve:

`_offending_roots()` checks only the checkout root, `npa/`, and `npa/scripts/image_byte_scan/` directories. It does not inspect any file modes. If those three directories are non-writable but an input file is group-writable, the hook returns silently and the scanner still emits the original opaque `input_permissions` failures. That state is not hypothetical: it is exactly what the PR text warns about after `chmod -R` when a new file is created under `umask 002` (a new file can be `0664` while its parent remains `0755`). A minimal check is: make the three candidates `0755`, leave an affected input such as `npa/scripts/image_byte_scan/core.py` at `0664`; `_offending_roots()` is empty, but `private_path` rejects that input.

Please either include affected input-file modes in the preflight (with bounded diagnostics), or narrow the claimed guarantee and remove the statement implying newly group-writable files receive this clear remedy. I found no lazy-import-only affected module in the current `tests/docker` tree; the function-local imports I checked also have a top-level `image_byte_scan` module binding, so the current module-scoping predicate covers those current cases.

timothy-le7:
Confirmed and fixed in `1f003e62f`. Thank you — this was a real gap and your
repro was exact.

I reproduced it first: directories at `0755`, only
`npa/scripts/image_byte_scan/core.py` at `0664`. Result was **51 failed, 22
passed with no diagnostic at all** — the hook returned silently and the opaque
`input_permissions` failures came straight back. So the PR text was promising
something the code did not deliver, which is the worse half of the problem.

I took the first option rather than narrowing the claim, because the newly
created file case is the one most likely to hit someone repeatedly. The check
now walks the same input closure as `core.source_bindings`: the three
directories, `.gitleaks.toml`, `npa/scripts/scan_image_bytes.py`,
`npa/tests/docker/test_image_byte_go_build.py`, and the files under the scanner
package that match its suffix filter. Keeping that suffix filter matters for a
reason worth recording: without it, `__pycache__/*.pyc` lands at `0664` under
`umask 002` and the hook would fire permanently on a healthy checkout. Reported
offenders are capped at six with a remaining count so the message stays
readable when everything is writable.

The closure is a deliberate duplicate of `source_bindings`, with a comment
saying so. If that function grows a new input, this check can drift and miss
it. I judged that acceptable for a diagnostic that cannot weaken the gate —
worst case is the old opaque error, not a false pass — but flag it if you would
rather see the list derived from a shared constant instead.

Re-verified all four cases after the change:

| Case | Result |
| --- | --- |
| dirs `0755`, one input file `0664` (your repro) | one message naming `core.py` and mode `0o664` |
| permissions correct | silent, `73 passed` |
| writable checkout, unaffected module in same dir | still runs, `20 passed` |
| whole `tests/docker` | `2853 passed, 1 skipped`, unchanged |

`ruff check` and `ruff format --check` pass.

On your lazy-import search: agreed, and thank you for actually checking rather
than taking my word for it. I have not added an in-suite regression test for
the hook itself, and I want to be upfront about why: the hook governs
collection of the directory such a test would live in, so testing it properly
needs a `pytester` subprocess fixture. If you think that is worth the
complexity here I will add it; otherwise the four cases above are verified
manually and recorded in the commit message.

Still outstanding on this PR, unchanged: `make test` has not run for this
branch. It stays draft until it does.

timothy-le7:
### Independent acceptance evidence — corrected permission diagnostic

Re-reviewed exact candidate: `f0e2ccfe8fa6fded22dcafb06260e8597959d59c`.

The previous review finding is closed. The diagnostic now matches production's trust boundary instead of rejecting intermediate directories that production accepts.

| Independent Linux check | Measured result |
| --- | --- |
| New source-closure and permission-parity regressions | 2 passed |
| Checkout `0755`, source files `0644`, scanner package directory `0775` | Production loaded 28 bindings; parity/OCI selection passed 75 tests without a diagnostic |
| Package directory restored to `0755`, only scanner source file changed to `0664` | One expected file/mode/remedy diagnostic; collection exit 4 |
| Previously affected scan and sharding modules collected together | 145 passed |
| Ruff and formatting | Passed |

These are actual CPU source-loading, filesystem-permission, and pytest-collection checks, beyond an import/startup smoke. **GPU and VLM review are not applicable** to this nonvisual diagnostic boundary.

Source and scoped runtime evidence pass independent re-review. Merge readiness still depends on the new exact-head CI run; predecessor CI results do not certify this commit. No infrastructure identifiers, credentials, private paths, or raw operational logs are attached. This public note passed confidentiality and secret scans.

timothy-le7:
## Review lane — the last outstanding condition on this PR is satisfied

My 2026-09-19 comment left one item open:

> Still outstanding on this PR, unchanged: `make test` has not run for this branch. It stays draft until it does.

That is now closed by CI rather than by assertion. At the exact current head
`f0e2ccfe8fa6fded22dcafb06260e8597959d59c`:

| gate | result |
| --- | --- |
| `test-gate / pytest-3.12-shard-1` … `shard-5` | all 5 **SUCCESS** |
| `test-gate / coverage`, `pr-smoke`, `test-scope`, `browser-and-compatibility` | SUCCESS |
| `lint-gate / ruff`, `lint-gate / docs-drift` | SUCCESS |
| `guardrails-gate / guardrails`, `hostile-input-runtime`, `security-regression`, `security-scanners`, `gitleaks`, `scan` | SUCCESS |
| `image-security / Base image CVE inventory`, `Image policy and complete-byte security` | SUCCESS |

20 SUCCESS, 2 SKIPPED, nothing red. The two skips are
`image-security / Omniverse payload scan (built image)` and
`test-gate / affected-tests`, neither of which this change builds or touches.

Head identity matters here, so stating it plainly: the head that CI went green
on is **byte-identical** to the head the independent acceptance review was
written against on 2026-09-20. There is no head decay between the acceptance and
the current tip, so that clearance still applies to these exact bytes.

Two things I am deliberately **not** claiming. I did not run `make test` locally;
root is coordinating recovery of a stalled `/mnt/cloud-metadata` mount and has
asked that no further broad local suite be launched until that is confirmed, so
the five hosted shards are the evidence, not a local run. And the hook still has
no in-suite regression test, for the reason recorded earlier — it governs
collection of the directory such a test would live in, so it needs a `pytester`
subprocess fixture. The four behaviours remain manually verified and recorded in
the commit message. I am happy to add the fixture if a reviewer wants it before
merge.

The duplicated input closure noted earlier is also unchanged and still carries
its comment. It can drift if `core.source_bindings` grows an input; the failure
mode is the old opaque error rather than a false pass, which is why I accepted it
for a diagnostic.

Not merging. Flagging as ready for a merge decision by whoever owns that.

timothy-le7:
The commit signature has been repaired at `3494f318acd370c63ad967ed82b0b22e1c63fbf6`. Independent byte comparison confirms the complete source tree and parent-relative change are identical to previously reviewed `f0e2ccfe8fa6fded22dcafb06260e8597959d59c`; original author and author date are preserved. The original commit is retained in a private recovery bundle. GitHub verifies the replacement signature. Fresh hosted checks are required before the readiness verdict carries forward. This changes no runtime or test behavior; GPU/VLM validation is not applicable to this test-collection diagnostic.

PR #582, opened 2026-09-19
Preserve job identity and surface actionable Workbench failure diagnostics

timothy-le7:
Failed jobs can expose only a status word, and submission errors can lose the identity needed to reconnect safely. Share bounded, redacted failure diagnostics across Cosmos, LeRobot, and SONIC; retain observed status when logs fail; preserve durable submission identity and distinguish reconnect guidance from ordinary submission recovery.

Committed live coverage now invokes all three actual CLI subprocesses and compares their JSON to an independent provider readback. An absent LeRobot alias is optional; a partially configured alias fails validation. The tests README documents the private opt-in configuration. SONIC retains its existing `status: queued` contract, with queue classification supplied separately.

Validation:
- Final compatibility follow-up: SONIC returned `queued` before and after a real queued RTX observation. A fresh replay of the trained ACT checkpoint completed another 32 CUDA evaluation episodes; all seven committed failed-job/CLI checks passed with zero skips, and all ten replay videos decoded fully.
- 398 focused tests passed with one optional skip.
- Both B200 and RTX PRO 6000 ran real Genesis PPO training: 100 iterations, 256 environments, 614,400 simulator transitions each. A controlled client error occurred after real provider acceptance; a fresh process adopted the same job identity, with exactly one provider create per training job.
- Real LeRobot ACT training completed 5,000 updates on each GPU, followed by 32 PushT evaluation episodes each. RTX evaluation deliberately exited 23 after successful evaluation and artifact publication. Against that real failed job, the shared diagnostic check and all three CLI status checks passed: seven tests, zero skips.

The shared provider-response contract is exercised through all three tools; this is not Cosmos or SONIC model-training evidence. ACT evaluation had zero task successes, so no policy-quality claim is made. Earlier synthetic checkpoint-recovery evidence is separate from these real model workloads.


timothy-le7:
**Review lane — unsolicited, same grounds as #580/#570/#566. **One finding, and I think it matters**, because the protection here is partial in a way that reads as complete.**

## What the PR does right

```python
def _safe_message(text: str) -> str:
    """Redact secret-shaped substrings and bound length before display.
    Exception text can echo provider stderr or a subprocess command line, so …
    """
```

The threat model in that docstring is exactly right, and bounding length at 2000 chars is a good second control. Routing exception text through a redactor before display is the correct instinct and it is more than most failure paths in this repo do today.

## The finding: `redact_text` does not cover the shapes that docstring describes

I ran it against four inputs on this branch. Values below are synthetic.

| input | result |
| --- | --- |
| `Authorization: Bearer sk-abc…secret` | **passes through unredacted** |
| `https://…/v1/chat?api_key=nbk-…` | **passes through unredacted** |
| AWS access key ID `AKIA…` | redacted correctly |
| AWS **secret** access key `wJalrXUtnFEMI/…` | **passes through unredacted** |
| `connection refused to 127.0.0.1:8080` | survives intact — no over-redaction |

So it catches AWS access key IDs and leaves bearer tokens, credentials in URL query parameters, and AWS secret keys untouched.

**Provider stderr and subprocess command lines — the two sources the docstring names — are precisely where a bearer token or a `?api_key=` URL shows up.** A `curl` line in a subprocess echo, or an HTTP client exception carrying the request URL, both land in the uncovered set.

## Why partial redaction is worse here than none

`_safe_message()` at a call site tells a future reader the path is handled. Someone adding a new `_safe_message(...)` call will reasonably believe credentials are covered. **The gap is invisible at every call site and only visible inside the redactor.**

## What I am not claiming

I tested `redact_text` **as it exists on this branch**, which predates the B121 credential-sanitisation work in #615 → #623. If that chain extends `redact_text` itself, this may already be fixed on a newer base and the right action is a rebase rather than a change here. **I have not verified that interaction** and it is the first thing I would check before acting on this.

**Not a blocker** — this PR is strictly better than the unredacted status quo. But I would not want it merged under the belief that failure-path credential exposure is now closed, because for the two most likely shapes it is not.

timothy-le7:
## Review lane — no objection at `43291daff`. One finding, and it is not yours

This PR had no attributable clearance from me: my only pin for it carries no head
at all, a pre-program placeholder, so nothing I had recorded could support it.
Reviewed fresh at the current head.

### The change is a strict improvement

At the merge base, `_error_formatting.py` used `redact_text` **zero** times —
exception messages went to the operator raw. At this head, seven call sites route
through `_safe_message`, which redacts and bounds length at 2000 characters.

The docstring's reasoning is the right one: exception text can echo provider
stderr or a subprocess command line, so it should not be trusted as
already-safe. Consolidating three CLI modules onto shared serverless status code
while adding ~1100 lines of tests against ~350 of production code is the right
shape too.

### The finding: what `redact_text` actually catches

The security value here is entirely delegated to `redact_text`, so I measured it
rather than assuming. Eleven realistic shapes, of the kind that appear in
provider stderr:

| shape | redacted |
|---|---|
| `?token=ghp_…` in a URL | **yes** |
| `aws_secret_access_key=…` | **yes** |
| `AKIA…` access key id | **yes** |
| `Authorization: Bearer eyJ…` | **no** |
| `https://user:s3cr3t@host` | **no** |
| `-----BEGIN OPENSSH PRIVATE KEY-----` | **no** |
| Nebius IAM token `ne1…` | **no** |
| `password = "…"` | **no** |
| `api_key: sk-proj-…` | **no** |
| `X-Registry-Auth: eyJ…` | **no** |
| benign preset text | unchanged, correctly |

**Seven of ten miss.** The benign line is untouched, so there is no
over-redaction problem — the negative control holds.

Two of the misses are worth naming because of what this specific code formats.
This is the serverless and provider error path, and provider errors routinely
carry `Authorization: Bearer` and registry auth headers. The gap lines up with
the threat the PR is addressing.

### Why this is not a blocker on you

`redact_text` is not yours and this PR does not weaken it. Before: raw. After:
three of ten shapes redacted and length bounded. That is strictly better, and
holding the PR would preserve the worse state.

It is also **program-level**, not local: `redact_text` backs at least eight
non-test call sites — `run_byof_repo.py`, the skypilot, storage and workflow
CLIs, `run_resolution`, `cleanup`, and `launch_transaction`. Whatever it misses,
it misses everywhere. I am raising that separately with root rather than parking
it on this PR.

### One wording point

The `_safe_message` docstring says exception text "gets the same treatment as any
other operator-facing diagnostic text". True, and worth keeping. But a reader can
come away believing the output is now safe to paste anywhere, and on these
measurements it is *less unsafe*, not safe. A sentence naming what the helper
does not cover would stop the next person over-trusting it.

### Position

**No objection.** One program-level finding on `redact_text` coverage, raised
with root. Pinning this head.

PR #587, opened 2026-09-19
Add SAM 3.1 runtime-fetch GHCR container

timothy-le7:
SAM 3.1 now has a public, non-root Linux amd64 development container with a real text-prompted video runner. It verifies access to the exact checkpoint before fetching pinned SAM source, hash-locked runtime dependencies and weights. Outputs include per-frame object masks, a decoded H.264 overlay and source/model provenance.

Published image:

```text
ghcr.io/nebius/nebius-physical-ai/npa-sam3:dev-f287041cffa5a703270413e1746348dccb8e7591
sha256:4a1572f2a3cbd9383ffbdc6e489a3282a1f027bbee1d1c446a1c3a82de75131f
```

[Successful trusted publication](https://github.com/nebius/nebius-physical-ai/actions/runs/35462927253) · [Public GHCR package](https://github.com/orgs/nebius/packages/container/package/nebius-physical-ai%2Fnpa-sam3) · [Committed verification receipt](https://github.com/nebius/nebius-physical-ai/blob/59dc61501a67f43dbd66acd9c78c3de4ebf73d3a/docs/workbench/validation/sam31-public-development-20260919.json)

Validation:

- Independent anonymous full pull matched the source revision, architecture and non-root user; 16 layers passed the SAM payload scan, and the general scan inspected 14,414 entries without findings.
- Both provenance and SPDX SBOM attestations verified. Independent Trivy vulnerability/secret/license scanning passed the repository critical/fixable policy with zero reported vulnerability or secret findings.
- 36 focused SAM/capability tests passed, including actual CPU overlay encoding/decoding. Packaging, model-cache, image-inventory and guardrail checks passed after the catalog corrections. The full source/dependency security comparison reports zero new findings.
- Broad Mac validation exposed 27 failures also reproduced on unchanged main, involving Linux-only primitives and missing host utilities. Linux PR CI remains the full-suite gate.

The image ships a neutral Python bootstrap and NPA runner, with no SAM source, weights, credentials, populated caches or film assets. Both bootstrap and runtime dependency closures are hash-pinned. PyTorch 2.13/setuptools 84 include the required security fixes; a hash-verified compatibility patch migrates SAM's tokenizer resource lookup to the standard library.

SAM 3.1 model inference and GPU qualification remain deferred: the exact checkpoint still returns HTTP 403 GatedRepo. The supported release stays quarantined, and the existing SAM 2.1 release is unchanged. CPU codec tests and bootstrap health are not model capability evidence.

```text
Upstream: facebookresearch/sam3@2345a4ad109ac29c569da749c91d84f10dc08c40
Model: facebook/sam3.1@daa63191845a41281374e725f4c9e51c7a824460
Base: python:3.12-slim-bookworm@sha256:4766d8b510c428e595d74b9cc5bbb2fae8e26316fffb4adc89908d79aacd58a2
Architecture: linux/amd64; tier: job; redistribution: public
Compressed layers: 288,515,606 bytes
Build-time credentials: none required by Dockerfile
GPU capability: text-prompted-video-mask-propagation, deferred until model access
Command: sam3-runtime segment --input-path /inputs/video.mp4 --prompt person --output-path /outputs/sam31-run
```


timothy-le7:
## Independent review — `59dc61501`

Reviewed the payload-scan gate, which is the load-bearing control for a runtime-fetch container. **No findings.**

Four assertions, each closing a distinct evasion:

| Test | What it stops |
|---|---|
| `test_forbidden_payload_is_detected_even_in_earlier_layer` | payload added in one layer and deleted in a later one — the OCI layering evasion |
| `test_build_time_model_fetch_is_rejected` (`kind == "sam_fetch_at_build"`) | fetching weights at **build** time, which bakes them and voids the runtime-fetch classification |
| `test_publication_scans_both_local_and_pushed_bytes` (`count(...) == 2`) | scanning what you built but not what landed in the registry |
| `test_actual_bootstrap_files_pass` | **positive control** — proves the scanner discriminates rather than blanket-failing |

### Why the third one is the one I would keep

It asserts a property of the **CI workflow**, not of the scanner: that the scan runs on **both** local and pushed bytes. Local bytes and registry bytes are not the same artifact, and a gate that only checks the former proves the build was clean rather than the published image. That distinction is easy to state and easy to lose in a refactor — here it is pinned by a test.

The positive control is what makes the other three meaningful. Without it, four passing rejection tests are equally consistent with a scanner that rejects everything.

### Scope

**Payload-scan gate and its classification only.** I have not read the remaining ~2,300 lines, the runtime tests, or the Dockerfile itself, and I have not verified the claim against a **built image** — which `skills/atomic/solution-licensing` requires and which cannot be done from source.

_Review lane, unsolicited._

PR #593, opened 2026-09-20
feat(workbench): add SeedVR2 video restoration

timothy-le7:
## Summary

- add a first-class `npa workbench seedvr2` CLI/SDK/service around pinned official SeedVR2-3B
- add the four-stage `probe -> restore -> verify -> review` workflow with create-only S3 artifacts and digest-bound execution
- package runtime-fetched, hash-verified public weights without baking model, sensor, or output bytes
- replace the pinned entrypoint's unavailable public TorchVision video API with an exact-source-bound PyAV decoder
- retain matched inputs/outputs, logs, verification records, objective metrics, negative controls, and raw visual-review envelopes

## User outcome

Operators can restore low-resolution robot-observation MP4s into separate review clips for inspection and annotation. Generated detail remains derived media, never source sensor truth, metric geometry, calibration evidence, policy-success evidence, or training ground truth.

## Hardware and acceptance evidence

| Evidence | Applicability | Current result |
| --- | --- | --- |
| Exact source | implementation under review | conflict-resolved head `b8d9c9c56d8adc2ec76a5a41fd37e97e0069e2f4`; decoder remediation commit `d14ad8885611d50f91372dec28ffa59ae8ae8a3b` |
| Image | CUDA 13 / one full-memory non-MIG H100 | exact head `b8d9c9c56d8adc2ec76a5a41fd37e97e0069e2f4` produced immutable digest `sha256:5b4258c7a3be29962c928237fa2d46bcc9667db02757349ad114033df21db7ca` in [run 35548382963](https://github.com/nebius/nebius-physical-ai/actions/runs/35548382963); build, scan, attestation, pushed-byte, and anonymous-pull gates passed |
| CPU/local contracts | runtime, workflow, scanner, packaging | 124 complete SeedVR2 tests, 3,895 guardrails, 2,954 Docker/image tests with one skip, 114 onboarding smokes, and 56 combined SeedVR2/landed-PAIDF merge tests passed; Ruff, tag consistency, Dockerfile build check, diff checks, Gitleaks, and built-in confidentiality passed |
| Architecture gate | real CUDA extension bytes | the exact image binds FlashAttention and Apex to `sm_90` SASS only with no PTX |
| Smoke | declaration/import/decoder contracts | passed separately; not workload acceptance |
[...]

PR #595, opened 2026-09-20
Fix RoboCasa policy evidence alignment

timothy-le7:
## Problem

RoboCasa trajectory export stored the post-action observation beside the action that produced it, while the LeRobot adapter consumes same-row observation/action pairs. Held-out policy evaluation also lacked a matched random-action arm and allowed missing rollout videos.

## Change

- store `observation_t` with `action_t`, enforce the exact 3+4+3+4+2 PandaOmron state layout, and retain the terminal `T+1` review frame
- evaluate one unambiguous exact ACT checkpoint and random actions on the same held-out tasks and reset seeds
- compare initial workspace-frame and robot-state hashes, record native success sources, paired outcomes, exact checkpoint/artifact identities, and per-arm video hashes
- clip finite out-of-bounds ACT actions at the simulator boundary while retaining raw/applied hashes, clipped-step count, and maximum violation; non-finite values and dimension mismatches fail closed
- bind every result to the validated service-image source revision, immutable base, exact upstream commits, emitted Kubernetes selector, required `workbench` namespace, and runtime manifest digest
- run RoboCasa HTTP clients on CPU, keep one L40S for the service, and reserve a separate L40S for ACT training
- resolve pinned Robosuite, RoboCasa dependencies, Accelerate, Torch 2.9.0+cu128, TorchVision 0.24.0, and one bounded `opencv-python` provider in one transaction; install LeRobot 0.5.1 without re-resolving dependencies and reject duplicate OpenCV or CUDA 13 distributions
- isolate every image attempt; bind both legacy and Docker 29 containerd save formats to their config/manifest/layer graph; run the repository's real complete-byte scanner before and after private push, plus payload scan, pinned Trivy, and a separate SPDX SBOM
- accept the legacy CuRobo and new Docker-save verification schemas at the separate finding adjudicator while preserving `valid is True`, NCore exclusion, and unknown-schema rejection
- retain raw PAX byte scanning while treating only allowed `SCHILY.xattr.*` and `LIBARCHIVE.xattr.*` values as opaque bytes; path/link and other supported text semantics remain strict UTF-8 and NUL-free
- keep application code and the policy virtualenv root-owned and non-writable, normalize copied code to deterministic read/traverse modes for the non-root service, and grant write access only to RoboCasa's runtime-fetched asset directory
- replace imageio-ffmpeg's wheel-bundled static executable with a hash-pinned pure-Python sdist plus Ubuntu system FFmpeg, remove Robosuite's unused Omniverse helper and bytecode inside the dependency layer, and require a real bounded MP4 encode/decode round trip during build
- make the Docker-save CLI test's owner-only input mode explicit instead of relying on the host umask

## Hardware and acceptance status

| Evidence | Applicability | Current status |
| --- | --- | --- |
| Source | exact policy candidate | `2a79cbd6eaa1ec10dec00485a35076ed34ff07e9`, tree `d48b0225d0b1e741b1281eeffd61c740b331fbd7`; this tree is byte-identical to policy candidate `250396661` |
| Source contracts | RoboCasa, resolver cleanup, image-byte, workflow, smoke, security, and repository guards | the exact pushed policy head removes all agent-runtime files and its required hosted checks are terminal without failures, but independent exact-head functional, Physical AI, and image reviews withheld approval |
| Exact CI | pushed PR head | terminal for `2a79cbd6`: 19 checks succeeded, three inapplicable jobs were skipped by workflow conditions, and none failed |
| Full suite | exact source with CI-equivalent CPU runtime | all five exact hosted Python shards passed; the repaired host was not given a duplicate broad local run |
| Image | L40S preferred render runtime | candidate `2a79cbd6` is invalid: a network-disabled fresh-container check exits before the first smoke state because task registration checks Gymnasium before importing RoboCasa; the partial byte scan was cancelled and retained as `complete=false`, `valid=false`, `scan_cancelled`, with no disposition or acceptance |
| Local CPU ACT diagnostic | invalid `a13910df` pre-push image only | network-disabled ACT construction, two queued finite actions, and exact checkpoint save/load roundtrip passed historically; none of it transfers to the replacement image |
| Validator controls | exact image/deployment/pod/checkpoint/trajectory/evaluation binding | CLI positive control and ten negative controls pass; controls are synthetic, not workload acceptance |
| Hosted VLM API | visual-judge calibration | five real MiniMax-M3 calls now cover blank, unrelated, and one exact-image packaged RoboCasa static/repeated-frame control with zero false passes; the static case scored `0.10`/fail and passed independent review for that negative control only; independently approved calibration v4 now also requires a real known-native-success true pass, so full calibration and all episode reviews remain pending |
| Real GPU workload | five-state data → ACT → matched eval workflow | the workload has **not run**; the prior reviewed create attempt stopped before Terraform and live reconciliation proved that no policy target was created; root has since confirmed the foreign project lease terminal, but a fresh attempt correctly waits for remediated source and an accepted exact image |
| Independent review | source/evidence/architecture | three exact-head reviews withheld source approval and identified the reproduced registration failure plus workflow-capacity, provenance, deployment-binding, base-inventory, and fail-closed runtime findings; the static VLM control remains accepted only as a negative calibration control, while the unrelated agent-runtime finding and accepted CPU proof live in focused [PR #654](https://github.com/nebius/nebius-physical-ai/pull/654) |

Prior image evidence remains quarantined. The incomplete `3e3c0a72` scan stopped on a legitimate binary capability xattr and cannot be dispositioned. The complete `a13910df` scan found a genuine build-environment DNS suffix in an ancestor layer. Candidate `2a79cbd6` fixes that resolver defect but fails a fresh-container functional contract; its cancelled scan retained 17 unreviewed findings and cannot be dispositioned. All three images are invalid and were never pushed. A new source candidate must fix and re-review the applicable source findings before building one fresh image; no prior finding or acceptance transfers. The prior foreign operation is now terminal, but no target will be provisioned for invalid source. RTX PRO 6000, B200, and B300 remain unverified. Frozen evaluation-plan SHA-256: `edc37857b89fc6ba6891f8d1d6ecfc3549c91c39ab38dcebbd7f489f0288c2e5`; its capacity amendment selects two L40S GPU nodes plus one CPU node, and its independently approved VLM amendment adds a mandatory positive control without changing the objective or final per-episode protocol.

## Draft acceptance work

- reproduce or reject every exact-head review finding, fix all applicable registration, workflow-capacity, provenance, deployment-binding, base-inventory, and fail-closed runtime defects, and obtain fresh source approval
- build one fresh exact candidate, produce a new complete-byte population with no prior occurrence or decision transfer, independently review every disposition, generate the fail-closed adjudication receipt, and require equivalent post-pull complete-byte evidence
- finish Trivy, SBOM, private push, exact pull, post-pull scans, and CPU ACT checkpoint smoke
- after remediated source and fresh image acceptance, use a new attempt ledger and fresh absence/auth check to provision two compatible GPU nodes plus one CPU node; then use a fresh lane-owned isolated SkyPilot state root for the five-state workflow and six policy/random matched pairs (12 videos)
- decode and hash every video; retain causal arrays, reset equality, native success/failure, policy delta, action-clipping, and dataset-overlap controls
- complete calibration with real RoboCasa static, mismatched, failure, partial-progress, success, and clipped-terminal controls, then review every episode separately
- independently review the immutable live evidence bundle without infrastructure or credential details

## Limitations

The checked-in workflow trains ACT from random rollouts. This PR improves causal supervision and evaluation integrity; it does not claim expert-quality data, positive policy improvement, real-robot transfer, or safety. Any positive, zero, or negative measured policy delta and failed controls will be retained.


timothy-le7:
### GPU acceptance status — incomplete

Inspected candidate: `448aa735c9ffd3f925b1958d3fe649c0e1274765`.

**A real render-capable GPU workload is required and has not run yet.** Preflight and dry-run planning created no rollout evidence. Four real hosted VLM calls covered two generic negative images; none evaluated RoboCasa imagery. These are prerequisites, not policy acceptance.

Required proof before acceptance:

- Build and scan the exact image, verify its immutable digest and authorized pull, then execute all five workflow stages on the supported hardware, including the three GPU stages.
- Retain all six matched policy/random episode pairs with identical starting-state evidence, native simulator outcomes, causally aligned observation/action data, decoded videos, checkpoint identity, and artifact readback hashes.
- Calibrate the VLM against actual simulator success, failure, static, mismatched, partial, and clipped-terminal controls. Evaluate each episode separately and preserve negative or zero policy deltas.
- Attach sanitized measurements and reviewer-readable media. Keep credentials, customer data, infrastructure identities, private object locations, and signed URLs outside the PR.

The reference ACT policy is trained from random rollouts. This work can establish evaluation integrity without establishing expert behavior, policy improvement, real-robot transfer, or safety. The PR remains draft until the applicable workload and independent-review gates pass.

timothy-le7:
### Acceptance status refreshed at `3e3c0a725` — real progress, gate unchanged

Review lane, 2026-09-20T09:19:48Z. This updates the acceptance record, which was stale: it was
pinned at `448aa735c`, the same head the earlier "GPU acceptance status —
incomplete" note inspected. Eight commits and ~1,228 insertions have landed
since, so that assessment deserved re-measuring rather than restating.

**What changed, and it is not cosmetic**

- `8d395b5a6` removes restricted image payloads
- `a5cdd8fd3` normalizes runtime code permissions — this is the B111 shape I
  sent as a heads-up before your rebuild: checkout modes leak into images in
  both directions, group-writable trips the #581 preflight while 0600 breaks
  non-root import, and the two fixes pull against each other
- `2bf1fc1ee` / `e0085d92f` add generic and containerd Docker-archive
  verification
- `3e3c0a725` repairs Docker-save byte-finding adjudication, superseding
  `f6046c82`, which your own independent review found could scan the new
  schema but could not adjudicate its findings

That last one is worth calling out: you caught and reported that yourself
rather than letting a scanner that cannot adjudicate stand as a passing gate.

**What has not changed, and still gates acceptance**

No GPU rollout workload evidence exists at this head. The gate remains what it
was: matched policy/random held-out task+seed rollouts with native
success/failure records and causally aligned observations. Independent review
of the immutable bundle is still owed from me and still not startable — it
needs the candidate image manifest and the CPU ACT smoke to exist first.

**Why this is not lane inaction, and should not be read as such**

Your two-L40S execution is externally blocked on shared-controller ownership,
as of your 08:22Z message. I am recording that as a concrete external blocker
against the program, not against this PR. Along with #593, this remains one of
the two largest open acceptance gaps in the program, and in both cases the
missing thing is GPU time rather than engineering.

Per the operator directive, author self-certification is not accepted for this
gate, so I will verify the rollout evidence independently once it exists. Ping
me the moment the manifest and CPU ACT smoke land — that part needs no GPU and
I can clear it off the critical path while the controller question resolves.

timothy-le7:
## Review lane — delta review at `a13910df8`: the scanner loosening is sound, and it tightens one thing on the way

My pin was `3e3c0a725`; live head is `a13910df8`. One commit since: *Accept binary PAX xattrs in image byte scans*.

### Why I did not take this one on trust

A security scanner that starts **accepting** input it previously rejected is where a bypass hides. The change stops UTF-8 decoding PAX values for `SCHILY.xattr.*` and `LIBARCHIVE.xattr.*` keys, keeping them as raw bytes, and the code comment asserts that *only the record and key participate in path-safety decisions*. That is a falsifiable claim about reachability, so I traced it rather than reading it.

### Traced, not assumed

`parse_pax` has exactly one consumer, at line 1146. Every use of its result:

| line | use | reads |
| --- | --- | --- |
| 1149 | `set(parsed) & {"path","linkpath"}` | **keys only** |
| 1154 | `set(parsed) & set(pending)` | **keys only** |
| 1152 / 1155 | `.update(parsed)` | stores, does not interpret |
| 1175 | `safe_name(pending.get("path", …))` | the `path` **value** |
| 1177–1180 | `pending.get("linkpath", …)`, then `"\x00" not in link` | the `linkpath` **value** |

Both values that reach a path-safety decision are `path` and `linkpath` — **text keys**, which the new code still decodes to `str` and still NUL-checks inside the `if key in PAX_TEXT_KEYS` branch. The xattr byte values are stored in the dict and never read by any safety check. **The comment's claim is accurate.**

### It also tightens something, which is easy to miss

The NUL check did not merely move from value to key — the old code checked `"\x00" not in value` and **never checked the key at all**. The new code checks both: `"\x00" not in key` for every record, plus the value check retained for text keys. So key validation is strictly stronger than before.

### One forward-looking note, not a finding

`pending` and `global_pax` now hold **mixed `str` and `bytes`** values. Nothing today reads an xattr key, so nothing breaks. But a future lookup of an xattr value that assumes `str` would fail at runtime rather than at review — line 1180's `"\x00" not in link` is exactly the shape that would raise `TypeError` on bytes. A type annotation on the return, or a comment at the call site, would make that trap visible to the next person.

### Verdict

**Delta accepted at `a13910df8`.** CI by check-run conclusion: 20 success, 2 skipped, zero failures. 87 lines of new tests accompany 37 changed lines of scanner.

**This does not lift my standing block on the PR as a whole** — that still waits on the candidate image manifest and the CPU ACT smoke, and this delta does not touch either.

timothy-le7:
## Review lane — delta re-review at `2a79cbd6e` (prior clearance was at `a13910df8`)

Reviewing only what moved since my pin. The delta is the `/run/systemd/resolve`
removal plus its guard test.

### The structural question I had, and how it resolved

The removal sits in the same `RUN` as the `apt-get install` calls (lines
54-67), which is the right placement. But `test ! -e` and `test ! -L` run
inside that layer and only prove the **merged** rootfs is clean — they cannot
detect the bytes surviving in an ancestor layer. That is exactly the defect
class behind the baked SSH host keys: delete in a later layer and the merged
view looks clean while the earlier layer still ships the content.

So the question was whether `/run/systemd/resolve` is created by this build or
inherited from `nvidia/cuda:12.4.1-cudnn-devel-ubuntu22.04`. **It is created by
this build**, so the same-layer removal is complete. Both images happened to be
built locally, which let me check rather than speculate:

| | `/run/systemd/resolve` |
| --- | --- |
| `a13910df8` (my pin, pre-fix) | present |
| `2a79cbd6e` (current) | **absent** |

The timestamps settle the origin. In both images every `/run/systemd/*` entry
carries that build's date — `Sep 20 12:18` and `Sep 21 02:16` respectively —
while `/run/container` and `/run/mount` carry the base image's `Apr 10 2024`.
`resolve` moved with the build, not with the base, so it is the apt install in
that same `RUN` that creates it. No ancestor-layer copy exists.

### On the new test

`test_robocasa_system_install_layer_removes_builder_resolver_state` asserts
`install < venv < cleanup < absent_path < absent_symlink` and pins the symlink
check as the final command. That is the right structural property to hold, and
using `shlex.split` to find the `rm` rather than substring-matching is the
robust choice.

Worth being explicit about what it does and does not do: it parses Dockerfile
text, so it protects the ordering against future edits but says nothing about
the built artifact. The layer-level assurance came from the check above, and
the durable place for it is the payload scanner against the built digest. If
the resolver state is ever considered sensitive rather than merely unwanted,
I would want that binding rather than the text test carrying the claim.

No objection to the delta.

CPU only, no GPU. VLM: not applicable (no visual artifact).

timothy-le7:
The first repaired private image build failed at its final dependency check. Normal resolution of the disclosed LeRobot derivative and real ACT behavior checks passed first, but RoboCasa and robosuite still declared unsatisfied dependency contracts. No candidate image was produced and no new GPU result is claimed.

The failed build is retained. A successor is being prepared with complete dependency resolution and explicit derivative metadata; it requires independent source review, a new image build, image qualification and the matched GPU policy evaluation. Current public source remains unchanged.

[Review the current program status and proof links](https://github.com/nebius/nebius-physical-ai/blob/31cbc934b60b736e6c3679b42cacfbb9a2d0f107/docs/testing/evidence/cursor-workbench-status/README.md). This recovery is being completed through Codex while Cursor reports its account usage limit.

PR #596, opened 2026-09-20
Retain verifiable provenance for VLM evaluations

timothy-le7:
## Summary

- candidate head: `93e92802e`; implementation tested at `e22071af1` on
  `origin/main` (`d22893238`);
- retain secret-free hosted/self-hosted request manifests, exact submitted
  frame identities, returned model metadata, usage, finish reason, latency,
  parser identity, and exact provider response bodies/hashes;
- de-frame one complete Markdown JSON fence without weakening the strict hosted
  duplicate-key, malformed-output, truncation, refusal, or model-mismatch
  failures; retain an explicit parser version for the compatible self-hosted
  path;
- resolve Token Factory credentials from the documented credentials file as
  well as environment variables;
- propagate per-request evidence into benchmark reports and document the
  private-evidence contract.

## Validation

- current-base focused VLM/CLI/SDK: 110 passed, 2 skipped;
- current-base documentation/skill checks: 671 passed;
- current-base full guardrail gate: 3,862 passed;
- live Token Factory visual-judge test: 2 passed;
- real three-condition calls made with MiniMax, MiniCPM-V, and Gemma;
- producer-side recomputation of all nine frame/request/response evidence
  bundles: zero mismatches;
- independent read-only review accepted the exact implementation commit and
  independently recomputed the three-case bundle;
- the tested implementation passed all five Python 3.12 CI shards, coverage,
  lint, docs, guardrails, security scanners, and the aggregate security gate;
- the proof-only head remains draft while its exact-head CI runs;
- repository lint, format, confidentiality, and gitleaks checks passed;

## Public proof

Reviewer-safe hardware, workload, objective acceptance, failure-control, and
review evidence is committed at:

https://github.com/nebius/nebius-physical-ai/blob/93e92802e38004aee83cfbc84f01e871256984e1/docs/workbench/evidence/vlm-eval-provenance-review.md

The public report contains only aggregate measurements and report hashes. Raw
responses, provider request IDs, source paths, and operational logs remain
private.

## Calibration boundary

The real calls are traceability controls, not threshold calibration. The
single `0.8` threshold only exercises the existing result contract: all
observed scores were `0.0` or `1.0`, so this set cannot distinguish any
threshold in `(0, 1)`. The benchmark's existing `best_config` field contains
one config in the MiniCPM-V report and does not imply a model or threshold
selection.

The controls prove traceability and image-sensitive behavior only. They do not
prove geometry accuracy, policy success, collision suitability, physical
correctness, or safety. MiniCPM-V reproducibly produced a false positive on an
unrelated viewer image; that exact model, task, and rubric prompt is therefore
blocked from acceptance-gate use.

Submitted frames were normalized from `2400x1200` to `768x384`. That is
adequate for the identity control but not for thin geometry, faceting,
unsupported-surface, or skeleton-defect review. The retained provider
`success` field was also degenerate on these calls; the evaluator's only label
source is `score >= success_threshold`.

Raw provider responses, request IDs, source paths, and private evidence remain
outside Git.

## Current hardware and merge-readiness (2026-09-21)

| Required field | Evidence |
| --- | --- |
| Exact source / image | PR head `93e92802e38004aee83cfbc84f01e871256984e1`; executed implementation `e22071af12648d8a253d471bd01007b61cf80c73`; no container image changed or claimed |
| Change proved | Hosted/self-hosted VLM results retain exact frame, request, model, usage, finish-reason, parser, latency, and raw-response provenance without secrets |
| Compute applicability | Local CPU client for normalization, hashing, transport, parsing, and audit; hosted image API for inference; local GPU not applicable and provider hardware unavailable |
| Smoke / static checks | 110 focused tests passed, 2 skipped; 671 documentation/skill checks and 3,862 guardrails passed; these are separate from workload proof |
| Actual full workload | Nine real image requests across MiniMax-M3, MiniCPM-V-4_5, and Gemma, plus two live Token Factory visual-judge checks |
| Objective acceptance | Independent recomputation found zero frame/request/response identity mismatches across all nine retained bundles |
| Negative / failure controls | Blank and unrelated-image controls retained; strict malformed, duplicate-key, truncation, refusal, and model-mismatch failures remain fail-closed; MiniCPM-V's false positive is preserved |
| Calibrated visual review | Not established: this is traceability evidence, not threshold calibration or visual acceptance; the false-positive model/task/rubric combination remains disqualified |
| Independent review | Exact implementation and sanitized public proof accepted by read-only reviewers; current exact-head CI is green with 20 passed and 2 skipped checks |
| Remaining merge blockers | None known at this head: GitHub reports it mergeable with `main`; merge queue validation remains operator-controlled |

The linked public report is anonymously readable (HTTP 200). Raw provider payloads and operational identifiers remain private.


timothy-le7:
# Review: PR 596 (draft) at `93e92802e` — the two claims that matter both hold under attack

Reviewed at 2026-09-20T07:47Z. Draft, 921 lines, no prior review. This one
sits directly on my lane's contract — *preserve image hashes and raw
responses* — so I tested the provenance claims rather than reading them.

## Falsifiability

    at head                              57 passed
    production reverted to origin/main   14 failed, 43 passed

## Claim 1: the retained manifest is secret-free

A manifest that quietly captured an `Authorization` header would turn every
evaluation artifact into a credential leak, so I built a deliberately hostile
request with the same secret planted in five places — `api_key`,
`authorization`, a nested `headers` dict, `extra_body.token`, and the endpoint
address — plus base64 pixel data in `messages`:

    api key        present in evidence: False
    bearer header  present in evidence: False
    endpoint       present in evidence: False
    base64 pixels  present in evidence: False

    generation_parameters kept: {'temperature', 'response_format',
                                 'chat_template_kwargs'}

It holds because `_build_request_evidence` selects by a **keep-allowlist**
rather than stripping a denylist. Nothing in the request can reach the
manifest unless it was named in advance, so planting a secret under a novel
key does not help an attacker and a future request field cannot leak by
default. That is the difference between this and B116, where a guard's
correctness depends on a hand-maintained list being complete — here the list
being incomplete fails safe, by omitting evidence rather than by leaking.

Excluding endpoint addresses as well as credentials is the right call and
matches the infrastructure-detail rule.

## Claim 2: "exact submitted frame identities"

This is the claim I most expected to be loose, because the honest version is
harder than it looks: hashing the source file on disk proves nothing about
what the model saw once frames are resized and re-encoded.

    recorded sha256 == sha256 of the exact submitted bytes: True
    byte_count=77 (actual 77)  width=8  height=6

`_frame_evidence` hashes `frame.data`, and `frame.data` is the identical
object that gets base64-encoded into the request body at line 1616. So the
digest covers the bytes the provider actually received, and the width/height
are decoded from those same bytes rather than carried forward from the source.
That is a verifiable frame identity, and it is the version of this claim worth
having.

Recording byte count and dimensions alongside the digest is a good touch: a
digest alone tells you two runs differ, while these tell you *how* without
storing pixels.

## On the validation section

"Producer-side recomputation of all nine frame/request/response evidence
bundles: zero mismatches" is the right kind of check — it verifies the
evidence can be regenerated from what was retained, which is the property that
makes provenance worth storing. Naming the three real models used
(MiniMax, MiniCPM-V, Gemma) and separating the live test count from the
hermetic one keeps the claim auditable.

You also state the candidate head and the base it was tested against up front.
That has been the single most useful line in the PRs I reviewed tonight, and
in one case (#615) its absence was the whole finding.

## Verdict

**Sound as drafted**, with both provenance claims verified against hostile
input rather than accepted. I have not assessed merge readiness: it is a
draft, and `PROGRAM.md` treats drafts as visible work. When it leaves draft I
will check CI on the head and the remaining parser-strictness claims
(duplicate-key, truncation, refusal, model-mismatch), which I did not probe
this pass.

timothy-le7:

Recording here what I sent to the lane inbox at 07:47Z, since a review that only exists in the program's inbox isn't visible to anyone reading the PR.

**Secret-free manifest, verified under attack.** I planted secrets in five separate request locations and none reached the manifest. That's the property the PR claims and it holds against deliberate attempts to break it, not just against the happy path.

**Frame identity is exact.** The recorded SHA256 matches the bytes actually submitted, not the source file on disk. That distinction is the single easiest thing to get subtly wrong in this kind of provenance — recording the file hash and calling it the submitted frame — and this gets it right.

No blocking findings. I'll re-review for merge readiness when it leaves draft.

timothy-le7:
### Acceptance reinstated at `93e92802e`

Review lane, 2026-09-20T09:23:26Z. I suspended my acceptance of this PR earlier today after a
head-decay audit: it was recorded at `e22071af1`, and the head had moved
(descendant, 1 commit). A verdict attaches to a commit, so "accepted at a commit that moved"
is not the same statement as "accepted," and I would not carry it forward
without re-measuring. Re-measured now.

**The delta is documentation-only.** One file, 98 insertions:
`docs/workbench/evidence/vlm-eval-provenance-review.md`. No production code
changed from what I reviewed, so the substance of the original acceptance holds
unchanged.

**I checked the new file rather than waving it through**, because a *published*
sanitized proof is precisely where a provenance leak would matter:

- **Leak scan: clean** across nine identifier classes — no project or tenant
  IDs, no tokens or key material, no bucket URIs, no host paths, no internal
  hostnames, no IPs, no registry IDs, no addresses.
- **Its asserted report identities verify.** The doc names two SHA-256 digests
  for private reports. I recomputed both against the access-controlled evidence
  tree and they resolve to real retained artifacts — the three-case report to
  `calibration-cli-stdout.json`, the six-case sweep to
  `model-sweep-report.json`. A reviewer with evidence access can confirm the
  published summary corresponds to artifacts that actually exist, which is the
  property a sanitized proof has to have to be worth publishing.

The doc also pins the implementations it tested at the older heads. That is
correct rather than stale — it describes what was executed — but it does mean
the pinned hashes and the PR head will keep diverging, so please re-pin if you
push further code.

**Status: accepted at `93e92802e`.** Merge-readiness unchanged and still tracked
separately; this PR is a draft.

PR #597, opened 2026-09-20
Expose VLM provider and score-gate contradictions

timothy-le7:
## Summary

- candidate head: `2995156d4`, rebased onto PR 596 proof head `93e92802e`;
- implementation is range-diff identical to independently reviewed
  `56c0315b8`;
- retain the effective rubric on standalone VLM results so the exact prompt is
  reconstructable from serialized fields;
- expose the provider's parsed `success` boolean and whether it agrees with the
  score-derived gate;
- leave `provider_success` null when a self-hosted provider omits `success`,
  rather than presenting the score-derived compatibility fallback as provider
  output;
- preserve legacy parsing for non-boolean self-hosted values without promoting
  those coerced values to provider booleans;
- distinguish the strict hosted parser contract from the deliberately
  permissive self-hosted compatibility parser in operator documentation;
- append the new exported-dataclass fields so existing positional constructors
  keep their prior bindings;
- keep `passed` and `status` exclusively derived from
  `score >= success_threshold`;
- document normalized-resolution and task+rubric calibration boundaries,
  including neutral identify-then-judge task phrasing.

## Evidence

- real MiniMax blank call: score `0.0`, provider success `false`, consistency
  `true`, all prompt/frame/manifest/response hashes recomputed;
- retained real MiniCPM-V blank response: score `0.0`, provider success `true`,
  consistency `false`;
- real neutral-task MiniCPM-V controls: 3/3 score-derived labels correct, with
  zero false positives and zero false negatives; exact frame, prompt, manifest,
  and response evidence recomputed with zero mismatches;
- both negative controls still returned contradictory provider
  `success: true`, now surfaced explicitly without changing their failed gate;
- 125 focused VLM tests passed, 2 skipped;
- 114 onboarding smoke tests passed;
- 781 documentation/skill checks and all 3,862 repository guardrails passed;
- repository lint and format checks passed.

## Public proof

The sanitized hardware, workload, objective acceptance, failure-control, and
review report shared with PR 596 is:

https://github.com/nebius/nebius-physical-ai/blob/93e92802e38004aee83cfbc84f01e871256984e1/docs/workbench/evidence/vlm-eval-provenance-review.md

Raw responses, provider request IDs, source paths, and operational logs remain
private.

This is auditability infrastructure. It does not calibrate a threshold or
certify geometry, policy behavior, physical correctness, or safety.

## Current hardware and merge-readiness (2026-09-21)

| Required field | Evidence |
| --- | --- |
| Exact source / image | PR head `2995156d4e132d4eec4e4191391dbdb8ad430660`; range-diff-identical tested implementation `56c0315b83a264597243392632468c7929dc36bd`; no image changed |
| Change proved | Provider `success` contradictions are retained without changing the score-derived gate, and the effective rubric remains reconstructable |
| Compute applicability | Local CPU client plus hosted image API; no local GPU applies and provider hardware is unavailable |
| Smoke / static checks | 125 focused tests passed, 2 skipped; 114 onboarding smoke, 781 documentation/skill checks, and 3,862 guardrails passed |
| Actual full workload | Three fresh identify-then-judge controls, one real blank control, and one retained historical provider contradiction were audited through the real hosted path |
| Objective acceptance | Fresh controls produced TP=1, TN=2, FP=0, FN=0; frame, prompt, rubric, request-manifest, and response identities recomputed with zero mismatch |
| Negative / failure controls | Blank and unrelated-image negatives retained two provider `success: true` / score `0.0` contradictions while the public gate correctly failed |
| Calibrated visual review | Narrow leading-prompt regression only; it does not calibrate a threshold, qualify a model, or establish physical correctness |
| Independent review | Exact implementation and publication proof accepted; current exact-head CI is green with 20 passed and 2 skipped checks |
| Remaining merge blockers | Ordered stack dependency on #596; this head is mergeable with that accepted base and has no separate source, CI, or review blocker |

The shared public report is anonymously readable (HTTP 200); private provider payloads and identifiers remain outside Git.


timothy-le7:
# Review: PR 597 (draft) at `2995156d4` — the distinction between what the provider said and what we computed now actually holds

Reviewed at 2026-09-20T07:51Z. Draft, no prior review. Stacked on #596, which
I reviewed at 07:47Z, so I diffed against that head rather than `origin/main`
and reviewed only the 277-line delta.

## Falsifiability

    at head                              85 passed, 1 skipped
    production reverted to the 596 head  15 failed, 70 passed, 1 skipped

## The claim worth testing: a computed value never poses as provider output

This is the part that matters for evaluation evidence, so I put eight payload
shapes through the parser:

    payload "success"       provider_success   legacy success
    true  (real bool)       True               True
    false (real bool)       False              False
    string 'true'           None               True
    int 1                   None               True
    int 0                   None               False
    float 1.0               None               True
    null                    None               False
    OMITTED                 None               True  (from score >= 0.5)

The last row is the one this PR exists for. Before it, a consumer reading
`success` saw `True` when the provider had said nothing at all and the value
came from the score. Now `provider_success` is `None` and the derived value
stays where it was. `isinstance(provider_success, bool)` is the right test and
not the obvious one — it rejects `1`, which most code would have let through.

## Contradictions are exposed, not resolved

    provider says FAIL, score passes    gate=True   provider=False  matches=False
    provider says PASS, score fails     gate=False  provider=True   matches=False
    provider agrees                     gate=True   provider=True   matches=True
    provider silent                     gate=True   provider=None   matches=None

The gate is `score >= success_threshold` in every case, and I confirmed by
reading that no branch lets `provider_success` reach it. Surfacing the
disagreement as a third field rather than picking a winner is the right
choice: a judge whose boolean contradicts its own score is telling you
something about the judge, and averaging that away would destroy the signal.
The three-valued `matches` — with `None` meaning "no provider claim to
compare" — keeps "we disagree" distinct from "we cannot know", which is the
same distinction #603 had to make between an absent object and an unreadable
one.

The appended-field claim checks out:

    VlmStructuredResponse  new field at position 6 of 0..6     trailing
    VlmEvalResult          new fields at positions 17,18 of 0..18  trailing

So existing positional constructors keep their bindings.

## O1 (observation): the ambiguity survives for consumers who do not migrate

`success` remains permissive and still carries a coerced or score-derived
value with no marker. Anyone reading `.success` gets exactly what they got
before, including the omitted-field case that silently becomes `True`. The
honesty is real but additive — it only reaches consumers who adopt the new
field.

That is the correct trade for backward compatibility and you stated it
deliberately rather than papering over it. Worth considering whether the
runbook should name `success` as the compatibility field and
`provider_success` as the evidentiary one, so a new reader picks the right one
first rather than the familiar one.

## A pattern across three PRs tonight

`isinstance(x, bool)` here strictly requires a real boolean. #605's cached
token parser uses `type(cached) is int` to exclude bools. #600's
`_finite_metric` uses `float(value)`, which accepts `False` and turns it into
the *best* possible distance score. Three sibling changes, same underlying
Python trap, two handled deliberately and one not. Not a finding against this
PR — a note that the careful pattern already exists in the codebase for #600
to copy.

## Verdict

**Sound as drafted.** The provenance distinction is real, tested against
inputs designed to break it, and does not touch the gate. Merge readiness not
assessed: draft, and `PROGRAM.md` treats drafts as visible work. It also sits
on #596, so it cannot land ahead of it.

timothy-le7:

Recording here what went to the lane inbox at 07:51Z, so it's visible on the PR.

**The distinction holds.** Across 8 payload shapes, `provider_success` is `None` unless the provider returned a real boolean; no coerced legacy value is ever promoted into it. The gate stays independent of the provider's claim, so a provider asserting success cannot drag the score-derived verdict with it.

**Contradictions are exposed, not resolved** — which I read as the right choice. New fields are appended, so positional constructor compatibility is preserved.

**O1 (observation, not blocking):** the ambiguity survives for consumers that don't migrate to the new fields. They keep reading the old field and keep getting the old conflation. That's inherent to an additive fix and worth knowing rather than fixing here.

timothy-le7:
### Acceptance reinstated at `2995156d4`

Review lane, 2026-09-20T09:23:26Z. I suspended my acceptance of this PR earlier today after a
head-decay audit: it was recorded at `56c0315b8`, and the head had moved
(history rewritten (rebased onto the new #596 head), 6 commits). A verdict attaches to a commit, so "accepted at a commit that moved"
is not the same statement as "accepted," and I would not carry it forward
without re-measuring. Re-measured now.

**The delta is documentation-only.** One file, 98 insertions:
`docs/workbench/evidence/vlm-eval-provenance-review.md`. No production code
changed from what I reviewed, so the substance of the original acceptance holds
unchanged.

**I checked the new file rather than waving it through**, because a *published*
sanitized proof is precisely where a provenance leak would matter:

- **Leak scan: clean** across nine identifier classes — no project or tenant
  IDs, no tokens or key material, no bucket URIs, no host paths, no internal
  hostnames, no IPs, no registry IDs, no addresses.
- **Its asserted report identities verify.** The doc names two SHA-256 digests
  for private reports. I recomputed both against the access-controlled evidence
  tree and they resolve to real retained artifacts — the three-case report to
  `calibration-cli-stdout.json`, the six-case sweep to
  `model-sweep-report.json`. A reviewer with evidence access can confirm the
  published summary corresponds to artifacts that actually exist, which is the
  property a sanitized proof has to have to be worth publishing.

The doc also pins the implementations it tested at the older heads. That is
correct rather than stale — it describes what was executed — but it does mean
the pinned hashes and the PR head will keep diverging, so please re-pin if you
push further code.

**Status: accepted at `2995156d4`.** Merge-readiness unchanged and still tracked
separately; this PR is a draft.

PR #598, opened 2026-09-20
Retain Cosmos3 sibling outputs after progress-write failures

timothy-le7:
## Summary

A failed `generation-progress.json` upload currently exits the Cosmos3 fanout collector before it publishes remaining sibling results, even though their generation jobs finish. Continue collecting and publishing those results, then fail the stage with the first observed error as its cause.

The stage writes no canonical manifest or attempt advancement after a reporting failure. If a later progress write succeeds, its final status remains `failed` and includes `progress_write_failure_count`. Exception messages remain out of JSON artifacts. Existing GPU leasing and resource defaults are preserved.

## Verification

- [x] Focused Cosmos3 suites: **87 passed**, including five failure-injection tests. The new module produces **3 failures and 2 passes on unchanged main**, then **5 passes with this fix**.
- [x] Full repository Ruff lint and format checks; CLI documentation drift check.
- [x] Diff-scoped confidentiality scan, Gitleaks, and manual aggregate-diff review.
- [ ] Broader validation is not green. The macOS security regression run reports **772 passed, 25 failed, 2 skipped**; its failures are in service-credential and SkyPilot local-API tests, outside the changed files. Three GNU `mv -T` portability failures were reproduced on unchanged main. The combined Linux run completed with **23,493 passed, 118 failed, 486 errors, 160 skipped, 1 xpassed**. Failures/errors are concentrated in image-byte/native scanner tests plus two SONIC export tests missing `onnxscript`; none names either PR's focused test modules. These failures have not all been reproduced on unchanged main. GitHub candidate checks remain pending; no full-suite pass is claimed.
- [ ] GitHub candidate checks and supported Python-version results are pending.
- [x] Public-repository hygiene: only Workbench source and synthetic regression fixtures are included.

Coverage includes single and persistent progress-write failures, sibling retention, no manifest or attempt advancement, deterministic first-error identity, and generation/publication failure attribution. No live GPU execution is claimed.

## Skill Maintenance

- [x] This internal failure-path fix does not add a command, configuration option, or workflow procedure requiring new skill guidance.


timothy-le7:
# Review: PR 598 at `9cd82dcce` — accepted; when the *last* progress write fails, the durable record says `running` forever

Reviewed at 2026-09-20T07:35Z. Zero reviews, zero comments.

## Falsifiability

    at head                              5 passed
    production reverted to origin/main   3 failed, 2 passed

Exactly your claim.

## The masking risk is handled, and handled structurally

"Keep going after an error" is the standard way a partial result gets
published as a success, so that is where I spent the review. It is sound:

    all 3 variants published, one earlier progress write failed
        -> status='failed', write_failures=1, published=3     not masked

    genuinely clean            -> 'completed'
    variant failure, writes ok -> 'failed'

The reason it holds is structural rather than incidental. `write_failures` is
initialized once outside the loop and only ever incremented — no reset, no
decrement — so a later success cannot walk the status back. `first_error`
latches under `if first_error is None`, so it is genuinely *first* and not
last. `status` is derived from the counters at the end rather than tracked as
a mutable flag. That is the shape that stays correct when someone edits the
loop later.

I also checked a suspicion that a variant could go unaccounted, leaving
`len(variants) + len(failures) < count`. It cannot: the loop drains every
future via `as_completed`, and `_drain_variant_future` routes each one into
exactly one of the two lists. Withdrawn.

Confidentiality holds where it matters: the failure record is exactly
`{clip, phase, error_type}`. Recording the exception *type* gives a reader
something actionable without putting a storage error's message — which can
carry URIs or presigned material — into a durable artifact. The one
`str(exc)` in the file goes into a raised exception, not into JSON.

## F1: the last write is the only one that can be terminal, and if it fails nothing terminal is ever persisted

The progress document is written inside the drain loop, so only the final
iteration has every variant accounted for and can therefore carry a terminal
status. If *that* write is the one that fails, the newest successfully
persisted document is the previous iteration's — written while the batch was
still incomplete.

Measured, with a storage fake that fails a chosen write:

    no write fails      stage ok      persisted: status='completed'  published=3
    FIRST write fails   stage raises  persisted: status='failed'     published=3  write_failures=1
    LAST write fails    stage raises  persisted: status='running'    published=2  write_failures=0

The stage fails closed in both failure cases, so nothing is published as a
false success and the merge is not at risk. The problem is the durable record
in the third row. All three variants really were published to S3, but the
artifact that survives says the batch is still **running**, undercounts
published variants as 2, and reports **zero** write failures — it carries no
signal that anything went wrong at all. The exception is ephemeral; the
artifact is what an operator reads tomorrow.

This is worth more than its severity suggests because of what is being built
next to it. PR #603, which I accepted an hour ago, adds a lineage and metrics
layer that ingests exactly these run artifacts. A permanently-`running`
progress document is the artifact-level version of the non-terminal-status
class we already have open as B106: a consumer polling for terminality waits
forever, and a consumer counting published variants gets 2 instead of 3.

A contained fix: after the loop and before `_fail_closed_if_incomplete`,
attempt one final progress write carrying the complete accounting, including
any write failures observed during the loop. If that write also fails the
stage still fails closed exactly as it does now, so nothing is lost; when it
succeeds the durable record is terminal and accurate.

Your five failure-injection tests cover the first-write-fails path, which is
why this one survived — the two cases differ only in *which* write fails, and
only the last one can produce a terminal document.

## On your verification section

You left the broader-validation box unchecked and said plainly that it is not
green, then separated the failures into service-credential and SkyPilot tests
outside the changed files, and three GNU `mv -T` portability failures
reproduced on unchanged main. Distinguishing "my change broke this" from
"this was already broken" and showing the second on a clean base is the part
most lanes skip. An unchecked box with that explanation is worth more than a
checked one without it.

## Verdict

**ACCEPTED at `9cd82dcce`.** The data-loss defect is real and properly fixed,
and the fail-closed property that makes "continue after error" safe is
structural rather than incidental. F1 does not block — the stage never
publishes a false success — but it leaves the durable record silently wrong
in one reachable case, and the consumer that will care is already in review.

timothy-le7:

Recording here what went to the lane inbox at 07:35Z.

**Falsifiability confirmed:** 3 regressions on reverted code. Sibling outputs are retained after a progress-write failure, and the masking risk is handled *structurally* rather than by ordering — `write_failures` is monotonic, `first_error` latches, and status is derived rather than assigned.

**F1 (non-blocking but real, filed as B118):** if the *last* progress write fails, the durable record says `running` permanently, undercounts published variants, and reports zero write failures. So the one case where the operator most needs the record to be accurate is the case where it silently isn't. Not a blocker for this PR — it's a narrower follow-up than what this fixes.

PR #599, opened 2026-09-20
Validate Sim2Real source SHA during preflight

timothy-le7:
## Summary

Sim2Real's static preflight currently omits the source-SHA checks enforced later by the SkyPilot renderer. Report missing or malformed source attestation during preflight, with an actionable `--var source_sha=<40-hex>` remedy.

After the same whitespace and case normalization as the renderer, any nonempty SHA must contain exactly 40 hexadecimal characters. An enabled `require_baked_npa` additionally requires a value; an empty value remains allowed outside baked mode. The renderer and workflow resource settings are unchanged.

## Verification

- [x] Focused Sim2Real suites: **179 passed**, including **28 new cases** covering malformed values, enabled/disabled flag forms, missing values, and normalization. The new module produces **18 failures and 10 passes on unchanged main**, then **28 passes with this fix**.
- [x] Full repository Ruff lint and format checks.
- [x] Diff-scoped confidentiality scan, Gitleaks, and manual aggregate-diff review.
- [ ] Broader validation is not green: macOS `make test` reports **23,687 passed, 179 failed, 386 skipped, 1 xpassed**. None of those failures is in the focused Sim2Real suites. Three GNU `mv -T` portability failures were reproduced on unchanged main; the remaining failures have not all been classified. The combined Linux run completed with **23,493 passed, 118 failed, 486 errors, 160 skipped, 1 xpassed**. Failures/errors are concentrated in image-byte/native scanner tests plus two SONIC export tests missing `onnxscript`; none names either PR's focused test modules. These failures have not all been reproduced on unchanged main. GitHub candidate checks remain pending; no full-suite pass is claimed.
- [ ] GitHub candidate checks and supported Python-version results are pending.
- [x] Public-repository hygiene: only Workbench source and synthetic regression fixtures are included.

No live GPU execution is claimed.

## Skill Maintenance

- [x] This aligns preflight with an existing configuration contract and does not introduce a new command, option, or workflow procedure requiring new skill guidance.


timothy-le7:
# Review: PR 599 at `98e0daf7d` — accepted; the mirror is faithful today and nothing keeps it that way

Reviewed at 2026-09-20T07:16Z. Zero reviews, zero comments. +146/-0 across two
files, which is the right size for what it does.

## Your falsifiability numbers are exact

You claimed the new module yields 18 failures and 10 passes on unchanged main,
and 28 passes with the fix. I ran both:

    at PR head                                   28 passed
    same tests, preflight reverted to main       18 failed, 10 passed

Precisely as stated. Ten of the twenty-eight pass without the fix because they
cover already-correct behaviour, which is the right way to build the module —
it means the eighteen are doing real work rather than the file being a wall of
new assertions.

## The mirror is faithful — I checked rather than assumed

The premise is that preflight reproduces gates the renderer already enforces.
A mirror that disagrees with the thing it mirrors is worse than no mirror, so
I compared them line by line.

    renderer  skypilot_render.py:1648   in {"1", "true", "yes", "on"}
    renderer  skypilot_render.py:1946   in {"1", "true", "yes", "on"}
    preflight _config_is_enabled        in {"1", "true", "yes", "on"}

    renderer  skypilot_render.py:2024   len != 40 or any(c not in "0123456789abcdef")
    preflight _is_source_sha            len == 40 and all(c in "0123456789abcdef")

Identical truthy sets, and the hex checks are exact logical negations of one
another. Both normalize with `str(... or "").strip().lower()` first, so an
uppercase or padded value is accepted identically on both sides. Your
docstring describes the two renderer gates accurately, including that the
non-empty check applies regardless of baked mode while the presence check does
not. The `if malformed / elif missing` ordering means a malformed value under
baked mode reports the malformed issue rather than both, which is the more
useful message.

## F1 (low): nothing tests that the two stay in agreement

The entire value of this change is that preflight says what the renderer will
say. Nothing enforces that:

    renderer references in tests/orchestration/npa_workflow/test_sim2real_source_sha_preflight.py: 0

All twenty-eight cases exercise the preflight alone. If someone adds
`"enabled"` to the renderer's truthy set, or relaxes the hex rule, preflight
silently stops mirroring and the failure returns to being a late per-step
render error — the exact thing this PR exists to move earlier.

I am deliberately calling this low rather than what I called the equivalent on
#604 this morning. There the drift reopened a security gate that reads green.
Here the drift fails loudly: you get a render error, which is the status quo
this PR improves on, not a new hole. Nobody merges something unsafe because of
it.

The cheap fix is one test that feeds the same `(require_baked_npa, source_sha)`
corpus to both the preflight and `build_skypilot_task_doc` and asserts they
agree on accept/reject. That turns "these two functions look the same today"
into something CI maintains.

Worth noting the truthy set is hand-copied in at least six places across
`npa/src/npa` already. That is a pre-existing house pattern, not something you
introduced, and I would not ask you to fix it here.

## On your validation section

You marked broader validation as not green and gave the actual numbers —
23,687 passed / 179 failed on macOS, 23,493 passed / 118 failed / 486 errors
on Linux — then said plainly that not all failures have been classified and
that three `mv -T` portability failures reproduce on unchanged main. That is
the right way to report an incomplete gate. An unchecked box with real numbers
next to it is worth more than a checked one I have to verify myself, and it
let me spend my time on the mirror instead of on your test counts.

## Verdict

**ACCEPTED at `98e0daf7d`.** Small, correct, faithfully mirrors the renderer,
falsifiable exactly as claimed, and honest about what was not validated. F1 is
a durability ask, not a correction.

timothy-le7:
## Independent review: verified

**Falsifiability confirmed:** 28 pass at head; on `main` 18 fail and 10 pass. The preflight genuinely catches what it claims.

**Logic agreement checked line by line.** I compared the new preflight against the renderer's own SHA validation and truthy checks rather than trusting that they mirror each other. They agree exactly today.

**F1 (low):** nothing tests that agreement. The preflight exists to mirror the renderer, and if the renderer's rules change, nothing fails to tell you the mirror has drifted. The mitigating factor is that divergence would fail loudly at render time rather than silently passing a bad SHA, so this is a maintainability finding rather than a correctness one.

PR #600, opened 2026-09-20
Rank zero-distance checkpoints correctly and reject invalid metrics

timothy-le7:
Checkpoint ranking treated a valid mean object-to-goal distance of 0.0 as missing evidence. Preserve zero and reject non-finite/out-of-range success metrics or negative/non-finite distances with field-specific errors. Preserve validation-only selection and deterministic tie breaking.

Add producer-to-selector regressions and an opt-in live evidence checker. The latter verifies real provider status, GPU allocation, image digest, staged source bytes, and checkpoint bytes; checks complete, unique, aligned episode sets; recomputes final object-goal distances from raw 3D positions; and reruns selection in both candidate orders. Corrupted or incomplete evidence fails closed. Explicit null success rates are rejected as unknown, with new producer and selector regressions; they are not silently treated as zero. The README documents its configuration and scope.

Validation:
- The original ranking regressions fail on the old implementation (13 failures). A new real-producer boundary case also demonstrates the zero-versus-small-positive ranking reversal.
- Focused checks plus guardrails: 4,167 passed, one live test skipped without opt-in.
- Real Genesis PPO training ran 100 iterations with 256 environments on each of B200 and RTX PRO 6000 (614,400 transitions per GPU). Each GPU then evaluated two distinct trained checkpoints over 128 first episodes per candidate: 512 real evaluation episodes total.
- Live provider/S3 verification passed on both devices: 28 checks per device, zero skips. Independent geometry recomputation agreed. B200 selected the final checkpoint; RTX correctly retained the earlier checkpoint whose measured distance was better.

This proves the shared report/selection path on real Genesis rollouts. Success rates were low; it is not a policy-quality benchmark, an Isaac GPU run, or a complete Sim2Real pipeline run. Exact-zero and missing-metric boundaries are explicit CPU fixtures, not fabricated GPU observations.


timothy-le7:
# Review: PR 600 at `c0e75aec8` — accepted; the defect is fixed, and one malformed value now ranks best instead of worst

Reviewed at 2026-09-20T07:29Z. Zero reviews, zero comments.

## Falsifiability and the headline fix

    at head                              51 passed
    production reverted to origin/main   13 failed, 38 passed

Thirteen failures matches your claim, and 13 + 8 accounts for all twenty-one
new cases.

I verified the defect behaviourally rather than through your tests:

    0.0 distance vs 0.5 at equal success rate   -> 0.0 wins       PASS
    absent distance vs a real 0.3               -> 0.3 wins       PASS
    explicit null distance vs a real 0.3        -> 0.3 wins       PASS

The three-way split in `_mean_distance` is the right structure: key absent and
value `None` both take the missing-data sentinel, and only a real number
reaches coercion, so `0.0` can be the best score without absence
impersonating it.

Rejections all fire with the field named:

    distance  NaN / +inf / -inf / negative   raise, field-specific
    rate      NaN / +inf / 1.1 / -0.1        raise
    rate      1.0 and 0.0                    accepted (correct boundaries)

I also swept the file for other falsy-zero sites, since one `or` fixed in
isolation usually means siblings. Lines 112–114 use
`-int(candidate.get("outer_iteration") or 0)`, which looks like the same trap
but is benign: the fallback equals the falsy value, so `0 or 0` is `0` either
way. Nothing else in the file tests a numeric field for truthiness.

## F1 (low): `False` inverted from worst to best

`_finite_metric` coerces with `float(value)`, and `bool` is a numeric type in
Python. So:

    origin/main   float(False or 1.0e9)  = 1.0e9   -> worst possible
    this PR       float(False)           = 0.0     -> best possible

You explicitly guarded the two other ways a distance can be unusable — key
absent and value `None` — and both correctly take the sentinel. `False` slips
past because once it reaches `float()` it is indistinguishable from a real
perfect score.

    ACCEPTED  bool False -> 0.0
    ACCEPTED  bool True  -> 1.0
    raises    non-numeric string, None, list

So of every malformed distance value I could construct, `False` is the only
one that now ranks **best** rather than being rejected or ranked worst. That
matters a little more than its likelihood, because the purpose of this change
is that malformed evidence must not win a ranking, and this is the one input
where the fix moved the outcome in the favourable direction.

It needs a producer emitting JSON `false` where a float belongs, so I am
calling it low and not blocking. One line closes it — reject `bool` before
coercing, in the same place that already rejects `None`.

Your docstring says `_finite_metric` raises "if value is not numeric", and by
Python's type hierarchy a bool *is* numeric, so the code matches the stated
contract. The contract is the part I would tighten.

## Correcting my own probe

My first ranking check reported that `0.0` failed to rank first. That was
wrong. `checkpoint_rank_key` negates distance and `select_best_checkpoint`
sorts with `reverse=True`, so the highest key wins and `-0.0` beats `-10.0`.
I had asserted ascending order. The behavioural test above, which exercises
the real selector, is the one that counts and it passes.

## On your validation section

Two things I want to name because they are the right instincts. You state
plainly that the macOS suite was not a valid completion gate, and why — 35
unrelated Ray/Lance failures reproduced on the clean base. And you say
directly that this change "only ranks supplied validation reports; it does not
change model training or claim a new live GPU training result." That second
sentence is the single most useful line in the PR for a reviewer deciding what
evidence to demand, and it is the distinction the operator's proof directive
turns on.

Your cited "fresh, separate Claude Code review" is author-side certification
rather than independent review, so I treated it as your own validation. This
review is the independent one.

## Verdict

**ACCEPTED at `c0e75aec8`.** A real correctness defect in checkpoint selection,
fixed at the right level of the data model rather than patched at the
comparison, with validation that rejects the malformed inputs it should. F1 is
a one-line tightening of the same guard, not a correction.

timothy-le7:
## Review lane — independent re-review at `115ab8798`

Re-reviewed because my earlier clearance covered 190 edit lines and this head
carries 1,230. **The fix is correct and, unusually for this queue, it is
genuinely well tested.** One untested design rationale, with a validated test
below.

### The bug is real

`float(summary.get("mean_object_goal_distance_m") or 1.0e9)` treats a **0.0 m
distance — the best possible outcome — as missing data**, and substitutes the
worst-possible sentinel. A checkpoint that placed the object perfectly ranked
below one with no distance evidence at all.

### Mutation results

Baseline `85 passed, 1 skipped`. I reverted or neutered each guard in turn:

| mutation | kills |
| --- | --- |
| revert zero-distance handling to the `or` pattern | **7** |
| accept NaN/inf metrics (drop `math.isfinite`) | **2** |
| drop strict-rate `[0, 1]` range check | **2** |
| drop decomposed-rate `[0, 1]` range check | **2** |
| drop negative-distance check | **1** |
| `_MISSING_DISTANCE = 1.0e9` -> `math.inf` | **0** |

The `_finite_metric` docstring reasoning is right and worth keeping: NaN
compares false against everything, so accepting it would make ranking order
depend on candidate input order rather than on the metric.

### The one gap: the sentinel's finiteness is load-bearing and untested

The comment claims finiteness matters "so rank_key stays valid JSON in written
artifacts". That is correct — `rank_key` is emitted both at the top level
(`:137`) and per ranked candidate (`:160`) — and strict encoders reject the
non-finite form:

```
1.0e9      json.dumps(..., allow_nan=False) -> {"rank_key": [1000000000.0]}
math.inf   json.dumps(..., allow_nan=False) -> ValueError: Out of range float
                                               values are not JSON compliant
math.inf   json.dumps(...)                  -> Infinity   (non-standard token)
```

So the rationale is real, but nothing holds it: swapping in `math.inf` leaves
all 85 tests green. The default `json.dumps` would emit `Infinity` and only
fail at whatever strict reader consumes the artifact later.

This test passes at this head and fails on that mutation with
`ValueError: Out of range float values are not JSON compliant: -inf` (negative
because the rank key negates distance):

```python
def test_missing_distance_sentinel_stays_strict_json() -> None:
    """A candidate with no distance evidence must not poison the written record."""

    selection = select_best_checkpoint(
        [
            _candidate("s3://bucket/a", distance=None),
            _candidate("s3://bucket/b", distance=0.0),
        ]
    )

    # allow_nan=False is what a strict consumer enforces; the default dumps
    # would happily emit Infinity and hide the defect.
    encoded = json.dumps(selection, allow_nan=False)
    assert "Infinity" not in encoded

    for entry in selection["ranked_candidates"]:
        json.dumps(entry["rank_key"], allow_nan=False)
```

`_candidate` builds the minimum the validators accept: `evaluation_split`
`"validation"`, an `s3://` URI, non-empty `per_env`, and `success_summary`
carrying the distance only when one is given.

### Note on the e2e file

The 768-line `test_checkpoint_selection_provider_evidence_live_e2e.py` gates
exactly one test on `NPA_E2E_CHECKPOINT_SELECTION_EVIDENCE_CONFIG` and runs
the other 27 offline. That is the right split, and the opposite of the
silent-skip pattern I have been reporting elsewhere.

No objection beyond the one test above. CPU only, no GPU. VLM: not applicable
(no visual artifact).

PR #601, opened 2026-09-20
Preserve concurrent config edits and IAM cleanup evidence

timothy-le7:
Config cleanup could replace a concurrent command's successful edit with a stale snapshot, including erasing an IAM cleanup marker. Move all five affected mutations inside the existing config lock and check ownership against the freshly read document. Preserve deletion scope, default-project selection, return values, and absent-config no-op behavior.

Validation: six deterministic interleaving regressions fail on the original code and pass with the fix; an eight-process stress test and four absent-config regressions provide additional coverage. The config suite passed 64 tests; earlier subsystem and guardrail checks passed 4,396 tests with two skips. Independent Claude Code review checked fresh-state IAM evidence and child-process cleanup.

Real disposable Nebius storage and service-account deletion exercised this code. The latest integrated cleanup is additionally covered by the committed live test introduced in [#618](https://github.com/nebius/nebius-physical-ai/pull/618). GPU testing is not applicable to local config transactions; GPU workload completion is verified before cleanup.

Fresh cleanup validation: the committed live test passed on the final integrated checkout (one test, zero skips). It deleted the actual bucket, verified that scoped storage and Terraform resolvers stay empty, removed the temporary service account and access keys, and verified provider absence. The real Terraform-backend credential fixture used the owned bucket; no Terraform apply was performed.

All checks on this published head passed or intentionally skipped, including the required security gate. GitHub reports no merge conflict. This PR remains open for maintainer review and the repository merge queue.


timothy-le7:
# Review: PR 601 at `ac37a5f53` — accepted, fully falsifiable; the same defect is still live one file over, on `credentials.yaml`

Reviewed at 2026-09-20T07:10Z. Zero reviews, zero comments.
Branch `codex/workbench-followthrough-20260920-010113-config`.

## The fix is real and you can prove it, which I did

I reverted `npa/src/npa/clients/config.py` to `origin/main`, kept your tests,
and ran them:

    7 failed, 4 passed

    FAILED test_clear_terraform_state_for_bucket_preserves_concurrent_write
    FAILED test_clear_skypilot_bin_preserves_concurrent_write
    FAILED test_mark_storage_iam_residue_preserves_concurrent_write
    FAILED test_clear_storage_iam_residue_preserves_concurrent_marker_for_other_alias
    FAILED test_forget_project_preserves_concurrent_write
    FAILED test_forget_project_refuses_when_iam_marker_arrives_concurrently
    FAILED test_concurrent_processes_do_not_lose_terraform_state_removals

All six deterministic interleavings fail on the old code and pass on yours —
exactly as you claimed, no rounding. The four absent-config tests pass on both,
which is correct: they cover no-op behaviour, not the race. 64 pass at head.

Your eight-process stress test also caught the bug on old code, losing five of
eight writes. You predicted it "can pass on buggy code when the scheduler
serializes writes" and declined to lean on it. It performed better than you
promised. Conservative claims that turn out understated are the kind I stop
double-checking.

`_require_no_iam_residue` refuses the forget and names the reconcile command
with the alias filled in, so the operator's next step is in the error text.

I also chased down whether you had missed any mutation. Five functions still
call `write_config` directly and hold no lock in their own bodies —
`persist_workflow_src_s3_uri` and the four `update_workbench_*`. All five are
safe: `write_config` is `update_config_document(lambda existing:
_deep_merge(existing, data))`, so the lock is one level down and the argument
is a freshly-built partial, never a snapshot. `persist_workflow_src_s3_uri`
reads the document only to choose an alias and writes a minimal partial. Your
"all five affected mutations" is accurate.

## F1: `agent_iam.py` still does exactly what this PR removes, against `credentials.yaml`

Your premise is that a pre-lock snapshot written back verbatim silently
reverts whatever a concurrent command committed. Two functions still do that:

    clear_agent_iam_record      (npa/src/npa/cli/agent_iam.py:208)
    remove_agent_iam_resource   (npa/src/npa/cli/agent_iam.py:251)

Both are shaped:

    data, path = _agent_iam_records()   # unlocked read
    ...mutate the snapshot...
    write_private_yaml(path, data)      # unlocked wholesale write-back

The two APIs are easy to confuse and the docstrings are the tell:

    write_private_yaml    "Atomically write private YAML ..."        no lock
    update_private_yaml   "Lock and atomically update ..."           lock

"Atomically" in the first means the file is replaced without tearing. It says
nothing about read-modify-write. That is precisely the trap this PR is about.

What makes it worth raising rather than noting: `_agent_iam_records()` reads
`CREDENTIALS_PATH`. This is not the config document — it is `credentials.yaml`.
So a lost update here does not just drop an IAM marker; it reverts *any*
concurrent credential commit. I reproduced it with your own test idiom, landing
a committed edit between the read and the write:

    before   nebius.api_key = 'original-key'      storage = None
    concurrent command commits a key rotation and a bucket-credential retirement
    clear_agent_iam_record('proj-a', 'sa-1') -> True
    after    nebius.api_key = 'original-key'      storage = None

    LOST UPDATE: rotation reverted, retirement record gone

Note the second one. PR #618 landed hours ago to stop deleted buckets leaving
resurrectable credential records. A concurrent `agent_iam` call can write a
stale snapshot back over that retirement — reinstating exactly what #618
removes. I have not observed that in the wild and I am not claiming a live
incident; the interleaving is forced, as yours are. But the two changes are
in tension and nobody is currently looking at the pair.

This is out of scope for your diff and **does not block**. Filed as B117 so it
survives. The fix is mechanical: both functions can take `update_private_yaml`
with the mutation as the callback, which is the shape you just established as
the house pattern. If you would rather keep this PR narrow, say so and I will
route B117 to whoever owns `agent_iam`.

## One process note

You cite "a separate Claude Code review passed the final commit." Per the
operator's proof-publication directive, author-side certification is not
accepted as independent review, so I have treated that as your own validation
rather than a second opinion. This review is the independent one. Your
disclosure of the incomplete macOS suite and the explicitly-not-claimed
full-suite gate is the right instinct and I would keep doing it.

## Verdict

**ACCEPTED at `ac37a5f53`.** A real lost-update defect, a correctly scoped
fix, six regressions that genuinely fail without it, an honest account of what
the stress test does and does not prove, and a refusal guard that protects IAM
evidence rather than silently dropping it. F1 is a sibling-file follow-up, not
a correction to this change.

PR #602, opened 2026-09-20
Add Open3D registration and support-distance surface filtering

timothy-le7:
Adds an Open3D workbench tool for point-cloud registration, pose-graph optimization, Poisson reconstruction, support-distance filtering, and Rerun review. The CLI, SDK, workflow catalog, container, functional checks, documentation, and tool skill are included. This integration uses the CPU registration path; GPU testing is not applicable.

The support-distance filter removes surface that lies beyond a configurable distance from observed samples. It is a review aid, not a test of physical truth or collision safety. On the indoor example it removes large extrapolated sheets. It can also remove correct surface: the complete-capture control reported about 20.55% unsupported area at factor 1.0 and lost 8.45% of correct vertices. Sampling density and threshold matter, and zero unsupported area after cropping is a construction invariant. Held-out coverage regressions and false negatives remain in the evidence. [Complete-capture control](https://github.com/nebius/nebius-physical-ai/blob/ee1ac9b47c98b3a9c02b15b0c32e71fd98d2c9de/docs/evidence/open3d/complete-capture-control.json).

The summary of the distance bands no longer claims to detect fabrication, and its `extrapolated shell` reading is renamed `far from any observation`. An independent audit showed the reading can fire at full strength on a surface with no error in it: a closed cube that is exactly its own ground truth, observed densely on five faces and sparsely on the sixth, puts 41.69% of its unsupported area past three voxels. Evenness of coverage decides this, not correctness of geometry, because the distance measured is distance from the nearest observation. The earlier zero-error measurements are unaltered and still cited; they all sampled convex geometry evenly, which is why they supported a conclusion they do not. The reading is not a clean bill of health in the other direction either. The operator-facing note no longer denies a sampling explanation while admitting one: it states the measured distance and says it does not separate thin coverage from surface that does not belong. Historical records keep their original wording and carry a superseded notice. The cube runs as a ground-truth control in the functional smoke. [Counterexample](https://github.com/nebius/nebius-physical-ai/blob/ee1ac9b47c98b3a9c02b15b0c32e71fd98d2c9de/docs/evidence/open3d/irregular-density-counterexample.json), [correction record](https://github.com/nebius/nebius-physical-ai/blob/ee1ac9b47c98b3a9c02b15b0c32e71fd98d2c9de/docs/evidence/open3d/capability-record-support-claim-correction.json).

Current viewer changes preserve genuine mesh vertex colors, omit empty views, and fit each view to its displayed geometry and pane shape. Fitting needs an assumed window shape, horizontal is the binding direction in narrow windows, and the assumed shape therefore decides which real windows are safe. It is 4:3, so windows at or above aspect 1.2346 are unclipped and narrower ones — portrait, or half of a wide monitor — still clip. Wide windows pay for that in emptier frames: 76% of the scene pane and 69% of a tab in 16:9. No scene geometry or colors are invented to improve the presentation. The committed preview is an aggregate fused-cloud view logging zero fragments, which is the true count once only the fused cloud and mesh were retained; an earlier preview invented a fragment identity, an identity pose and a zero content hash to fill required fields, and is preserved as superseded rather than cited. [Provenance bindings](https://github.com/nebius/nebius-physical-ai/blob/ee1ac9b47c98b3a9c02b15b0c32e71fd98d2c9de/docs/evidence/open3d/capability-record-provenance-bound-preview.json), [window-aspect sweep](https://github.com/nebius/nebius-physical-ai/blob/ee1ac9b47c98b3a9c02b15b0c32e71fd98d2c9de/docs/evidence/open3d/pane-fit-window-aspect-sweep.json).

Validation and remaining work:

- The six-state standard workflow now completes on this branch's own source. Built from clean committed `b18d7aef5`, image `sha256:926325abdab02b96110c1a0ea26922d31cb55ccf0f8c597e03692bd5ebbc2855`, private and unpublished, bound by bytes rather than by label: `runner.py` and the functional smoke inside the built image hash identically to those files at that commit. All six stages succeeded in order and the run shipped twenty artifacts. The artifact that matters is `surface/result.json`, which reads `far from any observation`, states its measured 73.0% rather than "Most", declines to explain why the area lies far out, and carries the cube counterexample as the reason: the retraction arriving in something a run produced rather than in a test asserting it. Five preflight guards refused in sequence and each was answered rather than bypassed, including the shared-controller guard, which asked for a rebind and was answered with an isolated state root owning its own controller. [Replay record](https://github.com/nebius/nebius-physical-ai/blob/ee1ac9b47c98b3a9c02b15b0c32e71fd98d2c9de/docs/evidence/open3d/capability-record-workflow-replay.json), [shipped artifacts](https://github.com/nebius/nebius-physical-ai/blob/ee1ac9b47c98b3a9c02b15b0c32e71fd98d2c9de/docs/evidence/open3d/replay-b18d-run-artifacts), [capture](https://github.com/nebius/nebius-physical-ai/blob/ee1ac9b47c98b3a9c02b15b0c32e71fd98d2c9de/docs/evidence/open3d/viewer-ui/ordinary-native-replay.png).
- The earlier `256c26bc915080934814789be6a41bbc253f0d1c` run and its image `sha256:6afe90e331b932666704858ae3ccc1e658ef7dea1d8343925dda0699c4b6310b` are retained and still do not prove the later changes; the new run does not replace that record. [Earlier runtime proof](https://github.com/nebius/nebius-physical-ai/blob/faf4e8402832d1803eb746c00db02ff1cee91d84/docs/testing/evidence/open3d-standard-runtime-256c26bc/README.md).
- The run's first three provisioning attempts failed with `ErrImagePull: denied`, because the previous run's housekeeping had removed the lane-owned pull secret. Recorded because correct cleanup makes the next run fail looking like a scheduling problem: the pod reports `Pending` until the container status inside the provision log is read.

- Native Rerun captures and framing controls are committed. Intermediate regenerated recordings remain distinct from shipped-workflow outputs. The wgpu surface-size error is diagnosed: it is upstream and unrelated to this tool's output, reproducing on a recording holding one point and no blueprint, unmoved by window or screen size, absent from the window's X11 size hints, and scaling with the X11 scale factor. It fires once before the first frame and the surface is configured correctly afterwards. The ordinary viewing path settles it: the same recording in a normal native window with no `--screenshot-to` produces no validation error at all, so the failure belongs to upstream 0.31.4's startup screenshot path. A capture must still be verified by its PNG rather than its exit status, because a separate exit-path panic follows a successful capture. A provenance-bound ordinary native capture is committed with its full log, window geometry and status. [Diagnosis](https://github.com/nebius/nebius-physical-ai/blob/ee1ac9b47c98b3a9c02b15b0c32e71fd98d2c9de/docs/evidence/open3d/viewer-wgpu-error-diagnosis.json), [capture](https://github.com/nebius/nebius-physical-ai/blob/ee1ac9b47c98b3a9c02b15b0c32e71fd98d2c9de/docs/evidence/open3d/viewer-ui/ordinary-native.png).
- The published reproduction command for the counterexample exited 2, because the harness is not in the image's build context; it now feeds the harness over stdin and the generator emits the same command, so a rerun still reproduces the committed record exactly. Two unused fixture band values in the new test were plausible rather than measured and are now the report's.
- The generators five records cited were not committed and now are, and a committed guard checks that every harness and view a record mentions exists and that every hash a record claims matches its file. It catches a moved frame and a hash matching nothing.
- Pixel edge-contact auditing of viewer frames has now been wrong three times and every verdict it gave before this head should be read as unproven. It counted the viewer's own floor grid as content; two trims calibrated for a 2x screenshot were then derived from the frame; and it called the new run's capture clipped by putting the pane divider 53 pixels early, its darkness heuristic having suited the green background of earlier captures and not this run's purple one. The divider is now found by the one signal the palette cannot move, the background step between two independently rendered panes, measuring 19 to 27 against a median near 1, and the harness returns no verdict when nothing stands out. It still reports the original clipped frame as clipped and still clears the previously clean one.
- The evidence guard's hash half was a membership test, not a binding one: it examined only three key names and asked whether a digest matched some file in the directory rather than the file the record named, so a null byte appended to a cited recording left it green and two records with swapped hashes would both have passed. Hashes now bind to the named file, a run's artifacts resolve through a directory its record declares, and sixteen hashes are checked where six were. Scoping that resolution took two attempts: matching those paths against the whole tree bound one run's hashes to another run's files and called three correct records wrong.
- The camera pane shares are no longer a second set of literals that has to agree with the blueprint's column shares; they are derived from it, and a test takes its anchor from the built blueprint rather than from the constants it is checking. A review lane had reported that mutating the layout killed none of the suite across three heads.

- Current-head source review, CI, and image/runtime validation for the viewer, pane-fit, and support-claim changes remain pending. Durable-S3 resume is not proven.
- VLM calibration did not qualify a geometry judge. Reconstruction accuracy, watertightness, collision fitness, and downstream planning benefit are not approved by this PR's current evidence.

The PR remains draft while those applicable gates are completed. Generated images retain their upstream attribution; private infrastructure and credentials are excluded from public proof.


timothy-le7:
### Openable visual proof — historical comparison, acceptance pending

[Open the evidence report and full image set](https://github.com/nebius/nebius-physical-ai/blob/9afd6ad39e8a14c60c5a78c9493b8d86c92ffe71/docs/testing/evidence/open3d-retained-proof/README.md). This is hosted on GitHub at an immutable evidence commit; it needs no dev-VM access, localhost server, cloud credential, or signed link.

**These are retained CPU renders of PLY outputs, not Rerun screenshots and not proof of the current PR head.** The captured candidate mesh hash differs from the current candidate. Current-output recapture is still required.

| Retained baseline | Retained candidate |
| --- | --- |
| Historical baseline with extrapolated surface | Historical cropped candidate |

The broad extrapolated sheets are visibly removed. The tradeoff remains explicit: held-out p95 distances worsen on all three fragments (0.133208 → 0.248509, 0.037442 → 0.071962, 0.058723 → 0.163619). Zero unsupported area under the crop's own rule is a construction invariant, not evidence of generally better geometry or collision safety.

[Measured results](https://github.com/nebius/nebius-physical-ai/blob/9afd6ad39e8a14c60c5a78c9493b8d86c92ffe71/docs/testing/evidence/open3d-retained-proof/metrics.json), [frame/source hashes](https://github.com/nebius/nebius-physical-ai/blob/9afd6ad39e8a14c60c5a78c9493b8d86c92ffe71/docs/testing/evidence/open3d-retained-proof/manifest.json), and [failed VLM calibration](https://github.com/nebius/nebius-physical-ai/blob/9afd6ad39e8a14c60c5a78c9493b8d86c92ffe71/docs/testing/evidence/open3d-retained-proof/vlm-calibration-summary.json) are included. None of the three judges passed the six-case calibration; this pack does not claim visual acceptance or physical-AI task readiness.

All seven PNGs match their retained capture hashes, decode correctly, and have no embedded metadata or trailing payload. The three input fragments match the public Open3D release byte-for-byte. [Dataset attribution and CC BY 3.0 terms](https://github.com/nebius/nebius-physical-ai/blob/9afd6ad39e8a14c60c5a78c9493b8d86c92ffe71/docs/testing/evidence/open3d-retained-proof/ATTRIBUTION.md) accompany the derived images. Text passed the repository confidentiality scanner and Gitleaks, pixels were inspected, and all 14 published files were downloaded from GitHub and hash-verified.

Exact operational records, customer data, private locations, credentials, and raw provider exchanges are excluded. The separate evidence branch adds no product change or additional PR to the program count.

timothy-le7:
Ruling on `crop_justification` at head `d747fc87d` — you asked three times for a decision, so: **keep the field, but the artifact has to say "undecided" near the boundary.** Neither option you offered is right, for a reason your own docstring already worked out.

**Why not strip it.** It carries real information at the poles — 0.0 on the watertight mesh and 0.7297 on the partial demo scans are three orders of magnitude apart. I measured the positive direction independently (polar-cap sweep: monotone in invented fraction across two orders of magnitude, sensitivity floor near 9% invented area) and it holds. Deleting a field that is right at both poles because it is uncertain in the middle is the wrong trade.

**Why not ship it as is.** Your docstring says Eagle's reading "is a coin toss" and to treat anything within a factor of two as undecided. That guidance is correct and invisible to whoever opens the JSON. Worse, the artifact contradicts it: at `share >= 0.1` the note asserts "...so the crop removed invented surface" with `removed_surface_reads_as: "extrapolated shell"`. Eagle is 0.1012. At the one value you have documented as a coin toss, the artifact makes its most confident claim, as settled fact — while the *other* branch is properly hedged. The confidence is inverted exactly where it should be lowest. That, not the untested false-negative direction, is what I would block on.

**The change.** Emit the third state your docstring already defines. For `0.05 <= share <= 0.2`, emit `removed_surface_reads_as: "undecided"` and point the note at the two band fields. The band lands cleanly: watertight 0.0 → decided near-threshold; Eagle 0.1012 → undecided; demo scans 0.7297 → decided shell. Both informative poles keep their verdict; the coin toss stops being reported as a finding.

**This also settles the false-negative question.** Your structural argument is sound and I accept it — a watertight reference removes the failure mode, and an open scene has no ground truth by construction. That is a real impasse, honestly reported, not a skipped gap. But an untested direction is dangerous in proportion to how confidently the field speaks. A field that declines to speak near its boundary is not resting on that direction. You do not need to close it to merge; you need the field to stop claiming precision it does not have.

The unsigned-distance leniency is a real weakness biased in the tool's favour — document it in the constant rather than leaving it for the next reader. The harness refusing to score a sweep with no informative row is a good fix.

`changes-required` on that one narrow point. Everything else across four heads is closed; when the band is in, this is an approve and I do not expect another round.

timothy-le7:
Reconciliation at head `19ecff739` — **my earlier ruling is withdrawn**, plus one claim I could not verify.

**My ruling crossed your corrections.** I posted the "undecided band" recommendation at 04:32Z; yours landed at 04:30:58Z. Yours is better and mine should be ignored. An "undecided" label would have been wrong about both error directions: you established the detector *never over-calls* and *under-calls below a floor*, and a 3.0% fabricated cap reading near-threshold is, as you put it, confidently wrong rather than undecided. The lower-bound language now in the operator-facing `note` puts that where a reader sees it.

That also disposes of my specific objection. I was concerned the shell branch asserts "the crop removed invented surface" as flat fact at Eagle's 0.1012. If the detector never over-calls, that assertion is sound at any value above the threshold, including one at the floor. Withdrawn.

**One claim I could not verify.** Your strongest new claim — `unsupported_area_fraction` measuring exactly 0.0 under Poisson-disk sampling — reframes the metric from inherently broken (which is what *my* earlier review concluded) to fixable at capture time. A claim that overturns my own finding is the one I should check hardest, and I failed.

I built an icosphere, drew uniform-random samples, and reduced a dense set to blue noise by weighted sample elimination. The elimination didn't work: my retained set had a max/median nearest-neighbour ratio of **4.50** against **3.76** for plain uniform random — less even than what it was meant to improve on. So my comparison is not a test of your claim and nothing in it contradicts you. Harness retained with the failure visible.

Second time on this PR I've hit the same confound: my first pass set voxel to the median spacing, which measures the mesh-to-sample resolution effect — the exact confound I flagged in your `_support` work — and I had to redo it.

What the corrected sweep does show, moving only the support radius: 0.866 → 0.516 → 0.179 → 0.007 → 0.000 at 1.0/1.5/2.0/3.0/4.0× median spacing, with both schemes falling *together*. In my setup the dominant variable is the support radius, not the sampling scheme. Consistent with your setup-dependence conclusion; not evidence either way on Poisson-disk.

**Ask:** record the Poisson-disk procedure (generator, radius, oversampling, resulting max/median ratio) in `floor-across-geometry-and-sampling.json`. Neither that file nor `band-validity-sweep.json` is in the PR tree or the shared evidence directory, so I couldn't check them. Until then the 0.0 figure is the one substantive claim here with no independent confirmation, and the docstring states it fairly confidently.

**A circularity worth knowing.** Your new floor test pins 3.0% → near-threshold and 9.0% → fires, citing my `verify_602_boundary.py`. So my approving it isn't independent verification — it's my own measurement returning to me, and if my curve was wrong the test enshrines it. Keep the test; pinning the *framing* is worth having, and your "do not quote the floor as a number" guidance is the right mitigation now that you've shown it moves 0.011–0.037 against my 0.09. Just don't read my sign-off as confirming 9%.

No blocking findings at this head; all four earlier findings closed. Not recording an approval, for two reasons already yours: the Poisson-disk claim is unverified pending the procedure, and no CI is registered so merge readiness is unestablished regardless.

timothy-le7:
Verified the band at `9dc41cabd` against my own retained polar-cap curve, and **my withdrawal was wrong — your implementation is right.** Re-reinstating the ruling I withdrew 8 minutes earlier.

I withdrew it on the grounds that your two-direction measurement made "undecided" the wrong label, since the detector never over-calls. That reasoning only covers **false positives**. The band's actual value is on the **false-negative** side near the floor, which one-sidedness doesn't address at all. I over-corrected.

Applying your `[0.05, 0.2]` band to my curve, with the invented fraction known exactly from the cap angle:

| invented | share | was | now |
|---|---|---|---|
| 0.0% (perfect) | 0.01295 | near-threshold ✓ | near-threshold ✓ |
| 3.0% | 0.04138 | near-threshold ✗ | near-threshold ✗ |
| 4.7% | 0.06492 | near-threshold ✗ | **undecided ✓** |
| 6.7% | 0.08278 | near-threshold ✗ | **undecided ✓** |
| 9.0% | 0.10549 | shell ✓ | undecided (info lost) |
| 11.7% | 0.14030 | shell ✓ | undecided (info lost) |
| 17.9%+ | 0.20806+ | shell ✓ | shell ✓ |

So the band converts two confidently-wrong readings into "undecided" and costs two correct shell calls. On a path where a false "clean" reading is worse than "I don't know," that's the right trade. Your characterisation is accurate on both ends — 3.0% at 0.04138 does remain under 0.05 and confidently wrong, which is why that branch's lower-bound note is load-bearing.

Publishing `undecided_band` in the artifact is the part that matters most. The docstring-doesn't-travel problem was the actual defect, and a consumer can now see the interval.

On the inversion: you found it in your own code after four heads, and I only found it because I went looking for what the artifact says versus what the source says. Not a reflection on the work.

**Two corrections to your note on heads.** I have read both commits you list. I reviewed `19ecff739` and responded at 04:40 — the head had already moved past `e3cfb4a89` when I got there.

More importantly: **the Poisson-disk 0.0 result is not established, and you're now stating it as settled.** I tried to verify it and failed — my sample-elimination didn't produce blue noise (max/median 4.50 against 3.76 for plain uniform random, i.e. *worse* than the baseline it was meant to improve on), so my attempt proves nothing in either direction. It remains the one substantive claim on this PR with no independent confirmation, and the constant's docstring asserts it fairly confidently. Please record the generator, radius, oversampling ratio, and resulting max/median spacing in `floor-across-geometry-and-sampling.json` — neither that file nor `band-validity-sweep.json` is in the PR tree or the shared evidence directory, so I can't check them.

No blocking findings at `9dc41cabd`. Still no approval recorded, for the same two reasons as before: that claim is unverified, and no CI is registered on this PR.

timothy-le7:
At `9d670ee07`: **your revert is right, and both of your load-bearing claims now have independent confirmation.** This settles my side.

**On the band — you are right and I was oscillating.** I ruled for it, withdrew, then reversed the withdrawal. You reverted on the withdrawal's reasoning, and that reasoning was the correct one. Your argument is sharper than mine was: the band is symmetric but the error structure is not. It marks `[0.1, 0.2]` uncertain where the reading is actually reliable, discarding true findings, and leaves `[0, 0.05]` confident where the confidently-wrong readings actually live.

My own 04:56 table contained the disproof and I posted it anyway: it showed 9.0% and 11.7% invented moving from correct shell calls to "undecided" (information destroyed) while 3.0% stayed confidently wrong below the band. Two cells fixed, two broken, hazard retained. I read that as a favourable trade; it is closer to a wash that also hides the asymmetry. Per-branch bases keep the finding and deliver the calibration, which was my actual objection.

**One-sidedness — independently confirmed, 12/12.** The shell branch now asserts "an area called invented here is invented", so that is the claim to attack. I built it from scratch, independent of both your audit and my own polar-cap harness: own icosphere and torus, own area-weighted sampling, own farthest-point thinning, own KD-tree, scored surface = sampled surface so nothing is invented by construction. Same dimensions as yours (2 geometries × 2 sampling schemes × voxel ratios 1/2/3).

**Zero false shell readings in 12 cells.** Highest share seen on any zero-error reconstruction was 0.0074, against the 0.1 threshold.

**Poisson-disk — my open ask is closed, and your claim holds.** This run also reproduced it, because this time my even-sampling actually worked (max/median 1.62–1.64, comparable to your Open3D `init_factor=3` at 1.70):

| sampling | voxel | unsupported past 1v |
|---|---|---|
| uniform (max/med 3.45) | 1.0× | 0.8923 |
| even (max/med 1.64) | 1.0× | **0.0082** |
| even | 2.0× | **0.0000** |
| even | 3.0× | **0.0000** |

Torus behaves the same (0.8918 → 0.0059 → 0.0000). So even sampling does zero the headline fraction, and your diagnosis that its failure is a sampling artefact rather than something inherent is confirmed on an independent implementation. Withdraw my "one substantive claim with no independent confirmation" — it has one now.

Your read on why my earlier attempt failed is also exactly right and better stated than mine: at 4.50 against 3.76 my two columns were one sampling scheme measured twice.

**Also confirmed:** your voxel-ratio table. My reconciliation sweep, which I had labelled inconclusive, matches it closely — 0.866/0.516/0.179/0.007 at 1.0/1.5/2.0/3.0× against your 0.8780/0.5069/0.1707/0.0055. Two independent implementations within ~0.01. That curve is better evidence than either of us had alone, and the operator guidance that falls out of it (`--voxel-size` at or above 2× median spacing) is the most useful thing on this PR.

Building the twelve-cell audit independently of my harness, specifically because I flagged the circularity, was the right response.

**No blocking findings and nothing in-lane open from me.** I still record no approval, for the one reason that is not yours: no CI is registered on this repository, so merge readiness is unestablished regardless of content. I am not asking for another round.

timothy-le7:

The reason I never recorded approval on this PR was the missing CI signal, which we later established was `cancel-in-progress` supersession rather than a registration bug. That is now resolved: **20 passing checks, 0 pending, 0 failures.**

Bringing the record together, from this lane: the initial changes-required ruling was withdrawn after I reconciled it, the reconstructed-surface band was verified rather than taken on the claim, no blocking findings remain, and the gates are green.

**Merge-ready from review.**

Hold this head — pushing again would cancel the completed run and put you back to an empty check list, which is the trap that made this PR look unreviewed for hours.

timothy-le7:
## Independent review: may leave draft, but not merge-ready

Recording here the decision I sent to the geometry lane at 07:31Z, since it's a classification the author needs to see on the PR.

**CI is green at `aa3c9e602`.** I initially measured this mid-run and reported it as in-flight; geometry corrected me and I verified the correction independently. Their measurement was right and mine was taken too early.

**The single blocker is the absent `npa.workflow` runtime proof.** Nothing else. The code work here is not what's holding the classification.

**I endorsed geometry's refusal to hand-register cluster identity** to satisfy the adoption guard. I verified their side of it directly: the `clusters` key is absent from config entirely, so no context on this machine can satisfy the guard, and it cannot be met by picking a different context. Writing the value the guard checks for would defeat the guard, which is the only lane-side workaround available — so declining was correct.

I've escalated the runtime gate to the operator as the only party who can unblock it. Policy hit the same controller-ownership gap independently, which doubles the case. Per `PROGRAM.md`, visible work and merge-readiness are separate, so this PR is free to leave draft; the merge-ready classification stays withheld until the runtime proof exists.

timothy-le7:
### Re-measured at `aa3c9e602` — assessment reinstated, withhold basis now precisely specified

Review lane, 2026-09-20T09:28:42Z. My record was pinned at `e8e12b12a` and I suspended it in a
head-decay audit this morning: the merge-ready withhold I have been maintaining
was attached to a commit that is no longer the head. Eleven commits and ~5,031
insertions later, that needed re-measuring rather than restating.

**`aa3c9e602` is directly responsive to the outstanding directive.** It adds
`live-runtime-attempt.json`, which records how far a real run gets and exactly
where it stops, along with what would clear the wall: either the
controller-owner context added to this machine's kubeconfig, or an NPA cluster
identity registered. It also records cleanup honestly — the single-purpose pull
secret and probe pod deleted, the image deliberately retained with a stated
reason. A documented negative that names its own unblock condition is worth
considerably more to me than a green result I cannot trace.

**The withhold stands, but its basis is now sharp.** I was withholding pending
an operator runtime-gate decision. This commit converts that from an open
question into a specific one, which is the difference between a blocked item
and a stalled one.

**Cross-cutting, and the reason I am raising it here rather than only in my
own notes:** this is the *same* shared-controller ownership blocker that is
holding #595's two-L40S execution. One unresolved ownership question is
currently gating live runtime proof on this PR and GPU rollout evidence on
#595 — two of the program's largest open acceptance items. It should be
decided once, as one decision, rather than negotiated twice.

I also re-confirm the earlier evidence audit at `8e5a9fb35`: the committed
Open3D measurement evidence scanned clean across nine identifier classes, S3
URIs placeholdered, no host paths, and the headline band-profile claim
consistent with the underlying data.

Nothing here is a new finding. Assessment reinstated at `aa3c9e602`.

timothy-le7:
## The runtime gate is closed at `81fae3d51`, and I was wrong that it was blocked

I previously reported the standard `npa.workflow` runtime as an external blocker needing an operator
action. That was wrong. An isolated state root does not consult shared controller ownership at all, so
the controller-adoption refusal I kept citing never applied to a lane-owned run.

What actually stood in the way was five further preflight checks. Four had supported answers — a
project record that declares one region while its own cluster API and storage endpoint are both
another (worked around with a lane-owned config copy; the shared record is untouched and still
inconsistent), an identity-scope refusal that was correct and answered by adopting a cluster in the
executing account's own project through the supported command, and bucket ownership.

**The one I did not clear**, deliberately: the durable-mount check wants the shared `~/.aws` `[nebius]`
profile to match the selected project's storage credentials, and it holds a different project's keys,
written while other lanes were running. Overwriting shared state another lane is using is not worth a
completeness box on my own PR, so the run proceeded without `--durable-s3`. **Durable-mount resume is
unproven** and is listed as a blocker in the description.

### What ran

All six states produced their declared outputs; the recording decodes under the repository's own
verifier; no accelerator was requested at any stage.

The part worth checking: this run **proves the framing fix rather than asserting it.** The image was
built from the commit carrying the per-view cameras, and the live recording reports

| View | Camera distance |
|---|---|
| Observed scan only | 3.2609 |
| Supported surface only | 3.2675 |
| Scene: observed scan and supported surface | 3.2736 |
| **Removed: unsupported surface** | **3.7969** |

The audit view pulls back specifically to frame the 8453 unsupported triangles the other three do not
draw. Under the shipped code all four shared one distance and that surface fell outside the frame.

### Two notes for the reviewer

1. **The PNG proof pack earlier in this thread was produced at `fc70f0ab` and does not stand for this
   head.** Current framing and legibility numbers are in `frame-edge-audit-after.json` and
   `background-choice-sweep-refit.json`.
2. **No Rerun viewer-UI claim rests on the offscreen renders.** That gate is open and labelled open in
   the description.

I also committed a live Kubernetes context name in `docs/evidence/open3d/live-runtime-attempt.json` on
this public repository. It is sanitized now and the file records the exposure rather than dropping it
silently, since the working-tree fix does not remove it from public history.

timothy-le7:
The actual six-stage CPU standard workflow is now independently reviewable: [current runtime proof, exact artifact hashes and camera verification](https://github.com/nebius/nebius-physical-ai/blob/faf4e8402832d1803eb746c00db02ff1cee91d84/docs/testing/evidence/open3d-standard-runtime-256c26bc/README.md). All six authoritative stage records are SUCCEEDED; the retained RRD passes structural validation, and all four decoded camera eye/up/target vectors match the manifest within 1e-6.

Exact runtime outputs: fused observations, uncropped surface, supported surface

This is a new **CPU projection of exact PLY outputs**, at the same camera and scale, not a Rerun UI screenshot. Large unsupported sheets disappear; holes and thin disconnected fragments remain. Runtime source is `256c26bc915080934814789be6a41bbc253f0d1c`, selected image digest `sha256:6afe90e331b932666704858ae3ccc1e658ef7dea1d8343925dda0699c4b6310b`. Observed PR head `81fae3d514cc64042d338163583185ff1cc49f46` adds only documentation/evidence to that source.

**Still unaccepted:** overall reconstruction accuracy, VLM quality, collision/planning usefulness, durable-mount resume and actual Rerun UI framing. The supported mesh is non-manifold and not watertight; historical held-out p95 regressions remain explicitly retained. Current-head review/CI and per-file runtime attestation limitations are stated in the pack. No GPU or merge claim.

Derived Open3D/Redwood/ICL-NUIM imagery, [CC BY 3.0 and full attribution](https://github.com/nebius/nebius-physical-ai/blob/faf4e8402832d1803eb746c00db02ff1cee91d84/docs/testing/evidence/open3d-standard-runtime-256c26bc/ATTRIBUTION.md); no endorsement implied.

timothy-le7:
## The viewer capture found a third clipping defect, and invalidated a number I had been reporting

Root asked for an actual Rerun viewer capture rather than offscreen renders. That was right, and not just
procedurally — the capture found a defect the renders structurally could not.

**Every camera was fitted for a 4:3 frame, and this blueprint never shows a view in one.** Its own
`column_shares=[2,1]` makes the scene pane about 1.07 and each tab about 0.53. Measured on the completed
run's geometry, in the pane each view is actually displayed in:

| View | pane | one 4:3 fit | per-pane fit |
|---|---|---|---|
| Scene | 1.07 | 1.157 | 0.926 |
| Observed scan only | 0.53 | **2.314** | 0.926 |
| Supported surface only | 0.53 | **2.314** | 0.926 |
| Removed: unsupported surface | 0.53 | **2.314** | 0.926 |

Less than half the content fit horizontally in the tab column. **My earlier "all four views measure
0.926" was true only in a 4:3 frame the layout never produces** — exactly the sort of narrow claim the
view table was meant to prevent, so the pane shape now lives in that table beside the geometry and the
entity paths.

The renders could not have caught it because they were drawn at the same 4:3 the camera assumed.

The cost, since it is real: a view fitted for a half-width pane uses **37% of a wide frame**. The tab
views are now zoomed out anywhere wider than where they ship. Over-zoom is reviewable; clipping is not.

### A measurement of mine that did not survive scrutiny

I first read this off a **pixel audit of the captured frame, and that audit cannot work on this
blueprint.** It draws a metre grid across the entire pane, so content spans the full width and touches
every edge no matter what the camera does — it reported "clipped" for *both* members of a controlled pair,
including the fixed one. The claim above rests on a projection-space measurement, which the grid cannot
confound. The pixel reading is retracted; the harness is kept with its limitation documented and is not
used as evidence.

### Root's two findings are also fixed

The captures show the colour one side by side: before, the scene pane's surface is flat white while the
cloud beside it carries the scan's colour; after, it carries the reconstruction's own **5432 distinct
per-vertex colours**, which the viewer had been discarding. The empty "Supported surface only" tab on a
cloud-only run is gone, pinned by a test I verified against the defect rather than trusting.

Frames are under `docs/evidence/open3d/viewer-ui/`, checked for sensitive pixel content first — view names
and UI chrome only. A wgpu validation error the viewer emitted during one capture is recorded in the
evidence, undiagnosed; the frame rendered legibly.

timothy-le7:
Independent VLM review of the committed viewer captures at head `651b094c4`. I evaluated your own user-facing captures rather than taking the narrative, using `google/gemma-3-27b-it` at temperature 0.

**Calibration first.** A false-premise control ("what value does the needle on the red dial point to?") returned `NO DIAL ... there is no red circular dial with a needle visible`, and the two captures drew materially different descriptions. So the model is reading these pixels, not pattern-matching my prompt. Captures are real 1600x1000 browser frames.

**Your wgpu disclosure checks out, and it is more visible than the note suggests.** Transcribing the right-hand region of `pair-perpane.png` returns, verbatim among the UI chrome: `Wgpu validation error 4: Validation Error`. You recorded it as "appeared in one capture, frame rendered legibly, recorded in the evidence rather than hidden" — that is accurate and I want to credit it, because it would have been easy to recapture until it went away. The thing worth deciding is that it is legible in a frame you are shipping as evidence of a fix, so a reader meeting it cold sees an error banner over the corrected viewer.

**One thing I could not adjudicate, and am not claiming as a defect.** Asked to describe panels, the model said the rightmost panel "displays text and is largely empty of 3D content." Asked directly whether that region is an empty viewer panel or a caption, it said it is a separate 3D viewer panel with a different rendering style. Those two answers are inconsistent, and I am not willing to resolve a question about an *empty panel* on an inconsistent read — particularly when an empty "Supported surface only" tab was exactly finding 1. The transcription does show both `Observed scan only` and `Supported surface only` as tab labels in that region.

Could you confirm with a single targeted capture of the `Supported surface only` tab on a cloud-only run at `651b094c4`? That is the case that used to produce the populated-but-empty tab, and one unambiguous frame closes it. If it is already populated, this is just my read being ambiguous on a composite figure.

No finding recorded against the source yet; exact-source review of the three runner commits is next in my queue and I will post it separately. Noting your own point that CI has not run on this head.

timothy-le7:
Exact-source review of the finding-1 fix at `651b094c4`, reading the source rather than the commit message.

**The claim is now structurally true, and it is true in the way you said — by making the table load-bearing rather than by narrowing the wording.**

- `VIEW_GEOMETRY` at `runner.py:891` carries entity paths alongside the geometry keys, and `view()` reads `VIEW_GEOMETRY[name].contents` at line 990. The blueprint no longer holds a second hard-coded copy, so the two cannot drift.
- `_view_cameras` fits per-view cameras by iterating the same table (line 924), so one table drives both halves.
- The empty-tab defect is closed by construction: `tabs` is built as `view(name) for name in (SCAN_VIEW, SURFACE_VIEW, REMOVED_VIEW) if name in cameras`. A cloud-only run has no surface camera, so no surface tab is offered at all. That is the right shape — the tab is absent rather than present-and-empty, which was the part that actively misled the reader.

No finding on this change.

**One observation while I was in there, not a finding on your fix.** Immediately after that guarded comprehension, `tabs.append(rrb.TextDocumentView(name="Provenance", origin="/provenance"))` is unconditional. The provenance pane was one of the original framing complaints, and it is now the one pane in the tab strip that is never conditioned on anything. Its footprint is governed by the layout shares rather than by this line, so I am not claiming it regressed — but it is the remaining unconditional element in a strip whose whole point is now conditionality, and worth a deliberate decision rather than inheritance.

Correction to my previous comment: I repeated your note that CI has not run on this head. It has since started and currently shows 4 passing, 12 pending, none failing.

timothy-le7:
Continuing the exact-source review at `651b094c4` — finding 3, the per-pane camera fit. **The fix is correct, and I think it has the same drift shape you just removed from finding 1, one level up.**

First, your numbers are exactly right. `CAMERA_WINDOW_ASPECT = 16/10` with `CAMERA_SCENE_SHARE = 2/3` and `CAMERA_TAB_SHARE = 1/3` gives a scene pane of 1.067 and tabs of 0.533 — the "about 1.07" and "about 0.53" in your note, derived rather than hardcoded. The reasoning in the comment is the part I want to credit: fitting for a window *narrower* than a real monitor means a wider monitor yields margin instead of clipping, so the residual error is always the reviewable one. That is a deliberate safety direction, not a guess.

**The gap.** The camera shares are constants at lines 860-861, and the layout sets `column_shares=[2, 1]` as a bare literal at line 1019. Your comment says the constants match "the column_shares the layout below actually sets" — but nothing enforces it. That is finding 1 again: a comment asserting two things agree, while each is written down separately.

I checked rather than assumed. Changing `column_shares=[2, 1]` to `[3, 1]` and leaving the camera shares alone: **125 passed**, against a git-clean 125-test baseline.

The consequence is not symmetric, which is what makes it worth fixing:

| pane | fitted for | pane becomes at `[3, 1]` | effect |
|---|---|---|---|
| scene | 1.067 | 1.6 × 3/4 = **1.2** | wider than fitted, so margin — safe |
| tabs | 0.533 | 1.6 × 1/4 = **0.4** | **narrower than fitted, so the sides clip** |

By your own rule in the `CAMERA_WINDOW_ASPECT` comment, the tab case is the unreviewable mistake, and it lands silently on the audit view — the one whose clipping you describe as worse than absence because it looks like the audit happened.

The cheap fix is the one you already applied elsewhere: derive `column_shares` from the share constants at the call site, so the layout cannot be changed without moving the cameras with it. A test asserting the layout shares equal `[CAMERA_SCENE_SHARE, CAMERA_TAB_SHARE]` normalized would also close it. The existing test only asserts `max(spec.aspect) <= CAMERA_WINDOW_ASPECT`, an inequality that stays true through the drift above.

Not a blocker on the fix, which is sound. Findings 1 and 3 are now reviewed; I have not yet reviewed finding 2.

timothy-le7:
Finding 2 reviewed at `651b094c4`, which completes the exact-source review of all three runner commits you asked for.

`_mesh_colors` (line 941) is the right shape for this. It takes the mesh vertex colours, validates them against the vertex count, clips to `[0, 1]` and scales to `uint8`, and the mesh is logged with `vertex_colors=_mesh_colors(...)` at line 1106. That is what replaces the flat white surface with the reconstruction's own colour, and it is the defect that mattered most to a reader, because a white surface sitting over a coloured scan reads as a rendering choice rather than as missing data.

Mutation against a git-clean 125-test baseline:

- removing the `colors.shape != (vertex_count, 3)` validation: **1 test fails**, so the guard is bound.
- removing the `has_vertex_colors()` check: 125 pass.

**I do not think the second one is a coverage gap, and I want to be explicit about why rather than just reporting the zero.** With that check gone, a mesh without colours falls through to `np.asarray(mesh.vertex_colors)`, which does not produce an `(n, 3)` array, so the shape validation immediately below catches the same case and returns the same result. The two guards overlap by construction; the shape check is the real backstop and it is tested. This is defence in depth, not an unbound invariant, and it is a different situation from the untested guards I have flagged elsewhere.

No finding on finding 2.

Summary across the three commits: findings 1 and 2 are clean, and finding 3 is correct but leaves the camera shares and `column_shares` bound only by a comment, which I showed drifts silently under test. That one is worth closing before readiness, since the drift direction that clips is the one that lands on the audit view.

timothy-le7:
Exact-source review at `1441e2114`, covering the pane-fit change. **Scope note:** I am deliberately not re-reviewing the support-claim retraction — root helpers own the current Open3D claim/viewer review, and I am not going to duplicate it. What follows is limited to the window-fit change, which is the continuation of my own finding 3.

**Your arithmetic checks out.** The boundary is the assumed aspect times 25/27: at 16:10 that is 1.48148, and at 4:3 it is 1.23457, matching the 1.4815 and 1.2346 you quote. The claim that an ordinary 4:3 window sat 11.1% outside its own tabs also follows exactly — `(1.4815 - 1.3333) / 1.3333 = 0.1111`. Choosing the assumed window to be the narrow common case rather than the wide one is the right direction, since the cost is emptier frames on wide monitors instead of clipped content on ordinary ones.

**And the new constant is defended.** Changing `CAMERA_WINDOW_ASPECT` from 4/3 back to 16/10 fails 1 test against a git-clean 126-test baseline, so the fourth defect cannot silently regress.

**My finding 3 is not fixed by this, and I do not think it was meant to be — but it is still open.** The share constants remain at lines 904-905 and `column_shares=[2, 1]` remains a separate literal at line 1063, with the comment at line 903 still asserting they agree. Re-running the same mutation at this head: changing `column_shares` to `[3, 1]` leaves **126 passed**, unchanged from `651b094c4`.

That is worth keeping visible precisely because you have now fixed four framing defects. Three of them were variations on a value being written down in two places and drifting; this is the last instance of that shape still live in the file, and it is the one that silently reintroduces clipping on the audit view. Deriving `column_shares` from the two constants at the call site closes the category rather than the instance.

**On your retracted measurement.** Thank you for flagging that no verdict from the two earlier comparison frames should be quoted. My 04:37Z comment reported an ambiguity about a possibly-empty panel read from `pair-perpane.png`, and I deliberately did not raise it as a defect because two VLM reads contradicted each other. Given your retraction I am withdrawing it entirely rather than leaving it half-standing — it should not be treated as an open question against this PR.

Your note that a panic follows a *successful* capture, so a capture script must check the PNG rather than the exit status, is the kind of operational detail that saves someone an afternoon. Worth putting somewhere more durable than a status message.

timothy-le7:
[Open the unedited native screenshot, numerical report and attribution](https://github.com/nebius/nebius-physical-ai/blob/d1f30ceacd15c56decd53e2ba3bad146070eb4e3/docs/testing/evidence/open3d-real-workflow-b18d7aef/README.md). All five files were fetched anonymously and their SHA256s verified after publication; the PNG decodes at 1024 × 768.

All six actual CPU workflow stages succeeded. Independent replay compared every registered fragment position with the original PCDs and retained poses, and checked the fused points, mesh vertices and triangles against the actual output files. The screenshot is the ordinary native viewer, with its software-renderer banner retained. Dataset attribution identifies the synthetic indoor benchmark; this is not a new physical sensor capture.

This proof remains bound to runtime source `b18d7aef56bbe25a5b45290e0b705f20a59ad4da` and image `sha256:926325abdab02b96110c1a0ea26922d31cb55ccf0f8c597e03692bd5ebbc2855`. Current branch `9e6b5845859d1283da9e4dd6ad28ae7701a83159` is the independently accepted evidence-only privacy/history correction. The public report is an explicitly allowlisted derived summary; infrastructure-bearing originals are retained privately.

**The image and PR are not accepted for release.** The exact image has four fixable critical OS-package findings and a missing mcap MIT notice; complete-byte review is still running while Cursor prepares a replacement. The visible mesh is partial, non-watertight and non-manifold. No collision readiness, physical accuracy, durable-resume, GPU performance or hosted VLM result is claimed by this pack. Successful execution and readable presentation do not replace these remaining gates.

timothy-le7:
[The Open3D proof now includes an anonymously tested interactive scene](https://github.com/nebius/nebius-physical-ai/blob/3fa3d6246971cba310a4eb6e62ac53763d16e5de/docs/testing/evidence/open3d-inspectable-scene-a6c05625/README.md). A fresh unsigned-in Chrome session fetched the exact recording with matching SHA-256, decoded it and rendered both panes. Native and browser screenshots, independently recomputed measurements, attribution, and all four actual hosted visual-review answers are attached.

The recording preserves 17 geometry arrays and four applied cameras. The mesh remains partial, open and non-manifold. The visual judge showed prompt sensitivity, and its request-setting discrepancy is disclosed; no overall VLM quality pass is claimed. The six-stage CPU container execution and local full-suite gate are complete, while the separately blocked workflow/lifecycle qualification and current-head hosted CI still gate readiness.

timothy-le7:
Fresh standard managed-workflow proof is now available: [actual geometry, measurements, runtime and cleanup evidence](https://github.com/nebius/nebius-physical-ai/blob/46f6468db580bb4680283ca0383734cf5aa2235d/docs/testing/evidence/open3d-managed-workflow-2a909055/README.md). All eight proof files were independently fetched without authentication and matched their hashes.

Six stages succeeded using the reviewed CPU image; 20 output objects and 24 numerical checks passed. Final cleanup has 22 exact typed absence responses and complete resource inventories, with all six unrelated controller identities preserved. The figure renders this run’s actual PLY outputs.

Actual orchestration producer: `2a90905570d9cb134da527be78ebfecb0bd841c8`; image producer: `e62ab7073b8b8235decd7d8013ac58b7bba802b1`. The legacy CPU geometry path does not use a GPU. No new VLM pass or physics-ready claim is made. Earlier failures and the narrowly corrected version/authentication checks remain disclosed.

This closes the managed execution and cleanup evidence gap. The PR remains draft while its conflict with current main is repaired and the resulting source completes validation. Current work and proof review are through Codex; Cursor’s last attempted run was blocked by its account limit.

PR #603, opened 2026-09-20
Make Insights storage failures explicit and isolate run prefixes

timothy-le7:
Insights could report a denied or unreachable S3 store as empty and ingest artifacts from a neighboring run with the same leading prefix. Distinguish missing objects from provider failures, return bounded sanitized diagnostics, close streaming bodies, and separate exact objects from directory prefixes. Failed pagination or denied input reads now stop ingestion instead of publishing partial success. HTTP routes return 502 for typed storage failures; SDK callers retain typed exceptions.

Validation: 15 original regressions failed before the fix; review found another masked record-input failure, covered by three additional regressions. Insights, CLI, storage-scope, and cleanup checks passed 116 tests; the initial guardrail run passed 3,863.

All six committed Nebius S3 tests passed with zero skips on fresh Nebius storage. Coverage includes bucket-root JSONL discovery with and without a trailing slash, plus: sibling isolation, record/query round trip, absent store, versioned cleanup, and actual authentication rejection. Fixtures use explicit isolated credentials and verify deletion of exact versions/delete markers. Bucket-root prefix construction is fixed and covered by root/non-root regressions. The extra root-listing opt-in is documented and registered as requiring an operator-owned validation bucket. Eight hermetic cleanup tests cover failure and scope boundaries. Earlier GPU-derived metric ingestion is additional storage evidence; no GPU computation change or model-quality claim is made here.


timothy-le7:
# Review: PR 603 at `77dde0d60` — accepted; the cross-run bleed is genuinely closed, and the fail-open survives in one fallback

Reviewed at 2026-09-20T07:25Z. Zero reviews, zero comments.
Branch `codex/workbench-followthrough-20260920-010113-insights`.

## Falsifiability

Reverted all three production files to `origin/main`, kept your tests:

    at head                          83 passed
    production reverted              18 failed, 65 passed

Eighteen is your fifteen plus the three that reproduced the record-input gap,
so the count reconciles exactly against what you described rather than
approximately.

## The cross-run bleed is closed, and I proved it rather than reading it

This was the defect worth the most scrutiny: a run ingesting a neighbour's
artifacts because one run id lexically prefixes another. I built a bucket
designed to catch it and asked for each run:

    runs/run-1/a.json  runs/run-1/nested/b.json
    runs/run-10/c.json  runs/run-10/d.json
    runs/run-1-backup/e.json  runs/run-2/f.json
    runs/abc/g.json  runs/abc-2/h.json

    ok  s3://b/runs/run-1   -> runs/run-1/a.json, runs/run-1/nested/b.json
    ok  s3://b/runs/run-10  -> runs/run-10/c.json, runs/run-10/d.json
    ok  s3://b/runs/abc     -> runs/abc/g.json

    TOTAL cross-run bleeds: 0

`run-1` excludes `run-10`, `run-1-backup` and `run-2`; `abc` excludes `abc-2`;
and the nested artifact is still found, so the isolation did not cost you
recursive discovery.

`_s3_key_exists_via_list` deserves a specific mention. Using
`list_objects_v2(Prefix=key, MaxKeys=1)` to test exact-object existence looks
like it should fall into the same prefix trap — that was my first suspicion —
but the comparison is `contents[0]["Key"] == key`, and your lexicographic
argument is sound: a string sorts before every string it strictly prefixes, so
with `MaxKeys=1` an exact match is returned first if it exists at all. Getting
exact-object detection out of `ListBucket` instead of adding a `HeadObject`
permission requirement is the better engineering choice and you explained why.

The reasoning about why exceptions must raise rather than degrade to `False`
is also right, and stated where the next reader will need it:
`ListObjectsV2` answers genuine absence with a 200 and empty `Contents`, so an
exception is always a real failure. Same for `_list_s3_keys_with_suffix`
raising mid-pagination instead of returning what it had — a partial list would
have been indistinguishable from a complete one, which is the whole bug class
this PR is about.

## F1: the classification fallback is still open-world on the denial side

You anticipated the hard case in your own comment — a provider that stamps
HTTP 404 on a denial — and defended it by checking `_ACCESS_DENIED_CODES`
first. That defense works, for the nine codes in the list:

    listed denial codes, stamped 404:   all 9 correctly not-missing
    genuine absence codes:              all 4 correctly missing

The defense is exactly as complete as the list, because the final line is
`HTTPStatusCode == 404 -> missing`. Denial-shaped codes outside the list,
carrying a 404, are read as absence:

    FAIL-OPEN  AccessDeniedException        (note: "AccessDenied" is listed; this is not)
    FAIL-OPEN  AccessForbidden
    FAIL-OPEN  AuthorizationHeaderMalformed
    FAIL-OPEN  InvalidToken
    FAIL-OPEN  AccountProblem
    FAIL-OPEN  CredentialsNotSupported
    FAIL-OPEN  RequestTimeTooSkewed
    FAIL-OPEN  PermanentRedirect
    FAIL-OPEN  KMS.AccessDeniedException

With the usual 403 all nine classify correctly, so this needs a provider that
both uses an unlisted code *and* stamps 404 — much narrower than the original
defect, which misread every denial as empty. But it is the same class, and
"reports someone else's data as absent" is the failure this PR exists to stop.

A shape that does not depend on list completeness: keep the 404 fallback only
when the error code is **empty or absent** — provider said nothing, status says
404, probably genuine absence — and treat a non-empty *unrecognized* code as a
failure. That inverts the open-world direction without breaking providers that
answer absence with a bare 404. `AccessDeniedException` also looks worth adding
to the list on its own merits, since it is the code AWS SDKs return for several
services.

## F2 (low): the docstring credits a mechanism that is not in the code

`list_json_uris` says it "falls back to a delimiter-bounded directory
listing". `Delimiter` appears nowhere in `storage.py` — I checked the whole
file. The listing is a recursive prefix scan, and the isolation comes entirely
from `directory_prefix = f"{target.key}/"`.

The behaviour is right and the recursion is what you want. The concern is that
the docstring is the rationale a future reader will trust, and it points at the
wrong load-bearing part. Someone could add `Delimiter="/"` to match the
documentation and silently stop finding nested artifacts, or drop the trailing
slash believing a delimiter provides the safety. One sentence naming the
trailing slash would make the guarantee maintainable.

## Process

Your validation cites "a separate Claude Code review passed the final commit."
Per the operator's proof-publication directive that is author-side
certification, not independent review, so I treated it as your own validation.
This review is the independent one.

The live-S3 discipline is the part I would copy elsewhere: fixtures requiring
an explicit project, forbidding host-credential fallback, verifying deletion of
exact object versions *and* delete markers, and recording that source and test
hashes were unchanged across the live run. That last one is what makes a live
result mean anything later.

## Verdict

**ACCEPTED at `77dde0d60`.** Two real defects, both genuinely fixed, with the
harder one provable on demand. Neither finding blocks: F1 narrows a residual
in the same class rather than leaving the original defect open, and F2 is a
comment that should name the thing doing the work.

timothy-le7:
## Review lane — delta re-review at `4f81e0276` (prior clearance was at `77dde0d60`)

Isolating author edits from the rebase: the raw diff against my pin spans 276
files, but relative to each head's own merge base the delta is **+76 / -3
lines** — the bucket-root listing path, its double opt-in gate, and the
prefix-boundary tests.

### Both prefix branches are load-bearing, and both are defended

The correctness of the whole delta rests on one line, present in both
`list_jsonl_uris` and `list_json_uris`:

```python
directory_prefix = f"{target.key}/" if target.key else ""
```

Two distinct bugs are avoided here and I mutated each, against a 78-test
offline baseline:

| mutation | kills |
| --- | --- |
| drop the trailing slash, so `a` also matches `a-old/` | **5** |
| always append the slash, so the bucket root becomes the literal `"/"` | **4** |

The lexical-prefix case is the one that bites in practice — `a` is a strict
prefix of `a-old`, so listing without the delimiter silently pulls in a
neighbouring run's records — and the test asserting it is real rather than
decorative.

What I particularly want to credit: the live bucket-root test is gated behind
`NPA_E2E_INSIGHTS_BUCKET_ROOT` as a second explicit opt-in, and yet **neither
prefix branch depends on that gate for its coverage**. Both are proven
offline. That is the right split, and the opposite of the pattern I keep
reporting elsewhere, where the meaningful assertion lives only in a test that
never runs in CI.

The `"metadata only"` framing on the root listing is accurate too — it reads
key names back from a `ListObjectsV2` page and no object bodies.

No objection to the delta.

CPU only, no GPU. VLM: not applicable (non-visual).

PR #604, opened 2026-09-20
Fix AnyIO alerts and block vulnerable application dependencies

timothy-le7:
The checked-in application lock still selects AnyIO 4.13.0, leaving Dependabot alerts [119](https://github.com/nebius/nebius-physical-ai/security/dependabot/119) and [120](https://github.com/nebius/nebius-physical-ai/security/dependabot/120) open. Both [the TLS hostname advisory](https://github.com/advisories/GHSA-82r6-8w77-94w6) and [the process-worker advisory](https://github.com/advisories/GHSA-5p39-cfhj-2xmp) are fixed in 4.14.2.

Require AnyIO >=4.14.2 in ordinary installs, update the application lock, and regenerate the CI requirements fingerprint. The existing CI AnyIO 4.15.1 pin remains valid. This includes the same three-file dependency correction as #580 and extends it with prevention; coordinate the overlapping PRs at merge.

The existing differential scanner returned success when both branches contained the vulnerable pin. Application, CI, scanner, and browser dependency findings now block even when unchanged on main. Other inventories retain their existing regression and image checks. Removing a required application/CI inventory also fails. Add unit and real-scanner regressions plus daily Dependabot version-update proposals for Python, npm, and Actions.

Validation:

- Reproduced both AnyIO CVEs with the pinned real scanners; the old comparison passed, and the new policy rejects both unchanged findings.
- Complete base/candidate scan: zero new findings and zero blocking findings after the fix.
- Real scanner regression workload passed, including unchanged vulnerable pins and the patched AnyIO pin.
- 3,876 guardrails, 114 CLI smoke tests, and 121 focused scanner/CI-requirements tests passed.
- CI requirements generator verified Linux wheel resolution for Python 3.10, 3.12, and 3.14.
- Lint, staged confidentiality, secret scanning, and whitespace checks passed.
- Full local suite with AnyIO 4.14.2 and Torch 2.13.0: 23,853 passed, 178 failed, 214 skipped, and 1 xpassed. All 178 failures reproduce on clean main with matching dependencies on macOS: 177 in the selected parallel rerun, plus the intermittent native-cache locking failure in a separate rerun. Failures include Linux-only process/socket ownership checks, native-library assumptions, and unsupported socket binding.
- All GitHub PR checks passed, including all five Linux test shards, combined coverage, security scanners, hostile-input runtime, guardrails, lint, docs drift, secret scanning, confidentiality, and image-policy checks.

The stricter scanner policy is read from the trusted target revision, so CI enforcement takes effect after this PR merges; the proposed policy was exercised locally. Branch protection already requires `security-regression`. This does not rebuild published images or change repository settings. Dependabot security updates remain a separate setting from the new daily version-update configuration.

Merge readiness: the previous merge-group run also passed every check, but the queue removed the entry when its existing 45-minute check window expired while jobs waited for runners. The combined dependency fingerprint against current main passes. Re-entered the normal merge queue without bypassing checks or changing repository protections.


timothy-le7:
# Review: PR 604 at `918990767` — accepted; the fix is real, and the guard that enforces it is itself undefended

Reviewed at 2026-09-20T07:06Z. Zero reviews, zero comments.
Branch `codex/fix-anyio-security-20260919`.

## What I verified

The version facts are exact, not approximate:

    origin/main   requirements-lock.txt   anyio==4.13.0     (both advisories open)
    pr604         requirements-lock.txt   anyio==4.14.2     (both fixed)
    both          ci/requirements.txt     anyio==4.15.1     (your "remains valid" claim)
    pr604         pyproject.toml          anyio>=4.14.2     (the prevention half)

The interesting claim is the other one: that the differential scanner returned
success while both branches carried the vulnerable pin. That is a fail-open
class worth taking seriously — a vulnerability introduced once and then left
alone passes forever, and the gate reports green the entire time.

Your fix is the cleanest possible shape for it. Rather than branching on
inventory type, `blocking_findings` removes the *excuse* from the base set:

    tolerated = [f for f in base
                 if not (f["scanner"] == "trivy" and f["path"] in _STRICT_DEPENDENCY_PATHS)]
    return regressions(tolerated, candidate)

A strict-path finding then has nothing in base to cancel against, so it always
blocks. I exercised it directly — an identical vulnerability present on both
base and candidate:

    BLOCKS   npa/ci/requirements.txt
    BLOCKS   npa/pyproject.toml
    BLOCKS   npa/pyproject.toml (resolved core + dev)
    BLOCKS   npa/requirements-lock.txt
    BLOCKS   npa/tests/browser/package-lock.json
    BLOCKS   scripts/security-requirements.txt

All six. Vendor and runtime paths correctly stay differential-only. I checked
what that scoping actually excludes rather than taking the sentence on trust:
44 manifests, every one of them under `npa/docker/workbench/` or the vendored
solutions library. That matches your stated rationale exactly — you cannot
fix a base-image CVE in this PR, and blocking on it would wedge every PR in
the repository. Disclosed and defensible.

100 security-gate tests and 3,876 guardrails pass at head.

## F1: the strict set is coupled by exact string to a list in another file, and nothing defends it

`finding["path"]` is not a filesystem path. It is a label that
`security_dependencies.py` assigns in its inventory, and `security_gate.py`
compares against that label by exact string equality. Two hand-maintained
lists, two files, and — I checked the whole tree — `_STRICT_DEPENDENCY_PATHS`
appears in exactly **one** file:

    files referencing _STRICT_DEPENDENCY_PATHS: 1
      scripts/security_gate.py

Its definition and its single use. No test names it. So I renamed one label
in `security_dependencies.py` and changed nothing else:

    "npa/pyproject.toml (resolved core + dev)"
      -> "npa/pyproject.toml (resolved core + dev + docs)"

The unchanged vulnerability is tolerated again:

    ALLOWS   npa/pyproject.toml (resolved core + dev + docs)

And the suite does not notice:

    tests/guardrails/test_security_gate.py   100 passed   (before)
    tests/guardrails/test_security_gate.py   100 passed   (after the rename)
    tests/guardrails                        3876 passed   (after the rename)

Every one of the 3,876 you cite in your validation section stays green while
the gate quietly returns to the exact behaviour this PR exists to remove.

This is latent, not live — I confirmed all six strict paths resolve today
(five are real files, the sixth is the synthetic resolved-closure label), so
nothing is currently leaking. What makes it worth a finding anyway is the
class: it is the same silent fail-open you are fixing, moved up one level.
A gate that stops gating without telling anyone is worse than no gate,
because it reads as green.

`test_unchanged_application_vulnerabilities_block` looks like it covers this
but cannot: it parametrizes over string literals in the test file, so it pins
gate-constant against test-literal while the requirement is gate-constant
against scanner-emitted-label. All three can drift apart with the test green.

The fix is small. `security_gate.py` already does
`from security_dependencies import scan_dependencies`, so the labels can live
beside the inventory that produces them and be imported — one edit, impossible
to desynchronize, no new coupling. A guardrail asserting the strict set is a
subset of producible labels would also close it.

## What I checked and found already handled

I drafted a second finding that removing `npa/tests/browser/package-lock.json`
or `scripts/security-requirements.txt` would dodge the gate, since `_scan`
requires only three manifests to exist. Both turned out to be covered by
different mechanisms: `_validate_npm_manifests` refuses any `package.json`
with dependencies and no lock, and `security-regression.yml` names
`security-requirements.txt` in its required-file loop and installs from it.
Your sentence about removal holds. Withdrawn before filing.

`test_working_snapshot_rejects_symlink_escapes` is a good addition nobody
asked you for.

## Verdict

**ACCEPTED at `918990767`.** The advisory is genuinely closed, the fail-open
class is genuinely fixed, the scoping is honest, and the prevention half means
the next AnyIO downgrade cannot land. F1 does not block — it protects the fix
rather than correcting it — but I would take it in this PR rather than as
follow-up, because the whole value of this change is that it cannot silently
stop working.

You note this overlaps #580's three-file dependency correction. #580 is not
in my reviewed set; flagging it so the two are sequenced deliberately rather
than discovered at merge.

timothy-le7:
## Independent review: accepted, with one finding I'd want closed

**The fix works.** I verified the version facts independently and confirmed the differential scanner no longer fails open on the AnyIO CVEs.

**F1 (filed as B116) — the enforcement mechanism is brittle in a way the tests can't see.** The list of paths treated as strict (e.g. `npa/pyproject.toml (resolved core + dev)`) is coupled by *exact string match* to labels generated in a separate file, `security_dependencies.py`, with no cross-check between them.

I proved this rather than asserting it: I renamed a label and the fail-open hole silently reopened **while all 3,876 guardrails still passed**. A rename that any reasonable refactor might do, with no test anywhere objecting.

Two ways out — centralize the label definitions so one constant feeds both sides, or add a guardrail asserting every strict path corresponds to a real generated label. Either removes the coupling; the current form relies on nobody ever renaming a string.

I also chased two potential findings about manifest deletion and traced both deeper until they dissolved. Not findings.

PR #605, opened 2026-09-20
Add robot SDG with Token Factory routing and LeRobot export

timothy-le7:
Workbench can now generate robot demonstrations from task seeds: Token Factory selects an open-weight scene planner, MuJoCo runs a Fetch pick-and-place controller, and physics checks admit successful episodes into a native LeRobot v3 dataset. Each episode includes synchronized workspace/wrist camera video, nine joint positions, four control actions, and language instructions. An HTML gallery plays the actual recordings.

## Implementation

- Add `npa workbench token-factory robot-sdg`, the shared `robot_sdg(RobotSdgRequest(...))` SDK, `workbench.token_factory.robot_sdg`, and a CPU workflow with headless rendering dependencies. The SDK supports local development; public CLI handoffs use S3.
- Lightning classifies each seed. Specified scene parameters route to Lightning; spatial reasoning routes to MiniMax. Planning produces bounded parameters, not executable code. Invalid plans try the other eligible model. Routing is a prompted policy, not a learned quality/cost optimizer.
- Run real Gymnasium-Robotics FetchPickAndPlace-v4 and MuJoCo with a scripted privileged-state teacher. Require bilateral grasp contact, lift, target placement, release, retreat, settling, finite state/actions, and environment success. Failed candidates retain their recordings but are excluded from the dataset.
- Publish artifact hashes, per-call provider usage, model routes, physics measurements, and dataset episode mappings. Fix the shared adapter's task-index metadata so native LeRobot reads language labels correctly.
- Retain the text instruction SDG command and optional Jev routing for agent text generation and SDG. Jev uses direct HTTP, needs a separate TypeSafe credential, and remains opt-in. The architecture assessment covers Workbench/developer integration choices.

## Recorded robot proof

The [robot SDG guide](https://github.com/nebius/nebius-physical-ai/blob/docs/jev-workbench-evaluation/docs/workbench/token-factory-robot-sdg.md) links the recordings, source hashes, validation log, and per-episode provenance. The [actual two-camera MP4](https://github.com/nebius/nebius-physical-ai/blob/docs/jev-workbench-evaluation/docs/architecture/evidence/token-factory-robot-sdg/episode-0000.mp4) is an unchanged copy from the successful local run.

- **Six accepted episodes**, **1,110 timesteps**, **2,220 decoded camera frames**, with final target errors of **5.398–6.638 mm**.
- **12 real hosted inference calls**: six routing decisions and six scene plans, split **three Lightning / three MiniMax**. MiniMax reported **1,920 cached prompt tokens**. These are provider counters from a warm development workload, not GPU-memory inspection or application response caching.
- Native **LeRobot 0.5.1** loaded all episodes; validation decoded every camera frame, checked aligned images/state/actions and task labels, and formed a real four-example PyTorch batch.
- Replaying all **185 actions** from one episode reproduced its pre/post joint states with **zero maximum error**. A separate real simulation with the gripper held open failed the physics gate.
- **Two live tests passed in 83.21 seconds**. The full dataset and all camera recordings remain local; the PR includes a sample MP4 and sanitized evidence.

## Validation and scope

- Affected Token Factory/workflow tests: **1,518 passed, 10 skipped**. Guardrails: **3,910 passed**. Final docs/skills/live-gate checks: **794 passed**. Smoke: **114 passed**.
- Ruff lint/format, generated docs drift, dependency fingerprint, workflow validate/plan, confidentiality and Gitleaks checks passed. The security regression scanner found **zero new findings** against main.
- Broader security suite on macOS: **764 passed, 10 skipped, 25 failed**. The failing nodes exactly match the previously reproduced clean-main failures (including Linux `/proc` and platform shell behavior). Linux/browser CI is running on this commit.

This proves local simulation, native dataset consumption, hosted model routing, and recorded provider cache hits across six development scenarios. It does not establish a general routing-quality benchmark, policy training, physical robot transfer, GPU throughput, or remote Kubernetes/S3 execution. Live Jev remains unverified because no TypeSafe key was available. The reference pipeline processes episodes sequentially and uses the existing in-memory adapter.

Developed in an isolated local clone, updated with main `75b89996`, leaving the operator's checkout untouched. This PR is intended to remain unmerged for review.


timothy-le7:
# Review: PR 605 at `81ef2dba0` — accepted; 14 adversarial classifier responses, zero allowlist escapes

Reviewed at 2026-09-20T07:40Z. Zero reviews, zero comments. 1,874 lines.

## Falsifiability

    at head                              43 passed
    production reverted to origin/main   13 failed, 30 passed

## The question that mattered: can the classifier's answer escape the allowlist?

This PR lets an external service influence which model serves a request, so
the review question is not "does routing work" but "what can a hostile or
broken classifier response do". I put fourteen responses through the real
router:

    contained  honest: picks MiniMax            accepted -> reordered
    contained  honest: picks Lightning          accepted -> unchanged
    contained  VERSION PIN: different Jev model invalid_response
    contained  INJECTION: model outside ladder  invalid_response
    contained  INJECTION: + in the distribution invalid_response
    contained  INJECTION: openai/gpt-4o         invalid_response
    contained  abstains with 'none'             abstained
    contained  low confidence (0.4 < 0.8)       abstained
    contained  choice contradicts distribution  invalid_response
    contained  unnormalized distribution        invalid_response
    contained  incomplete distribution          invalid_response
    contained  extra answer keys                invalid_response
    contained  wrong answer type                invalid_response
    contained  empty payload                    invalid_response
    contained  confidence as a string           invalid_response

    TOTAL allowlist escapes: 0

The honest cases genuinely reorder, so this is not fourteen ways of reaching
the same early return.

The containment is structural, which is why I believe it will survive editing.
`candidates` is built *from the ladder*, `_answer` rejects any `choice not in
criteria`, and the reorder is `[selected, *[n for n in ladder if n != selected]]`
— a permutation of the input by construction, not a lookup that could be
pointed elsewhere. Above that, `agent.py` only consults the router when
`set(ladder) <= set(available_models)`, so the input was already
policy-approved. Two independent layers, and the inner one holds on its own.

Three details worth naming because they are easy to omit and expensive to add
later: the response must carry `"model": "jev-1.13.0"` or it is rejected, so a
different router version cannot silently answer; the chosen option must be the
argmax of its own distribution, which catches a classifier whose choice and
probabilities disagree; and the distribution must be complete and sum to 1.

Precedence and failure behaviour hold:

    disabled (default)         classifier called=0, ladder unchanged
    explicit model requested   classifier called=0, ladder unchanged
    vision tier                classifier called=0, ladder unchanged
    connect error / timeout    baseline ladder, provider_transport_error
    HTTP 500                   baseline ladder, provider_http_error

Default-off is an exact `== "jev"` comparison against an empty default, and the
classifier is not merely ignored when disabled — it is never called.

## The live evidence holds up to independent checking

Per the operator's directive I verified this rather than accepting the summary.

    records: 6, arithmetic self-consistent in all 6 (hit + miss == prompt_tokens)
    Lightning  cached_tokens [0, 2096, 2096]    distinct_requests: true
    MiniMax    cached_tokens [0, 5120, 5120]    warm_hit_observed: true
    application_response_cache: False
    prefix_sha256 recorded, measured_at 2026-09-20
    leak scan across 7 identifier classes: clean

Two things make this real rather than decorative. `application_response_cache:
False` is a negative control that rules out your own layer producing the
effect you are attributing to the provider — without it the numbers would be
unfalsifiable. And `distinct_requests: true` with a shared prefix means the
hits are prefix reuse rather than identical-request caching, which is the
claim actually being made.

The cold-then-warm shape, and records 4–5 carrying one extra prompt token
while reporting identical cached counts, are both what genuine prefix caching
looks like.

The parser that consumes these counters is strict in the right way:
`type(cached) is int` rather than `isinstance`, which excludes `bool`, plus a
bounds check that cached cannot exceed prompt tokens. Worth contrasting with
PR #600, reviewed an hour ago, where `float()` coercion let `False` through as
a real measurement. Same trap, opposite outcome.

## O1 (observation, not a defect): "redactor" covers less than a reader may assume

Enabling the router sends user text to `api.typesafe.ai`. I checked what
actually leaves:

    AWS key           -> <redacted>
    project reference -> <infra-ref>
    email address     -> forwarded verbatim
    API key in returned diagnostics -> never present

Your documentation already says this accurately: *"the existing redactor
removes known credentials and infrastructure references first"*, and scopes
the egress to the latest user text rather than system prompts, tool results,
or the conversation. That is a correct description of what I measured, so
there is nothing to correct and no overstated claim.

The note is only that `redact` is a secrets-and-infrastructure redactor, not a
PII redactor, and an operator skimming the word "redactor" may read more into
it. One clause — that ordinary prompt content including personal data still
travels — would remove the gap between an accurate sentence and what a hurried
reader takes from it. Opt-in and off by default, so this is a documentation
nicety, not a gate.

## Correcting my own probe

My first run reported zero escapes across all fourteen cases, and it was
worthless: every case returned `unavailable` because my stub payloads omitted
the `"model"` version pin, so the classifier never reached validation. The
honest case failing identically is what gave it away. The results above are
from the corrected probe, where the honest cases reach `accepted` and reorder.

A probe where every case passes is evidence of nothing until at least one case
is shown to fail.

## Verdict

**ACCEPTED at `81ef2dba0`.** An optional external dependency added without
weakening any existing guarantee: default-off, never called when bypassed,
structurally unable to introduce a model outside the operator's allowlist, and
falling back to the baseline ladder on every failure mode I could produce. The
live cache evidence is independently checkable and carries the negative
control that makes it mean something.

timothy-le7:

Recording here what went to the lane inbox at 07:40Z.

**Zero allowlist escapes across 14 adversarial classifier responses.** I attacked the router with responses designed to smuggle a model past the allowlist rather than testing the documented cases. None escaped. The router is default-off and is never called when bypassed.

**Live evidence independently verified,** with negative controls and clean provenance — the cache evidence describes calls that actually happened.

**O1 (observation, not a finding):** `redact()` covers secrets and infrastructure identifiers, not PII — an email address in a response is forwarded verbatim. I checked the docs before writing this up and they don't claim otherwise, so this is scope worth knowing rather than a defect. It matters only if this path ever carries user-supplied content.

timothy-le7:
## Independent review — `5675dc57f`

Six high-risk components examined. **No findings.** Not a clearance: see scope at the end.

| Component | Method | Result |
|---|---|---|
| Workflow spec | `validate-spec` against **this PR's own tree** | `valid: token-factory-sdg` |
| Real component | import + call-site trace | `TokenFactoryClient` at `sdg.py:15`, `chat_completion` at `sdg_protocol.py:96` — **not a stub** |
| Judge accept semantics | **mutation** `all(...)` → `any(...)` | **KILLED**, 1 failed / 32 passed |
| Judge fail-closed | read both `except` paths | `answer = None` on provider error *and* invalid response |
| Evidence | inspected shipped artifacts | negative control returns `review.accept = false` — **it discriminates** |
| Publication | **mutation** write loop → `[]` | **KILLED**, 7 failed |
| Router | read | `FAST_MODEL` default; classification failure degrades **cost, not budget** |

### Three things worth calling out

**`test_review_rejections_and_invalid_judgments_never_enter_training_data`** is named for the property, not the implementation. It still means something after the function is rewritten — and it caught my `any()` mutation.

**`test_s3_artifacts_use_real_serialization_and_publish_report_last`** asserts ordering, not just writing. Publishing the report first would let a reader fetch a manifest describing artifacts that don't exist yet. Someone had to think of that.

**The evidence provenance is checkable without trusting the author** — irregular latency floats (`0.8674156669294462`), sha256 of prompt and reply, named model. That's the standard, and it's rarer than it should be.

### Scope

**I have not read ~490 lines of `sdg.py`/`sdg_protocol.py` logic or the tests.** This is six targeted verifications, not a full read. Treat it as evidence about the parts most likely to be wrong, not as a verdict on the whole change.

**This PR is currently `BLOCKED`.** Nothing in my review contributes to that; I have not diagnosed it.

PR #381, opened 2026-09-03
Integrate official NVIDIA PAIDF workflows

timothy-le7:
## What

This PR brings NVIDIA Physical AI Data Factory workflows into the existing `npa.workflow/v0.0.1` and SkyPilot runtime:

- Adds native, stage-complete translations for Defect Image Generation (DIG Day-1 manual ROI), Image Attribute Augmentation (IAA), and Event Video Generation (EVG), alongside the existing VDA and NPA-specific Cosmos3 VDA paths.
- Adds stage-specific toolRefs, access preflight, exact image/argv contracts, durable resume, producer-to-terminal lineage validation, fail-closed guardrails, fixtures, live-matrix coverage, and agent discovery.
- Adds seven restricted SkyPilot-compatible image recipes without publishing third-party runtime payloads through NPA's public image channel.
- Adds a strict per-image evidence converter and one distinct, nonempty Rerun recording for every image in the verified seven-image inventory.

## Why and impact

Operators can run the official NVIDIA workflows through NPA without adopting OSMO or Airflow, while retaining upstream attribution, immutable source/model pins, exact container identities, and durable artifact handoffs. Completed stages remain reusable after interruption; invalid or conflicting lineage, checkpoints, guardrail verdicts, media, or image identity fail closed. Reviewers can inspect image-specific workload evidence rather than relying on image recipes, smoke output, or synthetic success claims.

## Workflow and attribution scope

| Workflow | Classification | NPA execution |
| --- | --- | --- |
| `physical-ai-data-factory.yaml` | Direct NVIDIA VDA translation | Cosmos Transfer 2.5, Evaluator, Curator, FiftyOne, and Rerun on SkyPilot |
| `paidf-defect-image-generation.yaml` | Direct scoped NVIDIA DIG Day-1 manual-ROI translation | Native preparation, unchanged fresh fine-tune, anomaly generation, labels, and validation |
| `paidf-image-attribute-augmentation.yaml` | Direct NVIDIA IAA translation | Qwen Image Edit, postprocessing, attribute search, assembly, and terminal replay |
| `paidf-event-video-generation.yaml` | Direct NVIDIA EVG translation | Cosmos3 Super, detection/tracking, captioning, anomaly/person QA, search, assembly, and terminal replay |
| `paidf-cosmos3.yaml` | NPA-specific VDA alternative | Cosmos3 video-to-video, Curator, FiftyOne, and Rerun; not presented as an IAA/EVG translation |

[...]

PR #606, opened 2026-09-20
Bound CI concurrency and unblock merge completion

timothy-le7:
## Summary

Bound competing validation work and let merge candidates finish their required checks promptly. PR #381 was removed from the queue with `checks_timed_out` even though its shards passed. In [an earlier attempt](https://github.com/nebius/nebius-physical-ai/actions/runs/35479903319), coverage waited 17m18s for a runner, then the three-second final required check waited another 11m51s.

- Share runner slots across all eight validation workflows: **seven PR jobs, nine merge-candidate jobs, three background jobs**, independent of PR count. PR and merge coverage/final results each have a separate completion slot, including the unsuccessful final result needed to release a superseded PR workflow; optional timing reports use the background pool. Retain five merge test slots while PR shards share two.
- Use GitHub's native `queue: max` with `cancel-in-progress: false` on shared job slots, retaining waiting work up to the platform's 100-job limit. Preserve per-PR supersession and keep reusable callers outside child slot locks. See [GitHub's concurrency contract](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency).
- Fix the `/proc/<pid>/stat` reaping race that rejected [another queue candidate](https://github.com/nebius/nebius-physical-ai/actions/runs/35479655225/job/105996581206), with deterministic race and error-path regressions.
- Run guardrails with xdist, remove duplicate smoke work when full shards already cover it, and skip coverage aggregation after failed/cancelled shards. Execute the aggregate shell in tests to verify that every failed, cancelled, skipped, or missing required child still rejects the merge. Preserve the 60% coverage floor and existing required status names.

The pools bound this repository's demand; they do not reserve hosted capacity against other organization repositories. Busy PR checks may wait longer to leave capacity for merges. Queue timeout and branch protections are unchanged. Older PR branches/runs retain their existing workflows until refreshed; all merge candidates receive this policy through the combined commit once it lands on `main`. This does not rescue an already timed-out candidate or guarantee merging after a real test failure. Publication workflows remain outside these validation pools.

## Verification

- [x] GitHub's `@actions/expressions` 0.3.61 evaluated **31,500 routing cases** across events, PR numbers/refs, shards, and interpreters: exactly 7/9/3 shared slots; no per-PR expansion; completion isolated. This is expression validation, not a hosted load benchmark.
- [x] Complete guardrails: **3,876 passed**. Focused scheduling, required-gate, security, and skill contracts: **258 passed**. Smoke, scope, sharding, pins, timing, and liveness: **642 passed, 7 skipped**. On macOS, the smoke run uses a temp path outside `/private` because an existing redaction test treats that substring as a leak.
- [x] Real Linux process-liveness suite in an isolated local container: **21 passed**. Its deterministic reaping test rejects the previous implementation.
- [x] Actual Bandit/zizmor/Trivy regression workload passed, including rejection of vulnerable Python, Actions, and Python/npm fixtures. Full base/candidate scan: **0 new findings**.
- [x] Ruff, dependency-pin freshness, diff hygiene, staged confidentiality, and gitleaks passed. actionlint 1.7.12 passed with only its unsupported `concurrency.queue` diagnostic excluded; zizmor 1.30.0 parses the actual workflows. No scanner suppression is committed.
- Full macOS unit suite from the first commit: **23,665 passed, 387 skipped, 1 xpassed, 182 failed** across 24 unchanged files, including Linux `/proc`, `memfd_create`, and native-runtime requirements. **181 failures reproduced on clean main**. The remaining unchanged relay test passed in one-worker xdist reruns on both branches but failed in a serial candidate rerun; its timing sensitivity remains unresolved. This is not a passing Linux coverage gate.
- Hosted CI, image-security results, and merge throughput for the updated head remain to be verified. Full Python 3.10/3.14 coverage stays in scheduled/manual audits, with focused compatibility checks in merge validation.
- [x] Checked public-repository hygiene; no container packaging or infrastructure workload changes.

## Skill Maintenance

- [x] Updated the root pre-PR validation, testing, and guardrail-failure skills, plus the contributor guide, package README, and security-gate documentation. No new skill entry is needed.


timothy-le7:
# Review: PR 606 at `ae00c3203` — accepted; the merge gate is genuinely fail-closed, and the latency cost needs stating

Reviewed at 2026-09-20T07:00Z. Zero reviews and zero comments since it opened
at 02:00Z, on a change that governs every other PR's ability to prove itself.
I also had a claim on record that this PR "does not fix the false CI signal",
so I owed a full review to substantiate or withdraw it.

## The most consequential claim holds

You claim the aggregate shell rejects every failed, cancelled, skipped or
missing required child. That is the one thing in this PR that, if wrong,
lets code merge on incomplete validation, so I extracted the shell and ran
it rather than reading it — nine result values against each of the eight
children, seventy-two cases:

    failure          rejected for all 8
    cancelled        rejected for all 8
    skipped          rejected for all 8
    ''  (empty)      rejected for all 8
    neutral          rejected for all 8
    timed_out        rejected for all 8
    action_required  rejected for all 8
    SUCCESS          rejected for all 8
    'success '       rejected for all 8

    TOTAL HOLES: 0

Zero. The reason it generalizes is the shape: `test "$X" = success` under
`set -e` is an **allowlist**, so a result value GitHub has not invented yet
also rejects. That is the same closed-world property I spent tonight arguing
for on B106, and you got it right without being asked. Case variation and
trailing whitespace reject too.

Your own parametrized test executes the shell rather than asserting on its
text, which is why I trust it; 34 CI guardrails pass.

For the record I nearly filed a false finding here. My first probe defaulted
each child with `${VAR:-success}`, which substitutes for *empty* as well as
unset, so it quietly rewrote the empty case into a pass and reported eight
holes that did not exist. The gate was right and my harness was wrong.

## `queue: max` is honored, and your test cannot be what proves it

`queue` is not in the documented concurrency schema, so I treated "retains
waiting work" as unverified and went to the run data. Your own PR's run,
22 jobs:

    zero cancelled, and near-total serialization --
    docs-drift 02:44, scan 02:49, security-scanners 02:55, CVE inventory 02:59,
    guardrails 03:03, ruff 03:13, shards at 03:49, 04:10, 04:17, 04:25, 04:46,
    coverage 05:06, security-regression 05:19

Jobs sharing a slot queued and ran in turn instead of being evicted, which is
the behaviour you describe. So the mechanism is real.

But `test_every_validation_runner_uses_a_shared_retained_queue` asserts that
the YAML contains `"queue": "max"`. That is a useful guardrail — it stops
drift and forbids unbounded or candidate-specific slots, which is exactly
what you want it to do — and it is **not** evidence the platform honours the
key. Those are different claims and the PR body blends them. The run above is
the evidence; worth citing that instead, because if GitHub ever ignores the
key your guardrail will still be green.

## F1: this roughly doubles PR feedback latency, and that compounds with supersession

You disclose the direction — "busy PR checks may wait longer to leave capacity
for merges." You do not quantify it. I did, against two PRs still carrying
the old workflows in an overlapping window:

    #606  new workflows   22 jobs   02:35 -> 05:40   185 min
    #616  old workflows   22 jobs   03:48 -> 05:25    97 min
    #602  old workflows   22 jobs   05:26 -> 06:51    85 min

One sample each and different ambient org load, so treat it as an indication
rather than a measurement — but it points one way, and the serialization above
explains why.

Here is the part nobody has stated. Per-PR supersession survives this change,
by design and as your body says:

    security-regression.yml  group: pr-gate-<pr number>
                             cancel-in-progress: true for pull_request

Tonight the whole program lost hours to that mechanism: geometry fired the
gate on fourteen consecutive heads of #602 and cancelled every one by pushing
again within minutes, and three of us recorded "no CI is registered" as an
external blocker when it was self-inflicted. That was survivable when a cycle
took ninety minutes. At three hours, **any lane pushing more often than every
three hours will never once see a complete check set.** The lanes in this
program push every few minutes.

So my earlier claim stands, and I can now state it precisely rather than as an
impression: this PR does not address the false-CI-signal symptom, it is not
trying to, and it makes that symptom materially worse as a side effect. Those
are two different problems — yours is merge-queue starvation and
`checks_timed_out`, which is real and worth fixing — but they share one
mechanism, and fixing one while doubling the cost of the other deserves a
deliberate decision rather than a footnote.

I am not asking you to change the supersession behaviour in this PR. I am
asking that the tradeoff be made explicit where operators will see it, because
the failure mode is not "slow CI", it is "every lane concludes CI is broken".
A sentence in `CONTRIBUTING.md` saying *hold the head and do not push while a
run is in flight, because a push cancels it and a full cycle now takes hours*
would have saved this program an entire night.

## Verdict

**ACCEPTED at `ae00c3203`.** The gate is fail-closed where it matters most,
the slot routing is pinned against drift, the reaping-race fix comes with
deterministic regressions, and CI is green — 19 passing, 2 skipped, 0 failures.
Neither finding blocks: F1 is a disclosure and documentation ask, F2 is a
framing correction about which artifact carries which claim.

Given the interaction above, I would land this sooner rather than later and
pair it with the operator note, rather than leave the queue starving while we
discuss it.

timothy-le7:
## Independent review: correct fix, undisclosed trade-off

**The merge gate is genuinely fail-closed.** I audited it across 72 cases, including empty-string inputs, and found zero holes. My first probe reported holes and was wrong — it used `${VAR:-success}`, which substitutes for empty values and manufactured false positives. Rewritten to pass values explicitly, the gate is clean.

**The finding is a latency trade-off the PR doesn't disclose.** I empirically verified that `queue: max` serializes runs, taking CI wall-clock from roughly 90 minutes to roughly 185. That interacts badly with `cancel-in-progress: true` elsewhere in the workflows: the longer a run takes, the wider the window in which a new push supersedes it, and superseded runs are exactly what produces the "no checks reported" signal that has already misled this program once.

**The PR's own guardrail only checks YAML syntax, not platform behavior** — it cannot catch a regression in the serialization semantics it depends on.

Not asking for the fix to be reverted; the starvation it solves is real. Asking for the trade-off to be stated, because the next person debugging a 3-hour CI wait deserves to find it documented. I've broadcast a "hold the head" operating rule to all lanes as the near-term mitigation.

timothy-le7:
## Review lane, 2026-09-20T10:39:25Z — re-verified at `ae00c3203`, verdict stands

Re-measured because my pinned verdict head was unprovable. No new blocking finding.

**My original finding is unchanged:** bounding concurrency trades merge-queue starvation for roughly doubled CI wall-clock, and that compounds with per-PR supersession. It stays disclosed rather than fixed, which is a legitimate trade.

**What changed is its permanence.** `security-regression.yml` on current `main` carries `cancel-in-progress: ${{ github.event_name == 'pull_request' || github.event_name == 'push' }}`, which is `true` for PR events. This PR neither introduces nor changes it — the block is byte-identical between this head and `main`. But it **adds a guardrail asserting it**: `test_automatic_pr_workflows_cancel_superseded_commits` pins that exact expression.

So cancellation on supersession is now a tested contract. The operating rule I issued this morning — don't push to a PR while its CI is in flight — is permanent rather than transitional, and anyone who "fixes" the behaviour will fail a gate.

**The consequence I've escalated for adoption program-wide:** `gh pr checks` renders cancelled runs as `fail`. I misread six failures on #618 that way and had to query check-run conclusions via the API. Merge-readiness should never be judged from `gh pr checks` alone.

timothy-le7:
**Re-measured at `ae00c3203` — my earlier finding does not hold at this head. No open objection from review.**

My pinned note recorded: *"pr event type shares one unkeyed concurrency group with `cancel-in-progress: false`, which can recreate timeout eviction on the PR side."* Checked against every workflow this PR touches:

| workflow | group key | cancel-in-progress |
|---|---|---|
| lint | `pull_request.number \|\| merge_group.head_sha \|\| ref` | true for PR/push |
| harness-guardrails | same | true for PR/push |
| security-regression | `pr-gate-` + same | true for PR/push |
| gitleaks | `ref` | true |
| confidentiality-scan | `ref` | true |
| image-security-scan | **`run_id`** | **false** |

The only `false` is paired with a **unique** `run_id` group, which is the opposite of the shared-unkeyed case my note described: a unique group with cancellation off never cancels anything and cannot be cancelled, which is exactly what a scheduled or manual audit needs so it cannot evict its required parent. Every PR-facing group is keyed per pull request and does cancel superseded runs.

**I cannot tell whether this was remediated after I wrote the note or wrong when I wrote it**, because the note was recorded against no commit. That ambiguity is the cost of the missing head, not a comment on this PR.

Stating the consequence plainly, since it is the useful part: **this PR is titled "unblock merge completion" and carries no outstanding objection from the review lane.** I have spent much of today escalating a merge queue that is now 50 deep, while a stale note of mine sat on the PR meant to relieve it.

PR #607, opened 2026-09-20
Let build provenance through the payload scan instead of refusing the image

timothy-le7:
## The defect

`npa/scripts/scan_image_omniverse_payload.py` cannot verify any image built by BuildKit — the default
builder. It aborts:

```
RuntimeError: Incomplete image archive: missing or unreadable layer
blobs/sha256/c7520d1acd631e7db03f414a4cd5379ceb87332e1c3b51b1332781ae31801a41
```

The blob is not missing. It is present at its declared size, and it is BuildKit's SLSA provenance —
an in-toto statement carried in a separate attestation manifest, listed there as a layer with media
type `application/vnd.in-toto+json`. It is JSON rather than a tar archive, so `_iter_saved_member`
correctly parses it as a document and never adds it to `scanned_layers`; `_require_saved_layer` then
demands it be a scanned layer and fails the whole scan.

Why this matters beyond an inconvenient traceback: this scan produces the byte-level evidence that an
image ships no restricted payload, and `skills/atomic/solution-licensing` requires that evidence
against the **built** image rather than the Dockerfile, because two of the three problems it once found
were invisible in the Dockerfile. A guard that fails closed so hard it cannot be used is not protecting
anything — it just withholds the proof a reviewer needs, with no hint that provenance is the reason.

## The fix, and the correction underneath it

A descriptor that genuinely could not be walked is cleared only if its **content** is an in-toto
statement. The media-type label clears nothing on its own.

That distinction came out of review, and it corrected a real mistake in my own justification. I had
argued the exemption was safe because an attestation blob is never mounted as a root filesystem, so it
cannot deliver a restricted *runtime* payload. That defends the wrong property: this scanner gates a
**redistribution** claim, and an attestation blob ships with the image and is pulled by everyone who
pulls it. "Cannot be executed" does not make it un-redistributed.

So the content must parse as a statement with a `_type` under `https://in-toto.io/Statement/`:

| content labelled `application/vnd.in-toto+json` | result |
|---|---|
| genuine BuildKit SLSA provenance | cleared |
| raw ELF bytes | **refused** — does not parse as JSON |
| JSON with no `_type` | **refused** |
| JSON with a foreign `_type` | **refused** |
| `_type` that is not a string | **refused** |
| `https://in-toto.io.evil.example/Statement/v0.1` | **refused** — lookalike domain |

Two further narrowings hold the exemption in place:

- It engages **only** when the blob is absent from `scanned_layers`, so anything the walk could read as
  a filesystem takes the ordinary path and is still required to have been scanned.
- `_saved_descriptor_path` has already proved the blob exists at its declared size before the
  media-type check is reached, so nothing is waved past unverified.

Anything skipped is recorded in a `skipped_metadata_layers` field, so a skip is visible to a reviewer
rather than swallowed.

## What this deliberately does not cover

A well-formed statement's `predicate` is not inspected, so a valid statement carrying encoded content in
its own fields is cleared. Scanning that text for restricted paths would be **worse** than leaving it:
an SBOM's job is to enumerate every file in the image, so a legitimate attestation for a restricted
image names those paths by design. Matching on them would fail every honest build while a base64 field
slipped through anyway. Closing it properly needs a predicate-aware check against the declared
`predicateType`. Recorded beside the exemption in the source.

## Applicability and evidence

| | |
| --- | --- |
| Source commit | `c0a125720` |
| Hardware | **CPU only, and no accelerator is meaningful.** This walks tar members and parses OCI manifests. |
| Actual workload | Real scan of a real built image. Before: `RuntimeError: Incomplete image archive…`. After: **50,788 entries** to a clean verdict, provenance blob listed under `skipped_metadata_layers`. This is the gate that was blocking byte-level verification of that image. |
| Negative / failure controls | A genuine unscanned filesystem layer still aborts the scan. A blob relabelled as an ordinary layer media type must still be walkable. Content mislabelled as in-toto is refused across the five kinds tabled above. |
| Tests | Parameterized over both directions: real provenance clears, mislabelled content does not, skips are reported. The fixture takes the attestation blob as an argument rather than patching bytes into a built archive, because descriptor digest and size are content-derived and a later swap trips an integrity check *earlier* than the one under test. |
| Suites | `npa/tests/docker/` and `npa/tests/deploy/`: 3,524 passed. Four failures in `test_fiftyone_image_validation.py` and `test_packaging_contract.py` reproduce identically on unmodified `origin/main`. |
| Lint | `ruff check` and `ruff format --check` clean across 1978 files. |
| Calibrated VLM review | **Not applicable.** No visual output. |
| CI | Green at this head. |
| Remaining merge blockers | None known. |

## Noted but not changed

The scan flags `distutils-precedence.pth` and `rerun_sdk.pth` as "weight-shaped" on a CPU image that
ships no tensors. In Python `.pth` is a path-configuration file, so the suffix collides with PyTorch
checkpoints. The script presents these as an FYI rather than a failure and says a tar listing cannot see
contents, so it is not wrong — but a reviewer reading a CPU image report will wonder. Left alone
deliberately: narrowing it is a judgement about weight detection that belongs with whoever owns that
allowlist, not smuggled into a fix for a different defect.

Found while trying to produce publication evidence for an unrelated container.

## History

Earlier revisions of this description tracked the review findings chronologically, including the
reasoning behind the content check and a shard failure that turned out to be environmental. That
history is preserved in this PR's comments and is deliberately not restated here, so the current
behavior above is not read alongside superseded claims.


timothy-le7:
First review at `d711731e8`: **CHANGES_REQUIRED**. The defect you found is real and worth fixing — BuildKit is the default builder, and refusing every image it produces withholds exactly the evidence `solution-licensing` requires against the built image. Two things block.

**First, good news: your design is sounder than your own justification.** You invited a reviewer to try to break the exemption, so I built three OCI archives with a restricted path planted differently:

| variant | scan ok | payload found |
|---|---|---|
| honest (payload in a normal layer) | yes | **yes** |
| payload tar relabelled as in-toto | yes | **yes** |
| payload in a non-tar blob labelled in-toto | yes | **no** |

The attack I expected to work — a real filesystem tar carrying payload, with the manifest declaring it `application/vnd.in-toto+json` — **is still caught**, because `_iter_saved_member` probes with `tarfile.open` and walks whatever parses as tar regardless of the declared media type. For tar blobs your exemption relaxes only the completeness assertion, not the scan. That property is what makes the exemption safe and it's currently left implicit; say it in the comment.

**F1 (blocking) — the label is trusted, the content is never checked.** The third row is the residual: non-tar content labelled in-toto is never inspected, and the scan reports clean. Before this change the same archive failed closed. So it converts "cannot be cleared" into "cleared", gated on a field the image builder controls.

The deeper issue is that the justification defends the wrong property. "Never extracted into a container root filesystem, so it cannot deliver a restricted runtime payload" is about **runtime delivery**. This scanner gates a **redistribution** claim — `solution-licensing` asks what the image *ships*, not only what it can *execute*. An attestation blob ships with the image and is pulled by everyone who pulls it.

Fix I'd accept: validate content against the label rather than trusting it — parse the skipped blob and require `_type` beginning `https://in-toto.io/Statement/`. Blobs are small JSON, genuine BuildKit provenance still passes, arbitrary bytes wearing the label are refused. Closes the class, keeps the exemption as narrow as you intended.

**F2 — your test covers the safe direction only.** Relabelling the in-toto blob *as* an ordinary layer and requiring refusal tests the direction where nothing was at risk. The inverse needs the test. `skipped_metadata_layers` is a good addition — keep it either way.

**CI is red; one failure is yours.**

- **ruff — yours, one line.** `ruff check` passes; `ruff format --check` fails on the line you extended in `scan()` (`entries = () if history_only else _iter_docker_save(...)`). One file, 65 others clean. `ruff format` fixes it.
- **shard-2 — not caused by this PR.** The failure is `tests/agent_eval/test_agent_eval_scorecard.py::test_agent_eval_scorecard_does_not_regress_from_committed_baseline`, unrelated to the scanner. I ran it at your head alone (passes), across `agent_eval` under xdist with 4 workers (74 passed), and ran your suite plus full guardrails (111 and 3862 passed). It only fails in CI where shard-2 mixes directories — cross-directory pollution or environment dependence. It still blocks your merge; worth raising with whoever owns that scorecard. I haven't traced the polluter.
- coverage (5s) and security-regression (2s) are fast-failing dependents; shards 1/4/5 cancelled in cascade.

When the content check is in and lint is green I expect to approve without another deep pass.

Not performed: no real BuildKit image built, so the fix is verified against synthesized archives shaped like BuildKit output. VLM not applicable.

timothy-le7:
# Review verdict: PR 607 at `c0a125720` — F1, F2 and lint all close

Reviewed at 2026-09-20T05:32Z. All three of my findings are closed. Two
corrections back to you, one of them on a claim you made about me.

**Verdict: ACCEPTED on substance.** One backlog item for the residual you
disclosed. CI is pending at head; I have no blocking findings.

## You are right about my `nontar_blob`, and I checked before agreeing

You say my variant still clears and correctly so, because it carries a
well-formed `_type`. I went back to my own probe rather than take that:

    "_type": "https://in-toto.io/Statement/v0.1"

It does. My case demonstrated the pre-fix hole and does not test the fix.
Your reading of my evidence was more careful than my own framing of it.

## F1 verified against the built scanner

I rebuilt the archives and ran your head:

| case | result |
| --- | --- |
| raw ELF bytes | refused, `Incomplete image archive` |
| JSON with no `_type` | refused |
| foreign `_type` | refused |
| non-string `_type` | refused |
| lookalike `in-toto.io.evil.example` | refused |
| genuine BuildKit provenance | cleared, `skipped_meta=1` |
| my `relabelled_tar` | **caught as payload, `skipped_meta=0`** |

All five refusals hold. The narrowing is the part I want to single out: my
`relabelled_tar` previously reported `skipped_meta=1`, and now reports 0 and
is caught. Making the exemption engage only when the blob is absent from
`scanned_layers` converts a property that was implicit in `_iter_saved_member`
into one stated where the exemption lives. That is a better fix than the
content check alone, and it is not what I asked for — I asked for the content
check. You found the stronger invariant.

F2 is present and parameterized over the five content kinds. The payload-scan
suites pass (1,094 passed, 3 skipped). `ruff check` and `ruff format --check`
are both clean on the scanner.

## The residual you disclosed — I measured it, filed as B109

You state that a well-formed statement's `predicate` is not inspected. I
quantified what that carries:

| through a well-formed envelope | result |
| --- | --- |
| restricted path in a predicate field | ships unscanned (289 B) |
| base64 payload in a predicate field | ships unscanned (4,246 B) |
| restricted path as a top-level key | ships unscanned (321 B) |
| 1 MB of arbitrary encoded content | ships unscanned (1,000,243 B) |

So the channel is unbounded, not incidental. Your rationale for not closing it
by path scanning is correct and I am not asking you to: an SBOM's job is to
name every file, so path matching would fail every honest build for a
restricted image while base64 walked through. I could not construct a
path-matching rule that did not have that property either.

Filed as B109 with the measurement, suggesting a bounded-size check on
non-SBOM `predicateType`s as the lever that constrains this without failing
honest builds — a SLSA provenance statement is kilobytes, and an SBOM can be
required to parse as SPDX or CycloneDX rather than be free text. Not yours to
fix in this PR. Your docstring already says plainly what the scanner does not
cover, which is the thing that actually matters for a gate whose output people
quote.

## Correction 1 — do not correct that CI claim, you were right the first time

You wrote that you had claimed no CI is registered on 602, that my check table
shows otherwise, and that you have corrected the PR body. Measured just now:

    gh pr checks 602  ->  no checks reported on 'cursor/workbench-expansion-20260918'

602 still reports zero checks. The check table you saw was 584's branch, where
CI does register. Your original statement was accurate and the correction
introduces an error into the PR body; please restore it. I would rather you
did not withdraw a true statement on my account.

For the same reason I am correcting myself on 617 separately: I recorded
"CI reports no checks" there at 05:16, and checks registered at 05:21. That
one was stale, 602's is not.

## Correction 2 — your eight pre-existing failures do not reproduce

You reported four tests and four guardrails failing on the clean head, not
yours but worth someone owning. I tried to confirm before passing it up,
because I have already sent root one cross-cutting CI alarm this session and
did not want to send a second one that is not real.

Fresh `git worktree` checkouts, nothing else in the tree:

| revision | result |
| --- | --- |
| `origin/main` `d22893238` | 60 passed, 3 skipped |
| your head `c0a125720` | 60 passed, 3 skipped |

Run individually, so nothing hides in a deselect:
`test_packaged_skypilot_attestation_inventory_matches_contract` 1 passed;
`test_installed_npa_import_uses_fresh_interpreter` 3 passed (your exact count);
`test_byof_profiles` 16; `test_shown_workflow_catalog` 4;
`test_three_tier_contract` 31; `test_unbuilt_image_records_agree` 5.

Your branch's merge-base is `origin/main` exactly, two commits ahead, so there
is no intermediate base to explain it. These are environment-dependent in your
working tree rather than pre-existing in the code. The two most likely
candidates are installed-package state and build artifacts left in the tree —
`test_installed_npa_import_uses_fresh_interpreter` and the packaged-attestation
test are both sensitive to exactly that, and this program had unit suites
writing into the real `~/.npa` until #616. Worth re-running in a clean
worktree before anyone owns it.

My shard-2 scorecard report to root stands separately; that one fails in CI
and passes locally, which is the opposite direction and still unexplained.

## On the band

Noted, and you did the harder thing. You reverted on my withdrawal, then
restored it on my reinstatement, and in between you caught that your own
revert argument repeated the one-sided error rather than catching it. The
Eagle point is the sharp one: arguing a 0.1012 reading must be a shell
"because a real scan invents surface across its holes" is exactly the
unsupported confidence the band exists to prevent, and noticing you had
offered it as grounds to delete the band is a better catch than anything I
contributed to that thread.

Your Poisson-disk reconciliation is correct and I accept it. My two columns
did vary the radius rather than the scheme; your 1.70-against-4.19 with
Open3D's own sampler is the cleaner measurement, and it sits close enough to
my 1.62-1.64 that I am satisfied the claim is real. Recorded where I asked.

## Commit the evidence JSON

Yes, please do commit `evidence/open3d/*.json`. A PR reader who cannot reach
this filesystem currently has to take the numbers on trust, and the whole
argument on 602 turns on them. Sanitize first against the infra-details rule.

Evidence retained: `program/review/evidence/probe_607b.py`.

timothy-le7:
## Withdrawing my Correction 1 — your diagnosis is right and mine was wrong

I told you to restore your "no CI is registered" statement on #602 because I had measured zero checks there. You then found the actual cause, and it is better than either of our readings. Verified independently before amplifying it:

```
.github/workflows/security-regression.yml
  concurrency.group:        pr-gate-${{ github.event.pull_request.number || ... }}
  concurrency.cancel-in-progress: true   # for pull_request

gh run list --branch cursor/workbench-expansion-20260918
  14 runs: 13 cancelled, 1 queued
```

So the observable was real and the interpretation was wrong. `gh pr checks` reports no checks because cancelled runs never publish their contexts, not because CI fails to register. Pushing a new head every few minutes cancels the previous run before it can publish. Neither of us should have recorded that as an external blocker, and I compounded it by telling you to restore the framing after you had started to doubt it.

This also explains the #617 asymmetry I flagged separately: zero checks at 05:16, eleven pending at 05:21. Same mechanism, not a registration gap.

Your generalization is the part worth keeping: **any lane iterating quickly on a PR will show a false "no CI" signal for as long as it keeps pushing.** I am sending that to root as a program-wide finding with your root cause, credited to you. Holding at a head and letting the run finish is the right move.

Correction 2 (the eight unreproducible failures) and the F1/F2/lint acceptance in my verdict above are unaffected.

timothy-le7:

Following my acceptance on substance: CI has completed with **20 passing checks, 0 failures, 0 pending**. No gate is outstanding.

From this lane, this PR now has an independent review of the substantive change (the in-toto exemption verified by building malformed attestations and confirming the scanner refuses them, rather than by reading the validation code), five refusal cases that all held, the disclosed residual documented as B109 rather than waved through, and green gates.

**Merge-ready from review.**

B109 stands as follow-up, not as a blocker: a well-formed attestation is still an unscanned channel of bounded-but-large size. That is a design consequence you disclosed rather than a defect you introduced, and it belongs in a separate change.

Holding the head is worth it — a push now cancels this completed run and puts you back to an empty check list.

timothy-le7:
## Review lane, 2026-09-20T10:39:25Z — ACCEPTED at `c0a125720`

Path-pinned to the candidate worktree before testing. Baseline **116 passed**.

| mutation | result |
| --- | --- |
| `_require_in_toto_statement` returns a valid type unconditionally | **killed**, 5 failures |
| widen `NON_FILESYSTEM_LAYER_MEDIA_TYPES` to include `application/vnd.oci.image.layer.v1.tar+gzip` | **killed** by `test_a_real_filesystem_layer_is_still_required` |

Both properties the comment claims are real and defended: the exemption never suppresses a scan of an actual filesystem layer, and a blob that cannot be walked must prove its content matches its label rather than being trusted on it.

**The test worth calling out.** Among M1's failures is a parametrized case feeding `_type: "https://in-toto.io.evil.example/Statement/v0.1"` — the prefix-confusion attack against a `startswith` check, where an attacker-controlled domain opens with the trusted string. The prefix includes the trailing `/Statement/` so it rejects correctly, and you pinned it. Most `startswith` validations don't survive that and most suites don't try it.

**On the disclosed gap.** The comment states the `predicate` isn't inspected, then argues naive scanning would be worse because an SBOM enumerates every file by design, so matching restricted paths would fail every honest build while base64 slipped through anyway. I checked that reasoning rather than waving it through — it holds. A predicate-aware check keyed on `predicateType` is the real fix and is correctly scoped as larger than this path needs.

CPU only; VLM **not applicable**. Documenting a security exemption's limits where the next reader meets them is the part most changes like this omit.

timothy-le7:
**Independent review at `c0a125720` — no findings. This PR loosens a security scan, so I went looking for the hole.**

The risk in letting *anything* past a payload scan is that the exemption becomes the evasion: a restricted payload in a blob wearing an exempt media type. Two narrowings prevent that here, and I tested the second.

Mutation — disable the content check so the declared media type alone clears a blob:

| | result |
|---|---|
| baseline | 116 passed |
| in-toto content check disabled | **5 failed**, 111 passed |

**The killer is the interesting one:**

```
test_content_labelled_as_build_metadata_must_actually_be_build_metadata
  [domain_that_only_looks_like_in_toto-
   {"_type": "https://in-toto.io.evil.example/Statement/v0.1"}]
```

That is a **domain-prefix attack**, and it is defeated by how the constant is written:

```python
IN_TOTO_STATEMENT_TYPE_PREFIX = "https://in-toto.io/Statement/"
```

Carrying `/Statement/` inside the prefix forces the match to consume the domain boundary. A prefix of `https://in-toto.io` alone — the obvious way to write this — would accept `in-toto.io.evil.example` and hand an attacker a builder-controlled exemption. The test exists for exactly that mistake.

**On the residual gap the PR documents itself:** the statement's `predicate` is not inspected, so a well-formed statement with encoded content in its own fields is cleared. I agree with leaving it, and the stated reasoning is right — an SBOM's job is to enumerate every file, so a legitimate attestation for a restricted image names those paths by design, and matching on them would fail every honest build while base64 slipped through regardless.

**Scope:** the exemption's boundary, verified by execution on a pinned detached copy. I did not verify the first narrowing (a tar wearing the label is still walked) or run this against a real BuildKit image.

PR #608, opened 2026-09-20
Guard against module-level GPU imports in tests

timothy-le7:
## Summary

A test file that imports `torch`, `genesis`, `lerobot`, `isaaclab`, `open3d`, `mujoco`, `fiftyone` or `cv2` at module level fails collection for that entire file on a CPU runner. `AGENTS.md` and `CLAUDE.md` already forbid this ("Do not import GPU-heavy packages such as `torch`, `genesis`, or `lerobot` at module level in unit tests; use targeted imports or `pytest.importorskip()`"), but nothing enforced it.

This adds the guard to the existing `test_hygiene_guards.py`, alongside the other AST-based hygiene rules, following that file's established convention of pairing each guard with a test that proves it fires.

## User impact

Today the failure mode is a collection error naming an import, which the reader has to trace back to an unwritten rule. After this, it is a named violation that points at the fix:

```
npa/tests/.../test_x.py:7 imports torch at module level;
use pytest.importorskip or import inside the test
```

## Why these eight packages

Chosen empirically, not guessed. All eight are absent from the CPU unit-test environment, so a module-level import of any of them breaks collection. `rerun` was deliberately excluded because it *is* installed and therefore is not a hazard. `torch` alone is already `pytest.importorskip`-ed in 18 places, which is the codebase's own evidence that it treats these as optional.

## The guard cannot become stricter than the rule

Scanning only statements directly in the module body is what lets the four supported escapes through, since each is either nested or a call rather than a top-level `Import` node:

- a function-level import
- `try: import lerobot / except ImportError:`
- `if TYPE_CHECKING:` blocks
- `pytest.importorskip("torch")`

`test_heavy_import_guard_allows_the_supported_escapes` asserts all four still pass. A guard that only proves it rejects bad input is half-tested; over-rejection here would block legitimate GPU tests, which is the more expensive failure.

## Validation

- `npa/tests/guardrails/test_hygiene_guards.py`: **17 passed**, including the 3 added
- the rule holds across the tree today — **0 violations in 532 test files** — so this locks in a property that already holds rather than requiring cleanup
- `ruff check` and `ruff format` clean

## Scope

Test-only. No product code, no infrastructure, no dependencies. Hardware: **CPU only**; **VLM review not applicable** — this is a non-visual static guard with no image, video or RRD output.

## Review status

Draft. I am the author and this needs independent review before it is marked ready.


timothy-le7:
# Review: PR 608 (draft) at `0b8cf161e` — sound guard; its suggested remedy is the exact shape B110 reports as a defect

Reviewed at 2026-09-20T07:44Z. Draft, 87 lines, zero reviews. Reviewing it as
visible work rather than a merge candidate, per `PROGRAM.md`.

## The guard does what it says

I reimplemented it and ran sixteen shapes against ground truth — whether
importing the module actually raises and fails collection:

    flagged  breaks   case
      True    True    plain module-level import
      True    True    submodule (torch.nn)
      True    True    from-import
      True    True    aliased (import torch as t)
      True    True    second on a multi-import line
      True    True    cv2
     False   False    function-level              (supported escape)
     False   False    try/except ImportError      (supported escape)
     False   False    TYPE_CHECKING               (supported escape)
     False   False    pytest.importorskip         (supported escape)

    false positives: 0

Zero false positives is the property that matters most here, because this runs
against every test file and one spurious failure would get the guard deleted
rather than fixed. All four documented escapes pass through untouched,
including the two that are subtle — `try/except ImportError` genuinely
attempts the import but handles it, and `TYPE_CHECKING` never executes.

Choosing the eight packages empirically, by checking what is actually absent
from the CPU environment rather than guessing at what sounds heavy, is the
right method and worth saying.

## F1: the error message recommends the shape B110 reports as broken

The guard tells the author:

    use pytest.importorskip or import inside the test

B110, which I filed at 06:00Z from the shard-2 escalation, is exactly
"import inside the test". `npa/tests/workbench/test_detection_storage.py:78`
does a bare `import torch` inside
`test_training_checkpoint_and_metrics_use_each_successful_write`. CI installs
`npa[dev,adapter]` so it passes there, but a checkout installed as plain
`-e npa` gets `ModuleNotFoundError` and a **red test rather than a skip**.

So the two pieces of work disagree about whether a function-level bare import
is acceptable. Both are defensible within their own scope — a module-level
import fails collection for the whole file, which is strictly worse than
failing one test — and `AGENTS.md` does say "targeted imports **or**
`pytest.importorskip()`". But a guard that fires is a teaching moment, and
this one currently teaches a pattern that produces a red test on a plain
checkout.

Recommending `pytest.importorskip` alone would close both. The remediation
text is one line.

## F2: the guard catches nothing today, which is worth stating in the PR

    module-level heavy imports across npa/tests: 0
    test_hygiene_guards.py: 17 passed

That is not a criticism — a preventive guard with no current violations is
exactly what you want, and pairing it with a fixture proving it fires is the
right way to keep it honest. But it means this PR does not fix any present
failure, and specifically **it would not have caught B110**, whose import is
at function level. If B110 was part of the motivation, it stays open.

## F3 (low): the stated rationale is not quite true, and the comment is what the next person will trust

> Only statements directly in the module body execute at import time.

Statements nested in a module-level `if`, `for`, or `with` also execute at
import time. Measured:

    flagged  breaks   case
     False    True    if True: import torch
     False    True    if sys.platform == 'linux': import torch
     False    True    for _ in range(1): import torch
     False    True    importlib.import_module('torch') at module level

All four break collection and none is flagged. I am not asking you to catch
them — they are contrived in a test file, and widening the guard would risk
the false positives that currently sit at zero. The issue is the sentence.
The real invariant is narrower: *the three supported escapes are safe because
each one either handles the failure or never runs*, which is true, rather than
*nesting means it does not execute*, which is not. A maintainer extending this
guard will reason from the comment, and the comment currently licenses an
inference that does not hold.

This is the third time tonight a comment has credited a mechanism other than
the one doing the work — #603's `list_json_uris` docstring cites a
"delimiter-bounded" listing where the isolation actually comes from a trailing
slash. Worth watching for as a class.

One probe correction of my own: my first run measured "does this attempt a
heavy import" and flagged `try/except ImportError` as a miss. That was the
wrong ground truth — attempting is fine, failing collection is the problem.
The table above uses the corrected measure.

## Verdict

**Sound as drafted.** Correct scope, zero false positives, all six real
violation shapes caught, all four escapes preserved, and a self-test that
proves it fires. F1 is the one I would fix before this leaves draft, because a
guard's message is the only documentation most people will read, and this one
currently points at a known-bad pattern.

timothy-le7:

Recording here what went to the lane inbox at 07:44Z.

**The guard works:** it catches 6 real violation shapes, correctly preserves 4 legitimate escapes, and produces zero false positives on the current tree.

**F1 — the error message recommends the defect.** It tells the author to "import inside the test," which is the exact shape B110 reports as defective. A guard that steers people toward a known-bad pattern will keep producing the thing it exists to prevent. Worth rewording to point at `pytest.importorskip` or a module-scoped fixture.

**F2 — it catches nothing today.** Purely preventive, which is fine and worth saying plainly: it would *not* have caught B110, so it shouldn't be credited with covering that class.

**F3 (low)** — the stated rationale that "only module body statements execute" isn't quite true; nested `if`/`for`/`with` bodies at module level execute too.

None blocking. Re-review for merge readiness when it leaves draft.

timothy-le7:
Amended at `0d6b5bb08` per review. Two wording corrections only; no logic change.

**The import-time claim.** "Only statements directly in the module body execute at import time" was false — module-level imports nested in `if`/`class`/`for`/`with` also execute at import time. They are outside this guard *by design*, not by language semantics, so the comment now states the intended direct `Import`/`ImportFrom` scope instead of justifying it with an incorrect claim about Python.

**The diagnostic.** `use pytest.importorskip or import inside the test` → `use pytest.importorskip or an optional-dependency fixture`. Moving an absent dependency's import into the test body only converts a collection error into a test error.

Local results: 17 hygiene tests passed, `ruff check` clean, `ruff format --check` already formatted. Delta is 8 insertions / 6 deletions in one file.

I checked the new comment is true rather than just differently worded, by running `_heavy_module_import_violations` over synthesized files: a direct `import torch` is flagged, while imports nested in `if True:`, `class C:`, and `with contextlib.suppress(ImportError):` are not, and a function-level import stays a supported escape. Detection behavior and escape forms are unchanged.

timothy-le7:
Both unverifiable commit signatures have been repaired; current head is `8904262f4150866cd5444aa46a0e639174602e6e`. Independent byte comparison confirms every replacement tree and parent-relative change matches the original sequence, including original head `0d6b5bb08f67bc0912eba47a2c975e4f201acd69`. Original authors and author dates are preserved, and the original sequence is retained in a private recovery bundle. GitHub verifies both replacement signatures. Fresh hosted checks are required before the readiness verdict carries forward. No source behavior changed.

PR #609, opened 2026-09-20
Recover stage log attribution from durable waves

timothy-le7:
## Summary

- recover a missing stage-level managed-job ID only from the stage projection's
  exact durable `wave_key`, attempt, and stage membership
- pass the recovered job and task identities together to the live log boundary
- fail closed before a live query when matching waves disagree or the wave key
  does not match
- keep explicit projections, legacy no-projection reconstruction, and cached
  object-storage reads authoritative

## User impact

`npa workbench workflow logs --stage` no longer reports a false
`STAGE_JOB_ID_UNAVAILABLE` when a current-format or externally produced partial
stage projection retains `wave_key` but has a blank `managed_job_id`, while the
matching durable wave still carries the exact managed-job identity.

This does not claim recovery for ledgers that predate both projection fields.

## Hardware and acceptance

| Tier | Applicability and result |
|---|---|
| Source | Exact head `e962e6fbd46093a4f4b84ce72bc259ac1959ab77` |
| Hermetic CPU tests | Pass: 97 workflow CLI tests, including exact recovery, no-task fallback, explicit IDs, malformed candidate-wave attempts, missing/mismatched keys, conflicting wave IDs, cached mode, and legacy reconstruction |
| Smoke/static gates | Pass: 114 onboarding smokes; 3,862 guardrails; repository Ruff/format; docs drift; confidentiality; gitleaks |
| Actual CPU/object-storage workload | Pass: production resolver plus authenticated object storage for exact, ambiguous, and key-mismatch ledgers; exact attribution made one instrumented log call, both unsafe controls made zero calls, and cleanup left zero objects |
| Real SkyPilot controller log tail | Not performed: exact Kubernetes/provider ownership verification blocked the optional tiny CPU job before launch; no managed job was created |
| GPU | Not applicable: no GPU code path is changed or needed for this pre-executor diagnostic boundary |
| Visual/VLM | Not applicable: output is text-only |
| Independent review | `ACCEPTED` at the exact head after head-versus-base probes and exact-head gates; zero blocking findings |
| Required CI | Pass at the exact current head |

The authenticated storage workload is real; its `sky jobs logs` callable is
instrumented. This PR therefore does not claim a live controller boundary.

## Objective evidence

- Frozen plan SHA-256:
  `573b0a707d88522a4793cdb81c73459bd6cb62813036ffe7a4d710d2f92351ca`
- Frozen independent-review addendum SHA-256:
  `757c12fc9ca277ffb676b2cacd69db19e1c8c332d49bde4cc251d2b70f9b49f0`
- Baseline `d22893238`: exit 2, zero log-boundary calls,
  `STAGE_JOB_ID_UNAVAILABLE`.
- Candidate `e962e6fbd`: exact current-format projection exits 0 and sends one
  matched job/task pair; conflicting jobs exit 2 with
  `STAGE_JOB_ID_AMBIGUOUS` and zero calls; a key mismatch exits 2 with
  `STAGE_JOB_ID_UNAVAILABLE` and zero calls.
- A complete explicit projection ignores an unrelated malformed wave attempt
  and remains successful. A malformed partial candidate wave is ignored and
  cannot provide attribution.
- Cached mode remains object-storage-only and preserves its prior attribution
  payload.

## Reproduction

```bash
  npa/tests/cli/test_workflow_cli.py \
  -k 'partial_stage_projection or ambiguous_wave_job_attribution or different_wave_key or malformed_wave_attempt or cached_logs_do_not_recover'

npa workbench workflow logs \
  "s3://${BUCKET}/${RUN_PREFIX}/npa-workflow/manifest.json" \
  --stage "${STAGE}" \
  --json
```

Use a disposable run prefix and remove it after the proof. Keep bucket,
endpoint, project, cluster, and managed-job identifiers out of review output.

## Limitations

- When an exact recovered serial wave has no task observation, the command keeps
  the conservative stage-name selector; a controller may still return
  `VERIFICATION_UNAVAILABLE`.
- An explicit stage projection remains authoritative if it contradicts its own
  wave, matching the existing contract.
- A non-numeric attempt on the stage record itself retains the pre-existing
  conversion error; this change hardens candidate-wave attempts.
- A full repository serial run is not claimed here. It was stopped after an
  already-running duplicate lane suite was detected; the focused, smoke,
  guardrail, and required PR CI gates remain the acceptance basis.


timothy-le7:
# Review: PR 609 at `e962e6fbd` — accepted; it refuses to guess, and it labels what it recovers

Reviewed at 2026-09-20T07:55Z. No prior review.

## Falsifiability

    at head                              97 passed
    production reverted to origin/main    4 failed, 93 passed

## The failure mode this had to avoid

Recovering a missing job ID is inference, and the bad outcome is not an error
message — it is confidently showing an operator another stage's logs while
they debug. Silent misattribution is worse than `STAGE_JOB_ID_UNAVAILABLE`,
because the operator has no signal to distrust what they are reading.

It refuses to guess:

    if len(wave_job_ids) > 1:
        return recovered, "", "matching durable waves disagree on managed-job ID"

Ambiguity returns no job ID and a stated reason, rather than the first
candidate. Recovery only proceeds at `len(wave_job_ids) == 1`, and the
candidate set is narrowed by three independent agreements first — attempt,
`wave_key`, and existing `job_id` each filter a wave out on mismatch — plus a
structural check that the stage appears exactly once
(`states.count(stage) != 1`). So a single candidate means one wave survived
four separate ways of being wrong, not that only one was ever considered.

Failing closed *before* the live query is the right placement. A wrong ID that
reaches the log boundary produces real output from the wrong job, which is
indistinguishable from correct output.

## The detail I liked

    recovered["provenance"] = "runtime_wave_attribution_recovery"

The recovered ID is labelled as recovered rather than presented as if it had
been read from the projection. That is the same discipline as #597's
`provider_success`, reviewed twenty minutes ago: keep what the system was
told separate from what the system worked out, and let the consumer decide
how much weight to give an inferred value. Two lanes arriving at that
independently is a good sign for the codebase.

Keeping explicit projections, legacy no-projection reconstruction, and cached
object-storage reads authoritative means recovery is a last resort rather than
a new preferred path, so this cannot mask a regression in the primary sources.

## Verdict

**ACCEPTED at `e962e6fbd`.** Removes a false
`STAGE_JOB_ID_UNAVAILABLE` for partial projections without introducing the
much worse failure of confident misattribution, and marks its own output as
inferred.

timothy-le7:

Recording here what went to the lane inbox at 07:55Z.

**It refuses to guess, which is the right call.** Job-ID recovery proceeds only when a single candidate survives after multiple independent agreements; on any ambiguity it returns no job ID plus a reason rather than its best guess. Fail-closed on disagreement.

**It labels what it infers.** A recovered ID carries `runtime_wave_attribution_recovery` provenance, so a reader can tell a recovered attribution from an authoritative one. That's the difference between recovering data and quietly inventing it.

No blocking findings.

timothy-le7:
Selected for **merge batch 2**, order 6, in the [Codex-validated first ten](https://github.com/nebius/nebius-physical-ai/blob/75e32dd93380c120aee87ea4bccf2c662e0bbbc2/docs/testing/evidence/cursor-first-ten-20260924/README.md). Reviewed public head: `e962e6fbd46093a4f4b84ce72bc259ac1959ab77`.

timothy-le7:
Codex pushed the signed refresh `3725041464469af2ac80d51c701f0aa02afa0831` against current main, including the shared fix for #615’s two-minute Gitleaks job timeout. Required security scanning is preserved.

Local validation: **27092 passed, 135 skipped, 1 xpassed, 364 warnings in 744.10s (0:12:24)**. The dedicated hostile-input suite passed 859 tests; docs drift, guardrails, lint/formatting, confidentiality, full-range Gitleaks, and native scanner comparisons passed. The combined merge sequence passed 27,172 tests with zero failures.

[Open the reviewable local evidence](https://github.com/nebius/nebius-physical-ai/blob/22f1931cf32778e837b4385968adc0309ab08818/docs/testing/evidence/cursor-batch-two-recovery-20260924/local-validation.md) — verified anonymously accessible, with exact source identities, control outcomes, and log hashes. **Hosted CI is still pending; this is not a final merge-ready declaration.** No Cursor capacity was used, and no PR was enqueued or merged.

PR #611, opened 2026-09-20
Preserve active SkyPilot jobs during cleanup

timothy-le7:
## Summary
- align the shared cleanup/drain vocabulary with every non-terminal managed-job status in pinned SkyPilot 0.12.2
- cancel legacy `SUBMITTED` and current `WINDING_DOWN` jobs before any run-owned cluster teardown
- preserve non-terminal duplicate task rows and add command-level cancel→drain→down fault-injection coverage

## User impact
Repeat-safe workflow cleanup no longer skips a still-active queued or output-merging managed job and tears down its run-owned cluster first.

## Hardware and acceptance

| `87df770d1` | Exact run-owned cleanup handles all pinned non-terminal statuses | Host CPU/controller subprocess boundary. GPU and hosted API are not applicable because this runs before resource teardown; VLM is not applicable to text-only state. | 114 onboarding smokes passed; these are not counted as workload proof. | A disposable executable exercised the production queue, exact cancellation, drain, and down subprocess boundaries for both omitted statuses, recording cancellation before either down command. | Base: 9/9 frozen cases failed, with zero cancels, false drain success, terminal row collapse, teardown reached, and the pinned-contract mismatch. Candidate: 9/9 objective cases passed; persistent states produce an explicit error and zero down calls. Existing terminal and active-state controls remain green. | Exact-head independent review: `ACCEPTED`, no blocking findings. | None; required CI is green. |

## Objective evidence
- Candidate: `87df770d1cf74533bcc92643700eb15ca60be643`
- Candidate tree: `dfa834501b31638f976a368e7030fc894e4d55e4`
- Frozen evaluation-plan SHA-256: `cf4846b0876e2f08e1208499ad90c368bcd580f96cdd1d18fa35d52736eab618`
- Pinned runtime observation: SkyPilot 0.12.2 parses seven non-terminal statuses: `PENDING`, `SUBMITTED`, `STARTING`, `RUNNING`, `WINDING_DOWN`, `RECOVERING`, and `CANCELLING`.
- Baseline `SUBMITTED` and `WINDING_DOWN`: no cancel call, drain reported success, a mixed active/terminal task group collapsed to `SUCCEEDED`, and both exact run-cluster down patterns were reached.
- Candidate persistent-state controls: one exact cancel call, drain remains false, explicit non-terminal error, and zero cluster-down calls.
- Candidate convergence controls: command log order is exact cancel first, then the two run-owned down patterns only after queue state becomes `CANCELLED`.
- Complete baseline refresh: all 9 candidate cases fail against base production source; JUnit SHA-256 `5fa0f7266b2838966783f2e81abec235749757f8f4b5201c530ebffc4db20542`.

## Validation
- 9 objective cases passed
- 89 cleanup tests passed
- 142 cleanup/cancellation integration tests passed
- 956 SkyPilot orchestration tests passed, 1 skipped
- 114 onboarding smoke tests passed
- 3,862 guardrails passed
- repository Ruff/format, diff check, built-in confidentiality scan, and gitleaks passed
- required current-head CI passed
- retained complete-unit attempt reached 26% with no reported test failure before the environment terminated it after 22m22s; it was not repeated

## Reproduction
```bash
  npa/tests/orchestration/skypilot/test_cleanup.py \
  -k 'statuses_cover_pinned or all_pinned_nonterminal_statuses or command_orders_cancel_before_cluster_down or every_row_is_terminal'

  npa/tests/orchestration/skypilot/test_cleanup.py
```

## Limitations
- No cloud job or GPU was allocated solely to force a transient scheduler status; the actual production command boundaries execute against a disposable local SkyPilot CLI.
- Unknown future queue statuses retain their pre-existing behavior and are not claimed by this patch.
- NPA workflow labels such as `RETRYING` and `MANIFEST_PENDING` are not raw statuses emitted by the pinned SkyPilot managed-job queue and are intentionally outside this vocabulary.


timothy-le7:
# Review verdict: PR 611 at `87df770d1` — correct and complete, but fully subsumed by 614

Reviewed at 2026-09-20T06:13Z. **The change is correct.** The finding is not
about the code; it is that this PR and #614, which I already accepted, do the
same thing, and nobody has said so out loud.

## Your completeness claim is exactly right

You claim the change aligns the cleanup vocabulary with *every* non-terminal
managed-job status in pinned SkyPilot 0.12.2. Rather than check your list
against another list, I asked the installed library
(`~/.npa/skypilot-venv`, skypilot 0.12.2) what it considers non-terminal,
using its own `ManagedJobStatus.is_terminal()`:

    SkyPilot's non-terminal set : CANCELLING PENDING RECOVERING RUNNING
                                  STARTING SUBMITTED WINDING_DOWN
    yours after this change     : CANCELLING PENDING RECOVERING RUNNING
                                  STARTING SUBMITTED WINDING_DOWN

Identical, seven for seven. Not a superset, not a subset. The claim is exact,
and `WINDING_DOWN` in particular is the one that matters — a job merging batch
output looks idle and is not. 89 tests pass; CI is green, 14 passing.

## The finding: 614 already contains all of this

I accepted #614 earlier tonight. Comparing the two against `origin/main`:

- **every one of your four added source lines is also added by #614**,
  including both explanatory comments verbatim;
- four of your five added tests exist in #614 under identical names with
  byte-identical bodies;
- your fifth, `test_wait_for_jobs_terminal_keeps_all_pinned_nonterminal_statuses`,
  appears in #614 as
  `test_wait_for_jobs_terminal_keeps_nonterminal_and_unrecognized_statuses`,
  which is strictly stronger — it covers unrecognized statuses too.

You created #611 at 03:09Z and #614 at 03:31Z, both branched from `main`
rather than stacked, so #614 absorbed this work twenty-two minutes later
instead of building on it.

**This is safe, which I checked rather than assumed.** I merged both into a
scratch tree: clean in either order, git deduplicates the identical additions,
no duplicated function definitions in either file — the real hazard, since a
test name defined twice is silently shadowed rather than reported — and 293
tests pass on the merged result. So landing both breaks nothing. It is
redundant, not dangerous.

## The distinction worth keeping

These are not the same fix philosophically, and that is the one reason to
consider keeping both rather than just closing one.

Yours completes an **open-world** set: every status SkyPilot currently calls
non-terminal is now listed, so anything unlisted is treated as terminal.
Correct today, exactly as verified above. #614 adds a **closed-world**
predicate: only `SUCCEEDED`, `CANCELLED` and `FAILED*` are terminal —
which, checked against the library, is also precisely its terminal set —
so a status SkyPilot adds in a future release is treated as non-terminal by
default.

The difference only bites when SkyPilot introduces a status nobody has
written down yet. Yours would then tear down a live job; #614's would refuse
to. That is the residual I filed as B106, and it is worth noting your change
narrows B106 substantially: the two legacy CLI consumers that read
`NONTERMINAL_JOB_STATUSES` directly are now complete against the pin, so the
divergence only applies to statuses that do not currently exist.

## What I suggest, and it is your call

I am not going to tell you which PR to land, because there is a real tradeoff
and you are closer to the release pressure than I am:

- **Land 611 alone, first.** It is four lines in one file against #614's
  seven files touching `project_destroy`, the CLI, and cancellation. If the
  `WINDING_DOWN` teardown hazard should stop being live today, this is the
  version with almost no blast radius. #614 then rebases and its duplicate
  hunks vanish.
- **Close 611 as superseded.** #614 is accepted, is a strict superset, and
  gives the future-proof classifier as well. One PR, one review, done.

What I would not do is land both without a decision, because then the record
shows two PRs that each claim to fix this and no statement of which one
actually did.

**Verdict: ACCEPTED on correctness, no code changes requested.** Tell me which
route you take and I will record it; if you land 611 first I do not need to
re-review #614, since I have already verified the merged result.

## One on me

I reviewed #614 in depth — adversarial payloads, mutation testing, and an
enumeration of every managed-job status classifier in the tree — and did not
notice it duplicated an open PR that had been sitting there for twenty-two
minutes. I checked within the branch and not across the lane's open work.
That is the second stacking-and-overlap gap I have caught in my own process
tonight, and I have added a cross-PR overlap check to what I do before
accepting.

Evidence retained: `program/review/evidence/verify_611_status_completeness.txt`.

timothy-le7:
## Independent review: correct, but subsumed by #614

Recording here what went to the lane inbox at 06:13Z.

**The change itself is correct.** That is not the finding.

The finding is that **#614 already contains all of this and is a strict superset.** I accepted #614 at `b1123f912` after verifying its `WINDING_DOWN` false-convergence fix, and that work covers what this PR does plus the closed-world classifier.

**Recommendation: close #611 as superseded** rather than merging both. Landing two overlapping fixes to the same cleanup path creates a merge conflict for no added protection, and the pinned `SUBMITTED`/`WINDING_DOWN` vocabulary this PR introduces is already excluded from #619's scope as an explicit overlap exclusion — so the stack downstream is already built on the assumption that #614 owns this.

Nothing wrong with the code. It just arrived second.

timothy-le7:
## Review lane — triage: this is fully contained in #614, and #614's approach is strictly safer

I have been carrying this PR as untriaged and declining to assert it was redundant without checking. I have now checked, by diffing the two changes rather than from memory.

### Your change, verbatim, is already in #614

```python
# Pinned SkyPilot still parses legacy controller rows in this state.
"SUBMITTED",
# Batch output is still being merged; teardown is not yet safe.
"WINDING_DOWN",
```

Identical entries, identical comments, same file. [#614](https://github.com/nebius/nebius-physical-ai/pull/614) adds exactly this and then continues.

### Why #614 does not merely include it, but supersedes it

This PR adds two statuses to `NONTERMINAL_JOB_STATUSES` — an **open-world** list. Teardown is blocked for statuses on the list, so any status *not* on it is treated as terminal and safe to tear down. That fixes `SUBMITTED` and `WINDING_DOWN` specifically, and leaves the next unlisted status with the same failure.

#614 inverts the test at five call sites:

```diff
- if status and status not in NONTERMINAL_JOB_STATUSES and status != "UNKNOWN":
+ if _is_terminal_managed_job_status(status):
```

```python
return status in {"SUCCEEDED", "CANCELLED"} or status.startswith("FAILED")
```

That is **closed-world**: only explicitly terminal states are terminal, and anything unrecognised is treated as still running. The difference matters in exactly the situation your PR is about — **an unknown status can no longer cause a live job to be torn down.** Yours fixes two instances of the bug; #614 fixes the class.

### Recommendation

**Close this PR in favour of #614.** Nothing is lost: your two entries and their comments survive intact in #614, so the specific cases you identified remain covered.

I am recommending rather than deciding, because the disposition is the owner's. If you would rather keep this one and rebase #614 on top, that also works — but the two should not both land, and right now both are open drafts touching the same lines, which is a merge conflict waiting to happen.

### For context

[#619](https://github.com/nebius/nebius-physical-ai/pull/619) then unifies terminality on top of #614, and I reviewed it as sound. So the chain is this PR → #614 → #619, with each step generalising the last. Only the last two need to land.

timothy-le7:
Closing as superseded by #614. GitHub confirms this exact head `87df770d1cf74533bcc92643700eb15ca60be643` is an ancestor of #614 at `b1123f9128ce998c7bffcd9957a60da49474fa52` (three commits ahead, zero behind). The accepted status coverage and tests are retained there, together with the closed terminal-state classifier.

#614 has independent acceptance and all applicable checks passing. The downstream #619/#620 work continues from it. No implementation is discarded, no branch is deleted, and no PR is merged by this closure. This duplicate is excluded from the delivered-improvement count.

PR #612, opened 2026-09-20
Retain source-frame sampling provenance for VLM requests

timothy-le7:
## Summary

- candidate head: `d116cbbcb`; implementation commit: `3d702087c`, stacked on PR 597;
- upgrade hosted-evaluation evidence to schema v2 with source kind, zero-based source index, source count, and source video timestamp when known;
- retain requested sampling strategy, frame limit, selected indices/timestamps, and explicit metadata-completeness flags in the secret-free request manifest;
- use structured `ffprobe` frame metadata for video timestamps, preserve numeric extraction order beyond frame 999, and fail closed when source counts or timestamps are unavailable;
- cover image sequences, NumPy episodes, known-count videos, and unknown-count video fallback without inferring absent coverage.

## Hardware and acceptance

| Field | Evidence |
| --- | --- |
| Source | Public head `d116cbbcb`; tested implementation `3d702087c` |
| CPU applicability | Frame selection, normalization, hashing, video decode, and manifest construction run locally on CPU |
| GPU applicability | No local GPU applies: model inference is a hosted vision API and provider hardware is unknown |
| Smoke (not workload proof) | 114 onboarding smoke tests passed |
| Actual workload | One real three-image `MiniMaxAI/MiniMax-M3` request completed with HTTP 200, exact returned model, `finish_reason: stop`, 2,153 total tokens, and 3.128406 s latency |
| Objective controls | Image sequence selected `[0,2,4]` of 5; NumPy selected `[0,2,5]` of 6; 2 fps MP4 selected `[0,2,5]` at `[0.0,1.0,2.5]` s; unknown-count fallback retained null coverage |
| Failure controls | Incomplete timestamps fail closed; filename text cannot inject timestamps; extracted frame ordinals sort numerically |
| Visual result | Model distinguished samples-only from scan-plus-surface views but misread visible `1.0 m` as `10 m`; no geometry or visual-acceptance claim |
| Independent review | Reviewed implementation is `review-ready`; `ba77a34f7` and rebased `3d702087c` are range-diff and byte-patch identical; sanitized proof separately approved for publication |
| Remaining blocker | Exact-current-head remote CI |

## Validation

- focused VLM/CLI/SDK tests: 132 passed, 3 skipped;
- exact public head guardrails: 3,870 passed;
- repository lint and diff checks passed;
- built-in infrastructure confidentiality scan: 0 unresolved findings;
- gitleaks: 9 commits and approximately 69.5 KB scanned, no leaks;
- producer recomputation and independent review found zero mismatches across selected frame bytes/metadata, prompt, rubric, request manifest, and raw response.

## Public proof

The reviewer-safe evidence report is committed at:

https://github.com/nebius/nebius-physical-ai/blob/d116cbbcb/docs/workbench/evidence/vlm-frame-sampling-provenance-review.md

Raw media, provider responses and request identifiers, source paths, and operational logs remain private.

## Boundary

`coverage_complete` means each submitted frame has source-index metadata; it does not mean every source frame was submitted. Sampling traceability does not establish temporal causality, task completion, geometry accuracy, collision suitability, physical correctness, or robot safety.

## Current merge-readiness (2026-09-21)

| Required field | Evidence |
| --- | --- |
| Exact source / image | PR head `d116cbbcb03aefb8190952a41a883b8df812aa63`; tested implementation `3d702087c`; no image changed |
| Change proved | Submitted frames preserve source kind, source index/count, video timestamp, requested strategy, selected indices/timestamps, and honest metadata completeness |
| Compute applicability | CPU frame selection, decode, normalization, hashing, and manifest construction; hosted API inference; no local GPU applies and provider hardware is unavailable |
| Smoke / static checks | 132 focused tests passed, 3 skipped; 114 onboarding smoke and 3,870 guardrails passed; these do not replace workload proof |
| Actual full workload | One real three-image MiniMax-M3 request returned HTTP 200, exact model identity, `finish_reason: stop`, usage, and retained latency |
| Objective acceptance | Image, NumPy, known-count MP4, and unknown-count fallback selections matched the frozen expected indices/timestamps and metadata-completeness states |
| Negative / failure controls | Missing timestamps fail closed, filenames cannot inject timestamps, frame ordinals sort numerically, and unavailable coverage remains null rather than inferred |
| Calibrated visual review | Not an acceptance claim: the model distinguished the source views but misread visible scale; no geometry, physical-correctness, or safety inference follows |
| Independent review | Range-diff/byte-patch identity and sanitized proof accepted; current exact-head CI is green with 20 passed and 2 skipped checks |
| Remaining merge blockers | Ordered stack dependency on #597; this head is mergeable with that accepted base and has no separate source, CI, or review blocker |

The public report is anonymously readable (HTTP 200); exact private media and provider payloads remain access controlled.


timothy-le7:
# Review verdict: PR 612 (frame-sampling provenance) at `d116cbbcb` — ACCEPTED

Reviewed at 2026-09-20T06:07Z. **ACCEPTED, no findings.** This is the
cleanest change I have reviewed in this program.

## First, a gap in my own process

I accepted #617 on the strength of its delta over this PR, and this PR had
no review from anyone at the time — zero comments. Reviewing the top of a
stack while its base is unreviewed is not independent review of the top; it
just moves the unexamined surface one commit down. That was my error, not
yours, and this closes it. #617's acceptance stands, but it stands on this
review as well as its own.

## The claim that actually matters, and it holds

`VlmFrameEvidence.sha256` is documented as "Digest of the exact submitted
bytes". Provenance that records intent rather than what went on the wire is
decorative, so I tested the binding rather than read it. Five synthetic
frames with distinct content and distinct dimensions, hashed against the
base64 payload as `_build_request` constructs it:

| frame | sha256 vs wire bytes | decoded dims | byte_count |
| --- | --- | --- | --- |
| frame_0 | match | 64x48 | match |
| frame_1 | match | 65x49 | match |
| frame_2 | match | 66x50 | match |
| frame_3 | match | 67x51 | match |
| frame_4 | match | 68x52 | match |

The structural point is better than the result. `SelectedFrame` carries
`label`, `media_type`, `data`, and the four source fields — and no `width` or
`height`. So the evidence dimensions cannot be copied from a caller-supplied
value; they can only come from decoding the bytes being submitted. That makes
the binding a property of the type rather than of the current implementation,
which is the difference between a guarantee and a passing test.

Source provenance is exact for a known source: kind, zero-based index, count,
and timestamp all correct across all five frames.

## The digest covers the provenance, not just sits beside it

`request_manifest_sha256` recomputes faithfully from the manifest it ships.
More usefully, I checked whether the digest actually protects the new fields:
changing one frame's `source_index` from 0 to 999 moves the manifest digest.
So the provenance is inside the integrity boundary. A digest that covered
everything except the thing the PR adds would have looked identical in review
and been worthless.

The `coverage_complete` and `timestamps_complete` flags are the right
addition. They let a reader tell whether the model saw the whole source or a
subsample, which is exactly the question someone re-reading a VLM verdict
later needs answered and cannot otherwise recover.

## Mutation: five of five caught

| mutation | result |
| --- | --- |
| `source_index` dropped from evidence | caught |
| sampling block dropped from the hashed manifest | caught |
| `coverage_complete` hardcoded true (hides subsampling) | caught |
| `timestamps_complete` hardcoded true | caught |
| evidence schema version reverted to v1 | caught |

A perfect score, and the first I have recorded in this program. Notably the
two that hide subsampling are caught — those are the mutations that would
still produce plausible-looking evidence, so they are the ones worth testing
and the easy ones to miss. 113 tests pass, 1 skipped.

## Claims checked rather than read

Your report states the reviewed implementation at `ba77a34f7` is
byte-identical to rebased `3d702087c`. Verified by hashing the patches with
commit metadata stripped: both `bf5a07c1c1ce01df`. The claim is exact, so the
independent review at the pre-rebase commit legitimately carries.

Cross-checking your report against #617's: both record 2,023 / 130 / 2,153
tokens with different latencies (3.128406s here, 2.962663s there). Identical
token counts across two runs of the same prompt, rubric and image set is what
deterministic tokenization should produce, and the differing latency
confirms they are genuinely two calls rather than one result reported twice.
That is a consistency check your two reports pass without being designed to.

Hardware declaration is correct: hosted API, provider hardware unknown, no
local-GPU claim, client work on CPU. The published report carries no
infrastructure identifiers or credentials.

## On #597 underneath

While mapping the stack I found `56c0315b8`, the commit I issued the #597
verdict against, is no longer an ancestor of the branch — it was rebased to
`2995156d4`. I re-derived rather than assume: source is byte-identical
between the two, and the only content difference is one added evidence
document. My #597 verdict carries unchanged. Flagging it only so nobody is
surprised that the reviewed SHA no longer appears in the history.

## Merge

Nothing blocking from me. CI is pending; per the cancellation behaviour
geometry root-caused, hold the head rather than pushing while it runs.

Evidence retained in `program/review/evidence/`: `probe_612_provenance.py`,
`mutate_612.sh`.

timothy-le7:

Following my acceptance above: CI has now completed at `d116cbbcb` with **20 passing checks, 0 failures, 2 skipped**. No gate is outstanding.

So this PR has, from this lane: an independent review against the substantive claim (provenance binds to the bytes actually on the wire, verified by recomputation rather than by reading), a perfect 5-of-5 mutation score on the controls that matter, a verified byte-identical-across-rebase claim, a clean confidentiality scan, and now green gates.

**Merge-ready from review.** The only reason it is still draft is the stack: it sits on #597 (verdict carries, re-derived after its rebase) and is the base of #617 (accepted, CI still pending). Land in stack order.

Holding the head is worth it here — pushing again would cancel this completed run and put you back to an empty check list.

timothy-le7:
## Review lane, 2026-09-20T10:39:25Z — ACCEPTED at `d116cbbcb`, one non-blocking finding

**Scope correction first.** The diff against `main` is 1777 insertions, but this is stacked on #596 and #597, both already accepted. Against #597's head `2995156d4` the real new work is **608 insertions across 7 files**.

Path-pinned to the candidate worktree. Baseline **92 passed, 1 skipped**.

| mutation | result |
| --- | --- |
| `_source_indices_complete` always returns `True` | **killed** by `test_unknown_video_count_does_not_infer_sampling_coverage` |
| `request_manifest_sha256` computed over `{}` | **killed** by `test_api_result_retains_recomputable_secret_free_evidence` |
| drop the upper bound from the range check | **survived** |

Two mutants dying confirms the harness executed the mutated source, so the survivor is a real gap and not a path artifact.

### F1 (non-blocking): the upper half of the range check is untested

```
frame.source_index is not None and 0 <= frame.source_index < source_count
```

Removing `< source_count` leaves all 92 tests passing, so a frame with `source_index=500` from a 100-frame video would be accepted as **complete sampling coverage** and recorded as valid provenance.

Indices come from `enumerate` today, so I could not produce a real out-of-range value — which is why this is non-blocking, and also exactly why the guard exists. It validates an assumption the extraction code currently satisfies. One test with an out-of-range index closes it.

Mechanism fails closed on unknown source count and missing indices, and the manifest hash is genuinely bound to manifest contents. CPU only; VLM **not applicable** to the mechanism.

PR #613, opened 2026-09-20
Keep declarative workflow outputs visible

timothy-le7:
## Summary
- keep generic `npa.workflow.run.v1` artifact discovery on the exact run root after its manifest appears
- preserve unknown output formats while excluding `npa-workflow/` control ledgers and prefix-similar sibling runs
- retain the historical schema-v1 `artifacts/<stage>/` contract and surface declarative listing failures as workflow-state errors

## User impact
`npa workbench workflow artifacts` no longer loses real declarative outputs such as captions, reports, videos, or new file types when an in-flight run transitions from manifest-pending to a completed durable manifest.

## Hardware and acceptance

| `54d61f256` | Completed generic declarative runs enumerate their exact root without exposing controls or siblings | CPU plus authenticated object storage. GPU/hosted API/VLM are not applicable to nonvisual discovery. | 114 onboarding smokes passed; these are not counted as storage proof. | Exact-head production resolver/listing against authenticated object storage returned two run outputs, including an unknown extension; exact-stage listing returned one; cleanup left zero objects. | Base returned 0/2 completed-run outputs and leaked `PermissionError`. Candidate passes completed/pending stability, exact stage-component filtering, control/sibling isolation, typed failure, legacy-layout controls, and both normally produced bare-root collision cases. | Exact-head re-review: `ACCEPTED`; independently reproduced 2,436 orchestration/CLI tests and 3,862 guardrails. | None; required CI is green. |

## Objective evidence
- Initial candidate: `ecaa213869c2e8f1b83873ed29c5be6f3902724d`
- Remediated candidate: `54d61f256fe5efad0b279e109d503e4b179ab0c6`
- Candidate tree: `ab803eddec9ce18846dbd7631e7f50677dd8f425`
- Frozen evaluation-plan SHA-256: `57fbc803f7139c41fa7ceffaec1bcc32f8e17ae111710abe1fd03da79e7fac47`
- Review-addendum SHA-256: `2d8466ea9a62f80496090b1a028db6944ceb8560a8752d74984481af4d3d050f`
- Baseline: completed run-root case returned 0 of 2 outputs; exact stage case returned 0 of 1; provider failure escaped as raw `PermissionError`; historical schema-v1 control returned 1 of 1.
- Candidate: seven objective cases pass, with unknown-format visibility, exact path-component filtering, stable pending/completed layout, unchanged legacy behavior, and both review-found bare-root controls.
- Remediation: schema selection requires `run_prefix_uri` to prove the exact parent of a dedicated `npa-workflow` control prefix. A bare v1-shaped manifest and every normally produced shape for a bare run literally named `npa-workflow` retain historical listing.
- Live storage: returned `render/result.future-format` and `reports/metrics.json`; stage `render` returned only its output; durable control objects and a prefix-similar sibling were excluded; zero objects remained after cleanup.

## Validation
- 7 objective cases passed
- 30 run-resolution tests passed
- 1,489 declarative workflow orchestration/workflow CLI tests passed, 10 skipped
- 114 onboarding smokes and 3,862 guardrails passed
- repository Ruff/format, diff check, built-in confidentiality, and gitleaks passed
- independent exact-head review reproduced 2,436 orchestration/CLI tests (11 skipped), 3,862 guardrails, Ruff/format, confidentiality, and gitleaks
- required current-head CI passed

## Reproduction
```bash
  npa/tests/orchestration/npa_workflow/test_run_resolution.py

  npa/tests/orchestration/npa_workflow/test_run_resolution.py \
  npa/tests/cli/test_workflow_cli.py
```

## Limitations
- This enumerates objects below one already resolved exact run prefix; it does not broaden run discovery or bucket access.
- The stage selector remains an exact path-component filter. Outputs whose path does not contain the logical stage name remain visible in the unfiltered run list.
- Run-root discovery exposes every non-control file, including inputs and specs; it does not classify repository-specific output formats.
- The internal limit bounds objects scanned before control/stage filtering, so filtered listings may return fewer artifacts than the limit.
- The SDK's historical direct artifact helper is unchanged; this claim is for the shared resolver and workflow CLI.
- A deliberately degenerate `npa-workflow` run whose own manifest declares the shared parent can still select that parent; normal NPA-produced manifests do not use this shape.
- Copying a run tree to another bucket without rewriting `run_prefix_uri` silently falls back to the legacy layout; in-place manifests match by construction.


timothy-le7:
# Review: PR 613 at `54d61f256` — accepted; the sibling-prefix defense is the stronger version of the one in #603

Reviewed at 2026-09-20T07:53Z. No prior review. 23 production lines,
131 test lines.

## Falsifiability

    at head                              30 passed
    production reverted to origin/main    3 failed, 27 passed

Three regressions for a 23-line production change is proportionate.

## Both isolation properties hold

I built a bucket designed to break them:

    all artifacts returned:    out/a.json, eval/r.mp4, eval2/r.mp4,
                               stage/nested/deep.bin
    foreign runs excluded:     True   (runs/abc-2/..., runs/abc/wf2/...)
    control ledger excluded:   True   (npa-workflow/manifest.json)
    unknown format preserved:  True   (nested .bin)

    stage='eval' -> ['eval/r.mp4']
    does eval2 leak into eval? False

Two separate substring traps, both avoided. The run isolation comes from
`run_prefix = declarative_run_root + "/"`, so `runs/abc/wf/` cannot match
`runs/abc-2/...` or `runs/abc/wf2/...`. The stage filter uses
`stage not in relative.split("/")` — exact segment membership rather than
a substring test — so `eval` does not match `eval2`. The second one is easy to
get wrong and is usually only noticed when two stages are named unluckily.

Preserving unknown output formats while excluding the control prefix is the
right default for an artifact listing. A whitelist of known extensions would
have quietly dropped exactly the new file types this is meant to surface.

## The part that is better than #603

I reviewed #603 an hour ago, which fixes the same class of bug — a run
ingesting a neighbour's objects because one id lexically prefixes another —
and it earned an acceptance. This does something #603 does not:
`has_declarative_control_prefix` requires the manifest's `run_prefix_uri` to
equal `s3://{bucket}/{declarative_run_root}` exactly, so the root derived by
string manipulation is cross-checked against an independent record before it
is used.

That matters because `removesuffix("/npa-workflow")` is a string operation on
a path, and string operations on paths are how this bug class starts. Deriving
a value and then proving it against a second source is a stronger pattern than
deriving it carefully, and it is worth both lanes converging on.

There is also a client-side `key.startswith(run_prefix)` re-check after the
server-side `Prefix=`. Redundant today, and worth keeping: it is the line that
will still be correct if someone later changes how the listing is issued.

## Note for the lane

Two independent PRs in the same review window hit the sibling-prefix problem
in different subsystems — Insights storage in #603, declarative artifact
discovery here. That suggests the pattern is latent elsewhere rather than
being two coincidences. The shape to search for is a prefix or path built by
concatenation or `removesuffix` and then used in a `startswith` or S3
`Prefix=` without a trailing delimiter, and a name compared with `in` against
a path string rather than against its segments.

## Verdict

**ACCEPTED at `54d61f256`.** Real user-visible defect — declarative outputs
disappearing when a run transitions from manifest-pending to completed —
fixed without widening what the listing can reach, and with the derived run
root validated against the manifest rather than trusted.

timothy-le7:

Recording here what went to the lane inbox at 07:53Z.

**Both isolation properties verified** against an adversarial bucket: foreign runs excluded, the control ledger excluded, unknown formats preserved rather than dropped, and the stage filter matches on exact path segments so `eval` does not match `eval2`.

**The derived run root is cross-checked against the manifest** rather than trusted, which is what makes this stronger than #603's version of the same defense — #603 relies on a trailing slash for isolation, this validates the derivation.

**Pattern note (filed as B119):** this is the second independent appearance of the sibling-prefix bug in unrelated subsystems — here in artifact discovery, and in #603's Insights storage. Two instances in one review window suggests the prefix-matching idiom itself is the hazard rather than these two call sites.

timothy-le7:
Selected for **merge batch 1**, order 2, in the [Codex-validated first ten](https://github.com/nebius/nebius-physical-ai/blob/75e32dd93380c120aee87ea4bccf2c662e0bbbc2/docs/testing/evidence/cursor-first-ten-20260924/README.md). Reviewed public head: `54d61f256fe5efad0b279e109d503e4b179ab0c6`.

PR #614, opened 2026-09-20
Fail closed on contradictory cancellation state

timothy-le7:
## Summary
- reject contradictory durable-active/live-absent managed-job evidence before workflow cancellation can report a safe no-op
- keep standalone cancel non-terminal while allowing explicit project destroy to continue only from validated exact-absence evidence
- cover mixed live/absent jobs, enclosing runtime/manifest state, `WINDING_DOWN`, unknown future states, and teardown dependency controls

## User impact
`npa workbench workflow cancel` can no longer turn a durable non-terminal job whose scheduler record disappeared into `NO_ACTIVE_JOB` and a terminal/no-op teardown receipt. Standalone cancel remains exit 2 and preserves the contradiction. An explicit `npa destroy --all --yes` transaction may continue run-owned controller/cluster teardown only when every live exact job cancelled cleanly and every remaining error is an exact verified-absence contradiction.

## Hardware and acceptance

| `b1123f912` | Contradictory state fails closed; explicit destroy has one narrow convergence path that cannot outpace managed-job drain | Host CPU provider-boundary and command-level fault injection. GPU, object storage, hosted API, and VLM are not applicable to this nonvisual teardown decision. | 114 exact-head onboarding smokes passed; not counted as objective proof. | Production cancellation assessment, cleanup drain/convergence, CLI result construction, and project-destroy dependency execution ran with exact absent, mixed active/absent, `WINDING_DOWN`, unknown future state, cleanup-failure, and broadened-payload controls. | Initial candidate fixed false `NO_ACTIVE_JOB` but blocked destroy. The first remediation exposed a `WINDING_DOWN` false-convergence path. This head uses closed terminal classification, exact conflict-error matching, and independent validation of every live/cancelled/conflict row. | Replacement exact-head review: `ACCEPTED`; 150 adversarial controls found no status or payload that authorized teardown while a job remained live. | Exact-head automatic CI is green: 20 succeeded, 2 skipped, 0 failed; local full gate also passed. |

## Objective evidence
- Initial candidate: `f6ac237ecafcda090827d0771da29db5bfae7ec4`
- First remediated candidate: `04fd0d6ea341e628c5cb6a16186f6bd7e6826ac8`
- Incorporated cleanup-status ancestor: `87df770d1cf74533bcc92643700eb15ca60be643`; duplicate PR #611 is closed as superseded by this self-contained head
- Current candidate: `b1123f9128ce998c7bffcd9957a60da49474fa52`
- Current tree: `8ac93e3ff549613c6f55cd3055782e14efba10c8`
- Frozen evaluation-plan SHA-256: `a7d5633618f3162dc92acf3a7809e7317c248f5dc78eff8e08c50a448f3dc15b`
- Review-addendum SHA-256: `f79bac0b43b3290173798374bf6bf7ddae4ecfe997a6c63d3567d22b8fba4009`
- Baseline: nine supported durable non-terminal states plus exact queue absence all returned `NO_ACTIVE_JOB`, no error, and `no_cancellation_needed=true`; `WINDING_DOWN` without an ID and mixed terminal/non-terminal history also failed open.
- Initial candidate: those identities remained in `absent_jobs` with `VERIFICATION_UNAVAILABLE`, but one stale row permanently blocked `npa destroy`, including after live sibling jobs cancelled.
- First remediation: cancel still exits 2 and writes no terminal receipt. Its JSON identifies exact absence conflicts and sets `owned_teardown_allowed` only when they are the sole remaining errors. `npa destroy` validates that structure, records a degraded workflow phase, and continues owned dependencies.
- Second review found that cleanup could classify a still-live `WINDING_DOWN` job as terminal and that the destroy validator accepted a synthetic uncancelled live row.
- Current candidate: managed-job terminality is closed-world, so `SUBMITTED`, `WINDING_DOWN`, and unknown future states preserve infrastructure. Cancellation exposes the exact conflict-error set; project destroy requires exact equality and independently validates every absence conflict, cancelled live row, and durable-terminal row.

## Validation
- 168 targeted second-review controls passed
- all 293 cleanup, cancellation, workflow CLI, and integrated-lifecycle tests passed
- 152 destroy regressions passed, 5 skipped
- 2,479 broad orchestration/CLI regressions passed, 11 skipped
- independent review passed 6,012 orchestration/CLI tests (15 skipped, 1 xpassed) and 150 adversarial controls
- 114 exact-head onboarding smokes passed
- 3,862 exact-head guardrails passed
- repository Ruff/format, diff check, built-in confidentiality, and gitleaks passed
- a relative-`PYTHONPATH` full run imported the base worktree after `make` changed directory and is excluded
- corrected full suite: 23,956 passed, 326 skipped, 13 deselected, 1 xpassed; its only four failures required optional `torch`, and all four passed separately against the same candidate source with Torch 2.6.0 CPU
- exact-head automatic CI is green: 20 succeeded, 2 skipped, 0 failed

## Reproduction
```bash
PYTHONPATH=npa/src npa/.venv/bin/python -m pytest -q \
  npa/tests/orchestration/skypilot/test_cleanup.py \
  npa/tests/orchestration/npa_workflow/test_cancellation.py

PYTHONPATH=npa/src npa/.venv/bin/python -m pytest -q \
  npa/tests/cli/test_workflow_cli.py \
  npa/tests/unit/test_pr218_integrated_lifecycle.py
```

## Limitations
- Standalone cancellation intentionally preserves contradictory durable state and never terminalizes it.
- Only explicit project destroy consumes the structured allowance, and its workflow phase remains degraded for audit even when owned downstream teardown proceeds.
- Exact identities supplied by the operator with no contradictory durable active claim may still converge as absent.
- No controller was destroyed to manufacture a transient absence; doing so would violate the ownership and cancel-before-destroy contract.
- This head includes the independently accepted PR #611 pinned-status commit, so the safety fix is self-contained; PR #611 is now closed because its exact head is an ancestor of this candidate.
- A future live status fails closed during pre-cancel re-verification but cannot yet be cancelled through that boundary.
- The destroy-side validator does not independently cross-check payload `run_id`; the caller still invokes cancellation with the exact run ID.
- Durable-terminal rows skip live lookup and retain downstream closed-world controller drain plus SkyPilot's own in-progress guard as independent protections.


timothy-le7:
Independent review at exact head `04fd0d6ea`: **APPROVE**, with three non-blocking findings.

All three properties you asked about hold, verified directly rather than via your tests:

1. **Cannot widen to provider/cleanup failure.** 21 adversarial payloads against `_owned_workflow_teardown_allowance` (extra provider errors, live-not-absent conflict jobs, forged state/outcome/flag, blank and duplicate IDs, arity mismatches both directions, non-dict/non-JSON) were all handled correctly. Cleanup failure is excluded twice — the producer's `and not cleanup.errors`, and the consumer's arity check, since `errors` concatenates cleanup errors.
2. **Cancel-before-destroy ordering preserved.** I checked this as a scope question first, since the dependency gate admits `degraded` dependencies. That gate and the `degraded` status assignment are byte-identical to base (1096/1151, 1228/1295) — this PR only moves a cancel failure from `phase_errors` to `phase_warnings` under the narrow allowance.
3. **No terminal receipt for the contradiction.** Both allowance outcomes fall through to the `verification_failed` default, which is not in `TERMINAL_STATES`.

Your suites pass at this head (181).

**Findings — untested safety gates.** I also removed each gate one at a time and reran your suites. Five of eight mutations were caught; three survived:

- **F1 (worth one test):** dropping the blank/duplicate conflict-ID check survives. I confirmed what it holds back: padding the conflict list to `["7","7"]` lets its length absorb a second error, and that error can be a provider failure. As shipped this is rejected; with the check removed the allowance is granted. That is the exact widening you asked about, guarded by an untested check.
- **F2:** the `is not True` strictness is untested. Note that under a truthiness relaxation the string `"false"` would be accepted.
- **F3:** the producer's arity invariant is untested, backstopped today by the consumer's. But both sides implement the *same* proxy (`len(errors) == len(conflicts)`), so the consumer's re-validation is not independent — it would not catch a producer bug that preserved arity. Making the check content-based (structured errors with a `kind`, consumer requires every error be `durable_absence_conflict`) would make the layers genuinely independent and delete F1's whole class.

None of these block. The gates are correct; they rest on less test than the design implies, on a path whose blast radius is project destroy.

Not performed: no live destroy; reviewed as pure function plus call site. CI reported no checks at review time, so I ran the three suites locally. VLM not applicable (no visual artifact).

timothy-le7:
**Correction: my APPROVE at `04fd0d6ea` is superseded and should not be relied on.** It missed a real defect, and I want the record to say so plainly.

I verified the payload validator hard — 21 forge attempts, mutation testing of eight gates — and treated it as *the* gate. It wasn't. The convergence classification that runs before the payload is ever produced was the other path to the same bad outcome, and I didn't check it.

Worse, the signal was in the diff I read. I saw `WINDING_DOWN` being added to `_NONTERMINAL` in `cancellation.py` and moved on without asking who else classifies that state. Reproduced from each head's own source:

| | `04fd0d6ea` (I approved) | `b1123f912` |
|---|---|---|
| cancellation: `WINDING_DOWN` cancellable | True | True |
| cleanup: `WINDING_DOWN` terminal/converged | **True** | False |
| closed-world cleanup classifier | No | Yes |
| disagreement lets destroy proceed on a live job | **Yes** | No |

The PR added the state to one classifier and not the other, and `cleanup.py` was open-world (`status not in NONTERMINAL_JOB_STATUSES` ⇒ terminal), so `WINDING_DOWN` fell through to converged while pre-cancel re-verification called it cancellable. Destroy continues while batch output is still merging.

Two notes on the remediation, which I've read but not yet fully reviewed:

- `_is_terminal_managed_job_status` being **closed-world** is better than adding the missing state. Only `SUCCEEDED`/`CANCELLED`/`FAILED*` converge, so an unknown or future SkyPilot status preserves infrastructure rather than authorizing teardown. That is the correct default for this path, and it removes the whole class rather than the instance.
- Replacing count-based conflict/error matching with exact error equality closes my F3, which flagged that producer and consumer shared the same arity proxy and so weren't independent layers.

I have not yet completed review of `b1123f912` — I have not reproduced the forged uncancelled-live payload, and per your request I'm not inferring anything from the broad gates still running. Treat this PR as having **no current review approval**. Next verdict will be bound to `b1123f912` explicitly.

Generalizing the lesson for my own reviews: when a PR adds a state to an enum or status set, enumerate every classifier of that state, not just the one the diff touches.

timothy-le7:
**ACCEPTED** at exact head `b1123f912` (tree `8ac93e3ff`). Both reproductions you asked for are done.

**`WINDING_DOWN` false convergence** — reproduced at `04fd0d6ea` (cancellation says cancellable, cleanup says converged, destroy proceeds on a live job) and confirmed closed here. That's the defect my earlier approval missed; already corrected publicly.

**Forged uncancelled-live payload — could not get one past.** 24 adversarial payloads, all correct. The six aimed at your closed-world sweep — live `RUNNING`, `WINDING_DOWN`, `SUBMITTED`, an unknown/future status, an empty status, and a status omitted entirely, none in `cancelled_ids` — are all rejected, while the two legitimate shapes (terminal `SUCCEEDED` uncancelled; `WINDING_DOWN` properly declared cancelled) still pass. Also rejected: unrecognized `live_outcome`, `durable_terminal` with a non-terminal or empty persisted state, conflict rows with no genuine durable contradiction, cancelled rows already terminal or actually absent, conflict/cancelled overlap, duplicate job rows, blank errors.

Requiring `raw_errors == raw_conflict_errors` is the right shape and closes my F3 — the check is content-based now rather than an arity proxy, so the two layers are genuinely independent.

293 tests pass. Per your instruction I inferred nothing from the broad gates still running; GitHub reports no CI checks registered.

**F1 (third mention) — the duplicate-ID check is still the sole barrier and still untested.** I checked whether exact error equality subsumes it. It doesn't: a forged payload can declare the provider error inside `durable_absence_conflict_errors` too, satisfying equality. With conflicts padded to `["7","7"]` and a provider failure as the second "conflict error", the payload is rejected as shipped and **accepted** with that one check removed. Removing it also leaves your 293 green. One test closes it.

**F2 — mutation coverage.** Five of eight caught, and the important ones are protected: reverting cleanup terminality to open-world is caught by 5 tests, and your headline uncancelled-live sweep is caught — so the remediation is test-protected, not merely correct. Surviving: the duplicate-ID check (F1), the `is True` strictness, and your new conflict/cancelled overlap gate.

**F3 — "all paths share that classifier" is overstated (pre-existing, filing as backlog).** This is the check I failed last time, so I enumerated every classifier of managed-job status rather than trusting the diff. `cli/cleanup.py::_nonterminal_jobs` and the controller rebind guard in `cli/skypilot/__init__.py` still test membership against the legacy `NONTERMINAL_JOB_STATUSES`. They agree with `_is_terminal_managed_job_status` on all eleven named statuses — your `WINDING_DOWN` fix does reach both — but invert for one neither names: `DRAINING_V2`, `PROVISIONING`, `UNKNOWN` and empty all read as terminal there and live under the new predicate. `UNKNOWN` and empty aren't hypothetical (`_job_statuses` yields `""` for a row with a missing status). At those sites the queue would report `verified_terminal_only` and rebind would be permitted while that job is live.

Both call sites are unchanged at base and this PR doesn't touch either file, so it's pre-existing and not a blocker — the PR strictly improves both. But the remediation note claims coverage of "controller inventory paths," which isn't accurate.

Not performed: no live destroy; VLM not applicable.

timothy-le7:
Selected for **merge batch 1**, order 3, in the [Codex-validated first ten](https://github.com/nebius/nebius-physical-ai/blob/75e32dd93380c120aee87ea4bccf2c662e0bbbc2/docs/testing/evidence/cursor-first-ten-20260924/README.md). Reviewed public head: `b1123f9128ce998c7bffcd9957a60da49474fa52`.

PR #615, opened 2026-09-20
Fail closed on corrupt submission receipts

timothy-le7:
## Problem and outcome

A damaged, wrong-schema, identity-mismatched, or symlinked local submission
receipt could previously be treated like an absent receipt. A restart could
then overwrite uncertain evidence that submission had begun with fresh
`NOT_SUBMITTED`-looking state.

This change makes every receipt mutation path fail closed:

- uncertain pre-launch state is preserved byte-for-byte and blocks provider
  submission;
- corruption discovered after provider acceptance returns the accepted status
  and exact job identity with a sanitized warning;
- optional artifact handoff remains non-fatal, while storage credentials,
  loaded agent auth, URLs, and provider diagnostics are redacted before return,
  logging, or persistence.

The current-main integration also preserves main's strict server-selected
artifact provenance and pagination checks.

## Current exact source and validation

- Head: `eea74da2e0e084293e3facb4f2ed735b8c35b6d7`
- Tree: `93862c29d55bdabf4c11d5d95d44932f73e992b5`
- Main integrated: `99be7d39f4481aa2c29f72204175698da1a7a258`
- Frozen evaluation-plan SHA-256:
  `a0978f81e5330c5eec039998f2dc5348fdf74a0e66b6744c963433a96c17c118`

The frozen baseline failed all 18 receipt cases; the deep diagnostic baseline
failed 6/6 controls. On the current exact tree:

- 160 focused receipt/artifact/staging/render tests passed;
- 129 focused workflow CLI/runtime tests passed;
- Ruff, format, diff, built-in confidentiality, and gitleaks passed;
- a fresh independent review re-ran 239 scoped tests, proved eight strict
  provenance/selection/request definitions AST-identical to current main, and
  accepted this exact head with zero findings;
- two in-memory credential-propagation removals each exposed the intended
  synthetic secret, proving both storage and loaded-agent redaction controls
  are load-bearing;
- automatic CI is running on this published exact head.

Prior full-suite and independent-review results belong to
`7317da20a9b3d67aee11fde3a6d4ac22b1d0279e`; they remain historical evidence,
not a substitute for current-head acceptance.

## Applicability

| Field | Current evidence |
|---|---|
| Source/image identity | Exact source head/tree above. No container image is changed. |
| Actual workload | Production receipt mutation/planning, provider-boundary fault injection, artifact discovery/handoff, and final receipt writes run on host CPU against temporary corrupt receipts and adversarial diagnostics. |
| Negative controls | Truncated/invalid/non-object/wrong-schema/wrong-identity/symlink receipts; pre-launch write failures; post-acceptance corruption; provider/storage/agent credential-bearing failures. |
| GPU / hosted API | Not applicable: this is local workflow bookkeeping and diagnostic handling. |
| Visual / VLM | Not applicable: no image, video, or RRD quality claim is made. |
| Remaining blockers | Exact-head automatic CI. |

## Limits

The patch preserves uncertain receipts; it does not infer whether their jobs
launched or add a repair command. Operators must reconcile the retained bytes
and printed provider identity. Read-only callers may still use the forgiving
compatibility loader, while all mutation paths use strict inspection.


timothy-le7:
Independent review at exact head `586be99aa`: **APPROVE**, with one fix I'd like before merge and a coverage finding.

**Your three asks all hold.**

1. **No pre-launch write became best effort.** I enumerated all sixteen receipt-write sites and classified each. Three are best-effort plus one conditional suppression, and all four are post-launch/post-success; CLI 2441 and both `record_submission_plan` sites stay strict, as do all four writes in `artifact_load.py`. Those last ones are strict *and* run during the optional handoff, but your claim survives because `_load_paidf_artifact` wraps the call in `except Exception`, so they convert to a partial result. `stage-src` routes into `_fail`, which is a hard failure.
2. **Only accepted/adopted with an exact ID suppress.** Correct on both conjuncts; `.strip()` means a whitespace-only ID re-raises.
3. **Warnings cannot include receipt contents or credentials.** Verified empirically, not just by reading — `redact_text` is called without a `secrets` argument, so only pattern redaction applies and an arbitrary secret would survive if anything echoed it. Nothing does, because every mismatch message prints the *expected* value from the caller rather than the *found* value from the receipt. I planted a token-shaped and an arbitrary secret and drove all seven corruption paths (invalid JSON, non-UTF-8, non-dict, bad schema, wrong run id, wrong project, empty). No planted value reached any warning.

233 tests pass at this head.

**F1 — please fix: `detail` is unredacted where its sibling is redacted.** In `_load_paidf_artifact`, `detail` interpolates `{exc}` raw while `receipt_warning` from the *same* exception goes through `redact_text`. The enclosing block wraps `_storage_client(...)`, which holds `AWS_SECRET_ACCESS_KEY`, and `detail` is both returned and persisted into the receipt.

To be precise about what I showed: I did **not** demonstrate the key leaking — realistic endpoint failures didn't surface it. I did show `detail` carries raw botocore text verbatim, including full endpoint URLs with query strings, persisted unredacted. So: latent hazard, not a live leak, and the fix is one `redact_text` call that closes the class. Cheap enough to land in the PR that's about receipt handling. Not worth a re-review — push it and I'll treat this as approved.

**F2 — the new safety boundary is correct but unprotected.** Removing one gate at a time and rerunning your suites (233, incl. workflow CLI), four of five mutations survived: dropping the exact-job-ID requirement, accepting any transaction state, making the pre-launch write best-effort, and dropping warning redaction. Only "corrupt receipt silently treated as empty" was caught (35 failures).

That shape is worth understanding rather than just patching: N4's 35 failures show the *mechanism* is well covered — a corrupt receipt raises at the `submission_state` layer. What's uncovered is whether any given call site still propagates it, which is why the pre-launch mutation survives. Your "provider submit control proves zero calls" evidence tests the mechanism, not the site. N1/N2 are the two conjuncts of the suppression gate — the entire safety boundary this PR adds — and neither is exercised.

Four tests close all of it. Neither finding blocks; on a path where a lost receipt means a run you can't later cancel, I'd rather it stay right by construction.

Not performed: no live submit; CI still running at review time so I ran suites locally. VLM not applicable.

timothy-le7:
# Re-review: PR 615 at `8691395874` — my finding is fixed; the fix inverted the asymmetry

Reviewed at 2026-09-20T06:21Z. Head and tree match your claim exactly
(`8691395874b7…`, tree `28486c306c25…`), the four-file diff is as described,
and the frozen follow-up plan hashes to
`bb0b36e77e6f09a1f96012b5df801d5e9acb7d7bd342e1da1624c92a0d697200` — your
stated digest, so the plan you executed is the plan you froze.

## My original finding is genuinely fixed

I raised that the optional artifact-handoff `detail` copied raw provider
exception text while its receipt-warning sibling was sanitized. I tested the
new path by calling it, not by reading it: five adversarial provider
exceptions carrying a presigned-URL signature and credential, an
`aws_secret_access_key=` assignment, a `ghp_` bare token, and two explicitly
supplied credential values.

**Handoff detail: 0 leaks out of 5.** Every secret redacted.

## The new finding: you over-corrected one side

The same five cases through `_submission_receipt_warning`:

    [ok  ] url_query_string
    [ok  ] secret_assignment
    [ok  ] bare_token
    [ok  ] explicit_credential_patterny
    [LEAK] explicit_credential_plain
           output: submission receipt was not updated: login failed for
                   password hunter2

`_load_paidf_artifact` calls
`_sanitized_failure_reason(exc, secrets=tuple((credential_values or {}).values()))`.
`_submission_receipt_warning` calls `_sanitized_failure_reason(exc)` with no
`secrets` at all, so it gets pattern-based redaction only. Anything that
*looks* like a credential is still caught — which is why four of five pass —
but a low-entropy supplied value such as a plain password survives, because
no pattern matches it and nothing told the redactor it was a secret.

This is the exact asymmetry I reported, inverted. Before: handoff raw,
receipt sanitized. Now: handoff sanitized with explicit secrets, receipt
sanitized only by pattern.

Your re-review request asks me to verify that "optional handoff detail **and
receipt warnings** redact … explicitly supplied credential values." Against
that sentence, the handoff half is true and the receipt half is false.

**It is one line.** I checked whether the value is even reachable there
before asking: the `_record_transaction` call site is nested inside
`submit_cmd`, which has `submit_credentials` and `resolve_submit_credentials`
in scope. The second call site, in `_try_optional_submission_update`, has no
credential name in scope, so if you would rather narrow the claim than
plumb a value through, that is defensible for that one — but then the claim
should say pattern-based redaction rather than explicitly supplied values.

Severity is low and I want to be straight about that: it needs a plain,
pattern-invisible credential to appear inside an `OSError`/`ValueError` from
a receipt write. It is a claim-accuracy problem more than a live leak.

## Your other four items verify

**Item 2, warning-only scope — exact.** The gate reads:

    accepted = str(payload.get("state") or "").lower() in {"submitted", "adopted"}
    if not accepted or not str(payload.get("job_id") or "").strip():
        raise

Both conditions required, `raise` otherwise, and only `OSError`/`ValueError`
are caught so anything else propagates. Fail-closed, as claimed.

**Item 3, pre-launch strictness** — the same `raise` carries it to the
standard CLI boundary before any submit call.

**Item 4, mutation protection — this is the part that impressed me.** My
other finding was that these gates had no direct mutation protection. I ran
five removals against 77 tests:

    [CAUGHT] handoff detail reverts to raw exception text
    [CAUGHT] handoff drops the explicitly supplied secrets
    [CAUGHT] sanitize_reason removed
    [CAUGHT] redact_text removed
    [CAUGHT] no redaction at all

**Five of five.** Baseline 77 passed, restored clean afterwards. The second
one is the telling result: you have a test that specifically proves supplied
credential values are redacted in the handoff — which is precisely the
assertion the receipt-warning path lacks, and why that gap survived a
mutation suite this strong.

**Item 5** — accepted/adopted status and exact job identity remain visible
after post-launch corruption, consistent with the gate above.

## Verdict

**ACCEPTED with one non-blocking finding.** The hardening is real, my two
original findings are addressed, and the mutation coverage is the strongest
I have seen from this lane. Either pass `submit_credentials` into the
receipt warning or narrow the claim; I do not need to re-review for a
one-line change — tell me which you chose and I will record it.

CI at this head is 2 passing, 15 pending, no failures. Holding the head is
worth it; a push now cancels the run and empties the check list.

Evidence retained: `program/review/evidence/probe_615_redaction.py`,
`mutate_615b.sh`, `verdict_615_followup.txt`.

timothy-le7:
# PR 615: the receipt-warning finding is closed at `fb9e038f6`

Verified at 2026-09-20T06:45Z. Candidate tree `b14c8488a549` matches your
claim. **You chose the harder option and it holds.**

## The plumbing is real

You could have narrowed the claim to "pattern-based redaction" and been
defensible. Instead you passed the resolved values through every reachable
path, and the diff matches the description exactly:

- `_submission_receipt_warning` now takes `secrets` and forwards it;
- the transaction recorder builds `submission_redaction_secrets` from
  `submit_credentials.secret_values.values()` plus `resolved_access_key` and
  `resolved_secret_key`, and passes it;
- `_try_optional_submission_update` accepts and forwards `secrets`;
- both artifact-handoff receipt writes pass their `credential_values`.

## The tests bind it, which is the part I check hardest

Plumbing that no test enforces is plumbing that a later refactor removes
silently. I dropped the secrets at each newly plumbed site:

    [CAUGHT] recorder drops secrets        -> 1 failed, 77 passed
    [CAUGHT] optional update drops secrets -> 2 failed, 76 passed
    [CAUGHT] recorder secret tuple emptied -> 1 failed, 77 passed

Three of three, each caught independently, restored clean at 78 passed.

One note so you can disregard a result you might otherwise chase: my original
probe still shows a leak when it calls `_submission_receipt_warning(exc)`
directly with no secrets. That is now **correct behavior**, not a residual
bug — the helper's default is empty, and a caller with no credentials in
scope genuinely has nothing to pass. The probe was the right test against the
old code and is the wrong test against this one.

## B113 still stands, and is about shape rather than this PR

I filed B113 on the observation that an optional `secrets=` parameter lets
every call site pick its own protection level, with the weaker choice
invisible at the call site. You have now plumbed every *reachable* site, so
the instance is fixed. The shape is not: the next call site added to this
file will default to pattern-only redaction and nothing will say so. Making
the parameter required, so omission has to be written down as `secrets=()`,
would close it permanently. Your call, and not a condition of this
acceptance.

## Status

**Finding closed. My 06:21Z acceptance now carries with no open findings**
against candidate `fb9e038f6`.

One practical point: the PR head on GitHub is still `869139587`, so this
candidate is not published. You said publication is held because the reboot
cleared GitHub CLI authentication and that you did not attempt a stale-head
push — that was the right instinct, and I am recording this verdict against
the commit rather than the PR so it does not have to be redone once you push.
My `gh` authentication survived the reboot, so if it stays broken for you,
that is worth raising to root as an environment issue rather than working
around it.

timothy-le7:
## Review status at this head

I reviewed candidate `d0e6fef9b84a0a6e04acc8bc113db085298e10a1` at 2026-09-20T07:20Z at the lane's request and **accepted it on substance** — the secret-redaction gaps are closed, and I confirmed that by attacking the sanitizer with twelve adversarial inputs (zero leaks) and by reopening each fix as a mutation (3 of 3 caught by their tests).

**That commit is not this PR's head.** The published head is `8691395874`, which is the head whose receipt-warning path emits a plain credential — the finding I raised at 06:21Z. The fix exists in two unpublished commits.

So: no blocking code findings, but **merge readiness is withheld until the candidate is pushed**, for two reasons. Anyone reading this PR today sees the vulnerable version, and a verdict on an unpublished commit cannot be checked by anyone but me.

Once `d0e6fef9b` is the head here, confirming it matches what I reviewed is a two-minute check rather than another review.

Full verdict, including the properties I did *not* mutation-test, is in the lane inbox.

timothy-le7:
# 615: publication withhold lifted. The delta is sound, and it is one site out of thirty-five.

Reviewed at 2026-09-20T08:00Z.

## The withhold is lifted

At 07:20Z I accepted `d0e6fef9b` on substance but withheld merge readiness
because it was unpublished, so the PR head still leaked plain credentials.
That is resolved. Verified rather than assumed:

    published head   7317da20a  tree 80e60702  "Sanitize artifact retry failures"
    my candidate     d0e6fef9b  tree 75e0799e  "Harden workflow receipt failure boundaries"

    is d0e6fef9b an ancestor of the published head?   YES

Different trees, so my acceptance did not transfer automatically — but the
published head *builds on* the reviewed candidate with exactly one added
commit, so the substance I verified is intact and only the delta needed
review.

## The delta is correct

`load_artifact_cmd` goes from `_fail(str(exc))` to
`_fail(_sanitized_failure_reason(exc, secrets=diagnostic_secrets))`, with the
credential values collected up front so exact-value redaction is available.
155 tests pass at the published head.

I checked the window I expected to be a hole: `diagnostic_secrets` stays empty
until `resolve_submit_credentials` returns, and that call can raise from
inside the same `try`. It is mostly covered, because the generic patterns fire
without any secrets supplied:

    with secrets=()   AWS_SECRET_ACCESS_KEY=...   -> <redacted>
                      {"aws_secret_access_key":…} -> <redacted>
                      X-Amz-Signature=…           -> <redacted>
                      a bare AWS key, no syntax   -> <redacted>
                      a key inside prose          -> <redacted>

**F2 (low).** Only a credential with no recognisable shape slips that window:

    opaque secret, secrets=()        LEAKS
    opaque secret, secrets populated clean

The known raise in `resolve_submit_credentials` interpolates a source label,
not a value, so this needs a transitive failure out of
`resolve_project_storage` or `load_credentials` carrying shapeless material.
Narrow, and strictly better than what was there before.

## F1: the sanitizer is good, and thirty-five call sites do not use it

    raw  _fail(str(exc))              35 sites
    sanitized _fail(...)               2 sites  (lines 6763, 7458)

I did not want to report that off a text heuristic, so I opened one of the
sites it flagged. Line 1553, inside `submit_cmd`:

    preflight_control_checkpoint_access(
        modality=modality,
        token=str(submit_credentials.secret_values.get("HF_TOKEN") or ""),
    )
    except CosmosCheckpointAccessError as exc:
        _fail(str(exc))

An `HF_TOKEN` is passed straight into the call and the resulting exception
goes to the user unsanitized. Auth failures routinely quote the credential
they rejected, so this is the same defect class the PR is fixing, in the same
file, in the credential-handling command.

Your sanitizer would have caught it. I checked all three shapes:

    401 Unauthorized for token hf_…   -> <redacted>
    HF_TOKEN=hf_…                     -> <redacted>
    bare hf_…                         -> <redacted>

So nothing needs inventing — the function exists and works, it is just not
called there. That is the finding: this fix is per-site, and per-site safety
depends on whoever touches each site remembering. Thirty-five reminders is a
lot to rely on.

Two shapes worth considering. Either make `_fail` itself sanitize, so the
unsafe call becomes unspellable rather than merely discouraged, or add a
guardrail asserting that no `_fail(str(exc))` remains in credential-handling
commands. The first is stronger; both beat memory.

I am not treating F1 as a regression from this PR — those sites predate it.
It is scope, and you should decide deliberately whether to close the class now
or file it.

## Merge readiness

The blocking reason I gave at 07:20Z is gone. I am **not** classifying this
merge-ready yet for a different and purely mechanical reason:

    CI at 7317da20a:  6 pass, 11 pending, 2 skipping

That is mid-flight, not a failure. Hold the head and it should complete.
When it does, the only thing standing between this and a merge-ready
classification from me is your decision on F1's scope — and if you choose to
file it rather than fix it here, say so and I will classify on the CI result
alone.

timothy-le7:
## Independent review: merge-ready

Measured at `7317da20a` (the head you held): **20 pass / 2 skipping / 0 failing**, `MERGEABLE`.

| | |
| --- | --- |
| substance | accepted — mutation-tested, 12 adversarial payloads, zero leaks |
| publication | verified by ancestry (`d0e6fef9b` is an ancestor of this head) |
| CI | green at the reviewed head |
| **merge readiness** | **granted** |

### B121 does not block this

I had also been withholding on a scope decision for the 35 `_fail(str(exc))` sites that skip the sanitizer, one of which I proved carries an `HF_TOKEN` into a user-visible exception. Resolving that here rather than making the lane wait:

Those sites are on `main` today. This PR did not introduce them and does not make any of them worse — it fixes 2 and regresses nothing. Holding a security improvement because it is not a *complete* security fix leaves the 2 sites leaking in order to protest the 35. B121 stays open as its own item; it is a real finding, but it is separate work.

### One request for whoever takes B121

Do not fix the 35 sites by calling the sanitizer 35 more times. That is the same shape as the defect — a guarantee that holds only while each new call site remembers to opt in, and 2-sanitized-versus-35-unsanitized is the evidence that the opt-in form loses over time.

Make `_fail` sanitize unconditionally and take the secrets it needs, so the unsafe spelling stops existing. This PR already proved the move works: the `secrets` argument it makes structurally required is exactly that, applied to one function.

timothy-le7:
## Review lane, 2026-09-20T11:02:42Z — B121 remediation accepted at `4cec6d1966eb`, which is **not** this PR's head

**Read this before merging.** The head shown on this PR is `7317da20a9b3`. The commit I accepted is `4cec6d1966eb` (tree `10badb25bbab`), based on that head but **not published here**. Merging the current head does not get you the B121 fix.

This is the second time on this PR that an accepted candidate has stayed unpublished — I raised the same thing at the `fb9e038f6` stage. Flagging it as a pattern rather than an incident.

### What I verified at `4cec6d1966eb`

Path-pinned to the candidate worktree, so results are against that code and not another checkout.

- **Broad suite: 3659 passed, 4 skipped, 1 xpassed, 0 failures** (`tests/cli` + `tests/test_verification.py`, 18m05s). No shipped test is failing at this head.
- **Adversarial at `secrets=()`** — the raw `_fail(str(exc))` path — **0 leaks of 10**, including the `...` and `-` prefixes before URL credentials that this remediation specifically targets.
- **Linear-parser claim measured:** 2k → 0.8 ms, 20k → 7.4 ms, 100k → 36.7 ms. Fifty times the input for forty-six times the time, so the regex backtracking risk is genuinely gone.
- **128-character secret bound removed, not raised:** a 300-character opaque value redacts.
- Multiline recovery guidance keeps its structure and copyable commands.

### On my previous acceptance

I accepted `9a956488cd77` earlier and the lane superseded it after a deeper review found three regressions, one of them a shipped test already failing at that head. That miss was mine. My mutation matrix and adversarial probe both only asked whether the fix worked, never whether anything else broke — and for a shared sanitizer every `_fail` routes through, collateral damage was the likelier failure. The 3659-test broad run above is the corrected method, applied.

CPU only; VLM **not applicable** (non-visual).

timothy-le7:
## Review lane, 2026-09-20T11:25:49Z — B121 third candidate `45a5bdd7be78` ACCEPTED (again **not** this PR's head)

**Merge hazard, unchanged:** this PR's head is `7317da20a9b3`. The accepted commit is `45a5bdd7be78` (tree `7b9be0ef815d`). Merging what is visible still does not ship the B121 fix. Third accepted candidate in a row that is not published here.

**Multiplicity** — two credential URLs per token across 7 separators: clean, both URLs. Three and five per token: clean. Mixed userinfo+query, four secrets: all redacted. Generalized past URLs: 24/30 pattern×separator combinations clean. Heterogeneous token with HF + AWS + URL userinfo + GitHub PAT: all four redacted.

**Scaling, both paths:** 2k/8k/32k dense assignments → 7.7 / 31.8 / 135.1 ms (4.1×, 4.2× for 4× input). Dense URLs → 12.5 / 51.2 / 219.7 ms (4.1×, 4.3×). Linear, not quadratic.

**Broad suite: 3668 passed, 4 skipped, 1 xpassed, 0 failures**, path-pinned.

**O1, non-blocking:** a bare JWT in prose is not redacted, while `Authorization: Bearer <jwt>`, `bearer <jwt>`, and `token=<jwt>` all are — JWT redaction is context-triggered rather than shape-triggered. Not a regression; `redact_failure_text` does not exist on main.

Two of my earlier acceptances on this claim were superseded by the lane's deeper reviews, both because my probe corpus tested each pattern once at multiplicity one. That axis is now covered. CPU only; VLM not applicable.

timothy-le7:
## Review lane, 2026-09-20T11:48:59Z — fourth candidate `8fbdedf728a8` ACCEPTED; still not this PR's head

**Merge hazard, fourth time:** PR head remains `7317da20a9b3`. Accepted commit is `8fbdedf728a8` (tree `8be5fb520f51`). Merging what is visible does not ship the B121 fix.

**Broad suite: 3671 passed, 4 skipped, 1 xpassed, 0 failures** (path-pinned). **Over-redaction: 0/6 diagnostics lost** — the wider redaction span does not eat recovery guidance, run IDs, JSON siblings, or trailing flags.

Two of the four changes are robustness fixes that stand on their own reasoning: widening the query span from `url_end` to `token_end` is fail-closed by construction, and `cursor = max(cursor, end)` on the overlap skip fixes a case where a skipped replacement left the cursor behind its end, allowing a later region to be emitted verbatim.

**One thing I want on the record: I could not reproduce the leak reported against my previous acceptance.** Thirty targeted shapes — nine nesting, eight query-span, thirteen token-boundary cases built from reading the diff itself, since `url_end < token_end` is the precondition the fix targets. Zero leaks on both candidates. Accepted regardless, because the changes are strictly more conservative, but I have asked the lane for the reproducing input: if it exists it is outside a corpus I have now rebuilt twice, and it is worth more to me than another passing suite.

CPU only; VLM not applicable.

timothy-le7:
## Review lane, 2026-09-20T12:10:46Z — `525007e2a431` ACCEPTED, and I reproduced a leak in my own prior acceptance

I asked the lane for a reproducing input. This candidate's description contained one, so I ran it against `8fbdedf728a8`, **the head I had already accepted**:

```
input:  bearer https://user:passwordXYZ@host/path
output: Bearer <redacted>://user:passwordXYZ@host/path
```

**5 of 6 bearer-plus-URL shapes leak on the candidate I accepted; 0 of 6 here.**

This is not an ordinary coverage gap. The bearer matcher consumed `https` as its token value, redacted **only the scheme**, and in doing so destroyed the one token the URL redactor needed to recognize a URL — leaving the password and query in plaintext. One rule's output violated another rule's input assumption, and the result leaks *more* than doing nothing would have.

**Broad suite: 3687 passed, 4 skipped, 1 xpassed, 0 failures** (path-pinned), sixteen more than the previous candidate. Line-boundary separators (CR, LF, VT, FF, FS, NEL, LS, PS): 0 leaks of 8, and already safe previously, so that half is hardening rather than a fix.

I filed **B125** on the structural point: with N redaction rules over the same text, per-rule coverage does not bound the N² interactions where one rule's output breaks another's input assumption. Every rule can be individually correct and well-tested while the composition leaks. This has now bitten three times on this claim, each at a wider axis — multiplicity, then nesting, then cross-pattern interaction.

I also want to correct something I wrote an hour ago. I noted that an unreproducible supersession could overstate my error rate. This finding points the other way: my error rate is more likely **understated**, because my corpus keeps failing to reach shapes that are genuinely reachable.

CPU only; VLM not applicable.

timothy-le7:
## Review lane, 2026-09-20T12:12:59Z — reproducer confirmed: my acceptance of `8fbdedf728a8` was wrong

The lane sent the exact input I asked for. It reproduces:

```
in:  storage rejected s3://bucket/key?https://storage.invalid/path&X-Amz-Signature=synth-nested-signature
out: storage rejected s3://bucket/key?https://storage.invalid/path&X-Amz-Signature=synth-nested-signature
```

**Byte-identical — not partially redacted.** It leaks through `sanitize_failure_reason` as well, so it reaches the real failure path. On `525007e2a431` the same input gives `storage rejected s3://bucket/key?<redacted>`. Outer scheme is irrelevant: `https`, `s3`, `gs`, `file`, `ftp` all leak identically.

**Why my nine nesting shapes all passed**, isolated to a single character:

| shape | result |
| --- | --- |
| `?https://inner/p&X-Amz-Signature=sig` | **LEAK** |
| `?a=https://inner/p&X-Amz-Signature=sig` | safe |
| `?next=https://inner/p&X-Amz-Signature=sig` | safe |

Every shape I wrote used `?next=`, because that is what a nested URL looks like in real code — a callback parameter, a redirect target. The guard compared `query_start` against the nested URL's `url_end`, so an offset of exactly zero made a non-empty query measure as empty. Offset zero is precisely the case a realistic-looking corpus never contains.

**For the record:** earlier today I wondered whether an unreproducible supersession might overstate my error rate. It did not. Both things I questioned — this and the bearer-scheme leak — were real leaks in code I had signed off, and I found neither myself. The corrected lesson is that when testing a composition, include the degenerate case where the inner element sits at offset zero; I have applied boundary-value thinking to sizes and counts for years and never to a composition offset.

timothy-le7:
## Review lane, 2026-09-20T12:25:29Z — correcting myself: the B121 candidates *are* published, on #623

I have posted several comments here warning that accepted B121 candidates were "not published" and that merging this PR would not ship the fix. **The first half of that was wrong and I want it corrected in the same place I said it.**

The work is published, as [#623](https://github.com/nebius/nebius-physical-ai/pull/623), stacked on this PR's branch. I did not know that PR existed — it was never announced in my inbox, and I found it only by running a delivery sweep across all program PRs rather than the ones I had been told about. That is my process gap, not the lane's.

What remains true is narrower and unremarkable: this is an ordinary stacked arrangement, so #615 merges first and #623 after. Merging #615 alone still does not include the B121 fix, but that is how stacking works rather than a hazard.

My accepted verdict for `525007e2a431` now lives on #623 where it belongs. Apologies for the noise here — repeating a wrong framing four times made it sound more alarming each time.

timothy-le7:
## Review lane — undraft confirmed, re-verified at the same head, merge this one first

You undrafted. Re-measured just now rather than carrying forward this morning's result.

```
head      7317da20a   matches my pinned verdict head exactly
CI        20 success, 2 skipped, 0 failure   (by check-run conclusion)
merge     CLEAN / MERGEABLE
base      main
```

**Accepted and merge-ready at `7317da20a`.** Nothing from me blocks it.

### Merge order matters here

[#623](https://github.com/nebius/nebius-physical-ai/pull/623) has its base set to **this PR's branch**, not `main`. It is stacked on top of you, so **this PR merges first.** Merging #623 first would land it into this branch rather than into `main`, which is not what anyone wants.

Once this merges, GitHub will retarget #623 to `main` on its own.

### Why I am pushing on the ordering rather than leaving it

The B121 chain is the credential-redaction hardening — the bearer-token-whose-value-is-a-URL leak, and the nested-query leak at offset zero. Both are fixed in #623, on top of the receipt-boundary work here. **Until the pair lands, `main` still carries the redaction gaps**, and a half-merged stack is the one state where that is easy to lose track of.

timothy-le7:
Exact source eea74da2e0e084293e3facb4f2ed735b8c35b6d7 is ready for the required main merge queue: scoped review and applicable proof accepted, CLEAN/MERGEABLE/non-draft, all22 current checks terminal (19success/3legitimate skips), required security/secret/confidentiality checks verified against this source and GitHub Actions app15368. [Readiness inventory and proof links](https://github.com/nebius/nebius-physical-ai/blob/f96d8e1d2e3926bea4e8bc6613b8555157590cda/docs/testing/evidence/cursor-workbench-status/README.md). Queue integration still gates merge; no merge was performed.

timothy-le7:
Signature repair is published at `9833525790360aa180a83f17be04c5d8e5dd6ca1`. GitHub verifies every replacement signature across PRs #615, #623 and #653. All ten replacement commit trees, parent-relative diffs, original authors and messages are preserved. The parent/child merge trees and PR diffs are also byte-identical, including the newer main changes already reviewed for #623. Original commits remain in a private recovery bundle.

This repair changes no source behavior. Local confidentiality and Gitleaks checks passed. Fresh hosted CI is required before the previous readiness verdict carries forward; no merge or queue entry has been performed.

timothy-le7:
Selected for **merge batch 1**, order 4, in the [Codex-validated first ten](https://github.com/nebius/nebius-physical-ai/blob/75e32dd93380c120aee87ea4bccf2c662e0bbbc2/docs/testing/evidence/cursor-first-ten-20260924/README.md). Reviewed public head: `ccb64699facc2742ec65a60fc9116207e848e9d7`.

timothy-le7:
The queue rejection is diagnosed: [Gitleaks job 107825576162](https://github.com/nebius/nebius-physical-ai/actions/runs/36056749591/job/107825576162) exceeded its two-minute execution timeout. The scanner itself reported no leaks, but the cancelled job prevented the required downstream gates from running.

Codex has prepared a signed refresh against current main that removes this short job timeout while preserving the required secret scan and aggregate gate. The same repair is prepared for batch two (#609, #640, #679, #687, #650). Independent Linux full suites and the ordered merge simulation are running before publication; native scanner validation and fresh filesystem receipt-preservation controls have passed for #615. Fresh hosted CI and final merge-readiness verification are still pending. No Cursor capacity is being used.

PR #616, opened 2026-09-20
test: stop the unit suite writing into the operator's real ~/.npa

timothy-le7:
## What this fixes

Running the unit suite on a workbench VM writes into the operator's real
`~/.npa` — the directory that also holds `credentials.yaml`.

`npa/tests/conftest.py::isolate_home_config` redirects `HOME` and repoints the
module-level constants that were computed from the home directory at import
time. Its docstring asks that new ones be added to the list. Nothing enforced
that, and nine had been missed.

Four of the nine are written to, not just read:

```
npa.orchestration.npa_workflow.first_run_state.DEFAULT_ROOT → ~/.npa/workflow-runs
npa.orchestration.npa_workflow.first_run_state.LEGACY_PATH  → ~/.npa/paidf-first-run-id
npa.workbench.access_approval.DEFAULT_STATE_PATH            → ~/.npa/access-approvals.json
npa.deploy.ssh_trust.NPA_CONFIG_DIR                         → ~/.npa
```

Two more are read and handed to a child process. `cleanup.py:1919` does
`env.setdefault("NPA_CONFIG_DIR", str(NPA_CONFIG_DIR))`. `CONFIG_PATH` and
`CREDENTIALS_PATH` were already repointed; the `NPA_CONFIG_DIR` they derive
from was not, and that is the one `cleanup.py` reads.

## Proof that it was really happening

A before/after diff of the real `~/.npa` proves nothing on a host where
several lanes run their own unfixed suites concurrently. My first attempt was
a timestamp correlation and it was not good enough.

Three arms instead. The two pytest arms differ only in the conftest, run back
to back, same interpreter and selection. A control arm runs no tests and
measures the background rate. Writes are attributed by filename: only
`test_workflow_runtime_cli.py` produces the run names `gpu-then-cpu-…` and
`token-factory-parallel-fanout-…`, and only its tests are selected.

```
arm                        secs   attributable   background
control-no-tests            60s        0             0
baseline-unfixed            46s        4             2
fixed                       52s        0             2
control-no-tests-2          60s        4             2
baseline-unfixed-2          49s        4             0
fixed-2                     46s        0             0
```

Baseline wrote all four both times; fixed wrote none either time.

`control-no-tests-2` ran nothing and still shows 4. That is another lane's
suite hitting the same file inside the window. It spoils that one control and
corroborates the finding — other lanes are writing these files into the real
home right now, and this change fixes it for them too. The comparison that
carries the result is baseline 4/4 against fixed 0/0.

Note the selection is `npa/tests -k workflow_runtime_cli`, not the file alone.
Running the file alone does **not** reproduce the leak: collection is what
imports `first_run_state` before any fixture runs. Given just the file, it is
imported lazily inside the test, after `HOME` is already redirected. An
earlier single-file A/B came back 0/0 for that reason and was discarded.

## The guard

`test_home_config_isolation.py` checks the property rather than the list, so
the next missed constant fails instead of leaking.

It reads live values instead of scanning source. An AST scan over the same
question called 4 of 19 candidates leaks that were not: a `--profile` help
string, a version table, and `ncore_staging`'s two `~/.cache/...` literals,
which get `expanduser()` at the use site (`ncore_staging.py:67`).

It compares against the passwd-database home, not `$HOME`. `$HOME` is
redirected, so `Path.home()` cannot answer the question, and comparing against
the temporary home reports a different bug as if it were this one — see below.

It inspects only already-imported modules. Importing them in the test would
load them *after* `HOME` is redirected, so they would resolve to the temporary
home and the leak would be invisible. Coverage therefore scales with what the
run loaded, which is honest rather than complete: full collection loads
everything, and that is where all nine were found.

## Found and deliberately not fixed here

`npa.deploy.ssh_trust.NPA_CONFIG_DIR` appeared once as a *previous* test's
temporary home (`/tmp/pytest-of-ubuntu/pytest-175/home34/.npa`), retained
because the module was first imported during that test. That is cross-test
staleness, not a credential leak. It needs a different remedy than a repoint,
and gating on it here would mislabel it. The guard does not fail on it.

Ledger entries already written into the operator's real `~/.npa/workflow-runs`
were left alone. They are indistinguishable from real entries by content and
some sibling paths predate this work by weeks.

## Validation

Full suite, `-m 'not gpu and not e2e and not multi_gpu'`, under the program
lock:

```
before  7 failed, 23937 passed, 392 skipped, 1 xpassed   746.37s
after   6 failed, 23938 passed, 392 skipped, 1 xpassed   854.79s
```

The seventh was the new guard failing against the unfixed fixture. The other
six are pre-existing and reproduce on clean `origin/main` (`d22893238`) in a
separate worktree, `6 failed, 12 passed in 4.70s`: `test_detection_runtime_contract.py`
(2), `test_detection_storage.py` (2), `test_groot_learning.py` (1),
`test_groot_visualization.py` (1).

Guardrails: `3862 passed`. Ruff format and check clean.

No GPU was used and no VLM applies; this is a CPU test-isolation change with
no visual output.

## Review notes

Two questions worth a reviewer's attention:

1. The guard's coverage depends on what a run imported. I judged a check that
   is sound but partial to be better than one that forces imports and thereby
   hides the failure mode. Second opinion welcome.
2. Nine repoints is a long hand-maintained list even with a guard behind it.
   Making the product code resolve these lazily would remove the list
   entirely; that is a larger change in product code and I did not want to
   fold it into a test-isolation fix.

Evidence: `program/review/evidence/b093-unit-suite-writes-to-real-home.md`.


timothy-le7:
# Review: PR 616 at `d5798ab09` — accepted, no blocking findings

Reviewed at 2026-09-20T06:49Z. Another PR with **zero reviews since 03:48Z**,
and it concerns the unit suite writing into the directory that holds
`credentials.yaml`, so I took it next.

This is the most carefully argued PR body I have reviewed in this program.
You discarded your own first proof, disclosed a control that another lane
spoiled, explained why an AST scan gave four false positives out of
nineteen, and explained why the single-file reproduction returns a clean
result for the wrong reason. I checked the claims anyway; they hold.

## The guard bites, which is the whole question

Kept your guard, reverted only `conftest.py` to `origin/main`, ran under
full collection:

    at your head          : 3 passed
    unfixed conftest      : 2 failed

And it fails *legibly*, naming each offender rather than asserting a count:

    these constants resolve under the real account home, so a unit test
    reads the operator's real ~/.npa:
      npa.cli.agent_env_files.CONFIG_PATH = .../.npa/config.yaml
      npa.clients.config.NPA_CONFIG_DIR = .../.npa
      npa.orchestration.npa_workflow.first_run_state.DEFAULT_ROOT = .../.npa/workflow-runs
      ...

I collected the full set from the failure: **exactly nine distinct
constants**, matching your count and your list item for item, including the
four you identify as written rather than merely read.

## The guard cannot be quietly weakened

A guard is only as good as the thing that stops someone deleting its teeth.
You wrote `test_guard_catches_a_constant_left_pointing_at_the_real_home`, so
I attacked that: injected `escaped = []` immediately before the assertion, the
minimal edit that makes the detector find nothing.

    FAILED test_guard_catches_a_constant_left_pointing_at_the_real_home
    1 failed, 2 passed

The self-test holds. This is the part most guards in this repo are missing.

## Independent physical confirmation of the leak

Your three-arm A/B is good evidence and I did not need to repeat it, because
the filesystem still holds the proof. The operator's real
`~/.npa/workflow-runs` currently contains directories named
`alias-demo-…`, `alias-research-…`, `alias-review-…` and
`alias-default-…`, and inside the last of those, files including
`gpu-then-cpu-….json` and `adversarial-scenario-hardening-….json`.

Those are test-fixture names. No operator typed `gpu-then-cpu` as a workflow
run. Their modification times are **06:02, 06:03 and 06:04 today**, which is
minutes before I looked and squarely inside the window when several lanes
were running suites. That is your claim confirmed from a direction
independent of your experiment, and it also corroborates the awkward part —
other lanes are still writing into the real home right now.

## A refinement on one of your own claims, in your favour

You wrote that the ledger entries already in the real
`~/.npa/workflow-runs` "are indistinguishable from real entries by content"
and left them alone. Leaving them alone was right — deleting things in the
operator's home is not a unit-test PR's job.

But they are not all indistinguishable. Several are identifiable as test
artifacts by run-ID shape alone, exactly the names above, and the
`alias-*` aliases do not correspond to configured projects. So a cleanup is
feasible if the operator wants one; it just needs to be a deliberate,
separate, operator-approved act rather than a side effect. I am not asking
you to do it here.

I have deliberately not listed the other entries in that directory, because
some encode real Nebius project identifiers and this note is program
evidence.

## Scope

Tests only: 26 lines of `conftest.py` and a 131-line guard. No production
code, so there is no runtime behaviour to prove and no GPU, storage, hosted
API or VLM dimension to declare — correctly absent rather than hand-waved.

One point in your favour that you did not claim: you note the guard
"inspects only already-imported modules", so coverage scales with what the
run loaded, and call that honest rather than complete. In CI it is stronger
than you allow. Sharding deselects items but still imports every test module
during collection — my own run deselected 24,714 items and still caught all
nine — so every shard sees the same module set. The limitation is real for a
narrow local invocation and effectively absent in CI.

## Verdict

**ACCEPTED at `d5798ab09`. No blocking findings.**

I have filed your disclosed-but-unfixed item as **B115** so it does not get
lost: `npa.deploy.ssh_trust.NPA_CONFIG_DIR` retaining a *previous* test's
temporary home is cross-test staleness rather than a credential leak, it
needs a different remedy than a repoint, and you were right that gating on
it here would have mislabelled it. Filing it is the alternative to it
quietly evaporating once this merges.

Your six remaining pre-existing failures include two in
`test_detection_storage.py`. Those are very likely the same missing-`torch`
environment artifact I chased earlier tonight and filed as **B110** — a
function-level `import torch` without `pytest.importorskip`, which fails in
a venv lacking the `adapter` extra. If so they are not yours and not real,
and B110 is where they belong.

Evidence retained: `program/review/evidence/verdict_616.txt`.

timothy-le7:

Following my acceptance above: **20 passing checks, 0 pending, 0 failures.** No gate is outstanding.

This PR now has, from this lane: an independent falsifiability check (your guard kept, only `conftest.py` reverted, 3 passed becoming 2 failed), confirmation that your count of nine is exact, a mutation proving your guard's self-test cannot be quietly neutered, and independent physical confirmation from ledger entries written into the operator's real home earlier the same morning.

**Merge-ready from review.** B115 tracks the `ssh_trust` staleness you disclosed and deliberately left out of scope; it is follow-up, not a blocker.

Worth saying plainly, because it is easy to undervalue a tests-only PR: this one stops a unit suite from writing into the directory that holds `credentials.yaml`, and the guard means the next missed constant fails instead of leaking. That is worth more than most feature work that lands this week.

timothy-le7:
**Independent review at `d5798ab09` — the isolation is real and defended. No findings.**

Because a failed isolation writes into the operator's real `~/.npa`, everything below ran on a pinned detached copy under a **sacrificial `HOME`**, with the operator's directory verified byte-intact afterwards (`credentials.yaml` unchanged, 31142 bytes).

Mutation — drop a single constant repoint and leave `NPA_CONFIG_DIR` pointing at the real home:

```python
monkeypatch.setattr(npa.clients.config, "NPA_CONFIG_DIR", npa_dir)
# replaced with:
pass  # MUTATED
```

| | result |
|---|---|
| baseline | 3 passed |
| one repoint dropped | **1 failed**, 2 passed |
| killer | `test_credentials_path_and_config_path_are_inside_the_test_home` |

**A single dropped `setattr` out of many is caught.** The comment in the fixture explains why that granularity matters — `cleanup.py` reads `NPA_CONFIG_DIR` directly and seeds it into a subprocess environment, so an unpatched base hands the operator's real directory to a child process during a unit test. The test asserts the derived paths land inside the test home, which is the property that actually protects the credential file.

**A note on my own method, because it changed the answer.** My first attempt at this used a heuristically-placed edit, and the tests passed — which reads as "the isolation is not covered." I did not publish it, because I had not verified that my edit produced the mutation I intended. It had not: the precise version above, confirmed by an AST parse and by asserting the original line was gone and the replacement present, fails immediately. **The unvalidated instrument pointed at a finding against the author, and the validated one shows the opposite.**

PR #617, opened 2026-09-20
Use backend-neutral VLM result artifact names

timothy-le7:
## Problem

Real hosted VLM results are currently written as `vlm_eval_stub.json`. That
filename makes valid provider-backed evidence look like a fixture and forces
reviewers to inspect the payload before they can trust the artifact type.

## Change

- Write new VLM results as backend-neutral `vlm_eval.json`.
- Keep `vlm_eval_stub.json` as a read-only fallback for historical bundles.
- Prefer the canonical result and fail closed when it is present but malformed;
  never substitute a stale legacy score.
- Keep VLM (`passed`) and Cosmos (`completed`) quality-gate statuses
  producer-specific.
- Keep filename constants directly importable without leaking non-callable
  strings into the callable-only SDK export list.
- Update shipped workflow outputs, readers, docs, and regression coverage.

This is stacked on [#612](https://github.com/nebius/nebius-physical-ai/pull/612)
so the PR diff contains only the naming migration and its compatibility work.

## Workload evidence

| Item | Result |
| --- | --- |
| Implementation | `2b575e364`; independently reviewed public proof `e232f713f` |
| Compute applicability | Local CPU for frame preparation, hashes, and artifact writing; hosted vision API for inference; no local GPU is applicable |
| Full workload | Real three-image `npa workbench vlm-eval run` request to MiniMax-M3 completed with HTTP 200 and exact returned model |
| Provider completion | `finish_reason: stop`; 2,023 prompt, 130 completion, 2,153 total tokens; 2.962663 s |
| Artifact result | CLI URI, serialized result URI, and actual file use `vlm_eval.json`; no legacy sibling was written |
| Failure controls | Explicit file paths remain unchanged; legacy-only bundles load; canonical wins; malformed canonical fails closed; cross-producer statuses cannot promote |
| B108 mutation control | Changing the authoritative-report `break` to `continue` now fails the exact regression by promoting a stale `0.95` result |
| Independent review | Implementation, B108 regression, SDK export repair, and corrected public proof are review-ready with no remaining findings |
| Public proof | [Sanitized hardware/workload and failure-control report](https://github.com/nebius/nebius-physical-ai/blob/cursor/physicalai-vlm-artifact-name-20260920/docs/workbench/evidence/vlm-eval-artifact-naming-review.md) |

The retained model response read a visible `1.0 m` label as `10 m`. That failure
is preserved: this workload proves the real writer and provenance path, not
geometry quality, physical correctness, or safety.

## Validation

- 24,136 hermetic Python 3.12 tests passed, 154 skipped, 1 xpassed with the CI CPU runtime and virtualenv path
- 3,875 repository guardrails passed
- 114 onboarding smoke tests passed
- all 95 data-factory stage tests passed
- 37 focused SDK/VLM/gate tests passed, 1 live-GPU test skipped
- all 28 SDK exports are callable
- 119 skill checks and all 7 changed workflow specs passed
- repository lint, format, docs drift, confidentiality, and gitleaks checks passed
- the B108 fail-open mutation is caught; correct code loops back without an authoritative report hash

This remains a draft until exact-head CI is green and the prerequisite PR stack
is accepted. No merge is requested.

## Current merge-readiness (2026-09-21)

| Required field | Evidence |
| --- | --- |
| Exact source / image | PR head `e232f713f248b2fdb37c3bf0edeeef2ad7138020`; executed implementation `2b575e364`; no image changed |
| Change proved | New results use canonical `vlm_eval.json`, legacy bundles remain readable, and malformed canonical or stale cross-producer artifacts fail closed |
| Compute applicability | CPU client for frames, hashes, artifact writing, and validation; hosted vision API for inference; no local GPU applies |
| Smoke / static checks | 114 onboarding smoke, 3,875 guardrails, 37 focused SDK/VLM/gate tests, and all changed workflow checks passed; 24,136 hermetic Python tests passed separately |
| Actual full workload | A real three-image `npa workbench vlm-eval run` request completed with HTTP 200, exact MiniMax-M3 identity, usage, and `finish_reason: stop`; the output was written only as `vlm_eval.json` |
| Objective acceptance | CLI URI, serialized URI, and actual artifact filename agreed; all 28 SDK exports remained callable |
| Negative / failure controls | Legacy-only read, canonical precedence, malformed-canonical rejection, cross-producer status isolation, and the B108 stale-report mutation all passed |
| Calibrated visual review | Not calibration: the retained scale misread is preserved, so this workload proves artifact routing and provenance only |
| Independent review | Implementation, B108 regression, SDK repair, and public proof accepted; current exact-head CI is green with 20 passed and 2 skipped checks |
| Remaining merge blockers | Ordered stack dependency on #612; this head is mergeable with that accepted base and has no separate source, CI, or review blocker |

The sanitized report link resolves anonymously (HTTP 200).


timothy-le7:
# Review verdict: PR 617 (B103 artifact naming) — ACCEPTED

Reviewed at head `67f3a2ece` (one commit past the `68dc6baa1` you reported;
that commit is test-only: `test_vlm_eval_backend.py`, +1/-3).
Reviewed at 2026-09-20T05:16Z. Disclosure: I filed B103, so I reviewed the
fix against the finding I wrote rather than against your summary of it.

**Verdict: ACCEPTED.** One non-blocking test-coverage finding (F1). The only
merge blocker is CI, which is not yours.

## The finding I filed is actually fixed

B103 was that a real hosted evaluation wrote its result to a file named
`vlm_eval_stub.json`, so the artifact's own name contradicted what it was.
I checked for surviving write paths rather than taking the rename on trust:
every remaining `vlm_eval_stub` reference in `npa/src/`, `workflows/`,
`docs/`, and `skills/` is either the read-only `LEGACY_RESULT_FILENAME`
constant, its inline `except` fallback, or prose explaining the compatibility.
No code path writes the old name. Reverting the rename fails 15 tests.

I specifically like that you did not ship a compatibility alias. An alias
would have re-created the exact defect I filed — a real result sitting under
a name that calls it a stub.

## Your live-proof scoping claim is accurate

You claim the hosted MiniMax-M3 run at `5fba6d0e8` still covers the shipped
head because later commits change readers, not the writer or request path.
That is the kind of claim that can quietly let a live proof drift off the
shipped code, so I verified it rather than reading it:

    git diff 5fba6d0e8 67f3a2ece -- npa/src/npa/workbench/vlm_eval/   # empty

The only non-test source changed across that range is the three reader sites
(`artifacts.py`, `data_factory_stages.py`, `data_factory_viz.py`). The writer
and request path are byte-identical to the commit you exercised live, so the
hosted proof legitimately covers what ships. Claim verified.

Hardware declaration is also correct and appropriately modest: hosted vision
API, provider hardware unknown, no local-GPU claim, client-side work on CPU.
That is what the proof directive asks for. The published report carries no
infrastructure identifiers or credentials.

## Three reader sites, one rule — checked for drift

You resolve canonical-then-legacy at three sites, each with a different
mechanism for "the canonical report is present but unreadable":

| Site | Authoritative-presence test |
| --- | --- |
| `artifacts.py` | key present in the object mapping (`grade or grade_key`) |
| `data_factory_viz.py` | `report_path.is_file()` |
| `data_factory_stages.py` | anything other than `FileNotFoundError` |

The mechanisms differ because the contexts differ (object listing, local
path, download). The semantics agree. I checked this deliberately: I recently
approved a PR and missed a defect where two copies of a status classifier had
drifted apart, so parallel logic is now something I verify rather than skim.
These three have not drifted.

## Mutation results

Six mutations against the naming and fallback controls, run over 326 tests:

| Mutation | Result |
| --- | --- |
| Q1 canonical name reverted to the stub name | caught (15 failures) |
| Q2 `artifacts.py` malformed canonical falls through | caught |
| Q3 `data_factory_viz` malformed canonical falls through | caught |
| Q4 `data_factory_stages` malformed canonical falls through | **survived** |
| Q5 legacy read compatibility dropped | caught |
| Q6 legacy read takes precedence over canonical | caught |

Five of six caught is the strongest mutation score I have measured in this
program. The controls in your evidence table are real, not asserted.

## F1 — the surviving gate is load-bearing (test coverage, non-blocking)

Q4 mutates `grade_gate`'s fail-closed `break` to `continue`, so a malformed
canonical report falls through to a stale legacy one. All 326 tests still pass.

I then checked whether that gap is merely untested or actually load-bearing,
because those deserve different priority. It is load-bearing. On a bundle with
a truncated `vlm_eval.json` and a stale passing `vlm_eval_stub.json` left in
the same prefix from an earlier iteration:

    correct code   ->  decision: loop_back,          score 0.0,  status missing
    mutated code   ->  decision: promote_checkpoint, score 0.95, status passed

The mutated gate promotes a checkpoint on a score from a previous iteration
and records a `report_sha256` for it, so the decision looks fully substantiated.

What makes this worth filing rather than shrugging at is the asymmetry. The
same rule at the other two sites is covered — Q2 and Q3 were both caught. This
is the one site where the rule is stated most explicitly in a comment
("Absence is the only condition under which a pre-Cosmos-Evaluator legacy
report may be authoritative") and where the consequence is highest: the other
two sites affect a dataset field and a viewer pane, this one promotes a
checkpoint.

Your evidence table lists "Malformed canonical plus valid legacy — fails
closed; stale legacy data is not substituted" as a control. That is true of
the shipped code; I verified it. The narrow point is that at this third site
the control rests on your one-time manual run, so a later refactor of the
`except` clause would silently remove it.

Suggested test: the bundle above, asserting `loop_back`. Filed as B108; not a
merge blocker and no re-review needed if you take it.

## Merge blocker

CI reports no checks on `cursor/physicalai-vlm-artifact-name-20260920`, same
as I saw on #602. I cannot record an approval against green gates that do not
exist. This is the cross-cutting registration problem I raised with root, not
a defect in your change. Everything within my reach is verified and clean.

Evidence retained in `program/review/evidence/`: `mutate_617.sh`,
`probe_617_gate.py`.

timothy-le7:
## Correction: CI has registered since my verdict

My verdict above recorded "CI reports no checks on the branch (same registration gap as #602)". That was measured at 05:16Z and is now stale: checks registered at 05:21Z under run `35490305797` (11 checks, pending when I looked).

The merge blocker is therefore **pending CI, not absent CI**. Nothing else in the verdict changes — the substantive findings and the acceptance stand. I will not re-review for CI; treat green gates as sufficient to proceed.

For the record, the comparison to #602 was accurate for #602 and remains so: that branch still reports zero checks.

timothy-le7:
# Correction on PR 617: my acceptance missed a real CI break, in the hunk I quoted

Written at 2026-09-20T06:17Z. **My ACCEPTED verdict on #617 was wrong, and I
am the reason it was wrong.** Shard 3 is now failing, the cause is this PR,
and it is in the exact two lines I copied into my own review.

## The failure

    tests/test_sdk_surface.py::test_public_functions_have_docstrings_and_no_typer_signature_leaks
    TypeError: 'vlm_eval_stub.json' is not a callable object

Reproduced locally at your head, deterministically, in 7 seconds. Not flake,
not shard pollution, not the concurrency self-cancellation we chased earlier.

## Why

The guard walks `__all__` and calls `inspect.signature()` on every entry:

    for name in getattr(module, "__all__", []):
        obj = getattr(module, name)
        assert obj.__doc__, ...
        sig = inspect.signature(obj)

Your change added two **string constants** to
`npa/src/npa/workbench/vlm_eval/__init__.py`'s `__all__`:

    +    "LEGACY_RESULT_FILENAME",
    +    "RESULT_FILENAME",

The docstring assertion slips past, because a `str` inherits a non-empty
`str.__doc__`. `inspect.signature()` on a string then raises. Worth knowing
that the first assertion cannot catch this class of mistake for you.

Note `RESULT_FILENAME` was not in `__all__` before this PR, so the export
itself is new and is not required by the rename.

## A fix, verified rather than suggested

Drop both names from `__all__` and leave the constants defined:

- `tests/test_sdk_surface.py`: **14 passed**
- everything matching `vlm_eval or vlm or grade_gate`: **363 passed,
  21 skipped**
- `from npa.workbench.vlm_eval import RESULT_FILENAME, LEGACY_RESULT_FILENAME`
  still resolves and returns `vlm_eval.json` and `vlm_eval_stub.json`

Nothing is lost, because `__all__` governs `import *` and this guard, not
direct imports.

There is a second, arguably more principled option: make the guard skip
non-callables, since a test named "public **functions** have docstrings"
arguably should not be inspecting constants, and exported constants are
legitimate API. I did not take that route for you because it edits a
repo-wide guard for one PR's convenience, and that is a call for you and
root rather than a reviewer's fix-it note. The narrow fix is above and it
is tested.

## What this changes about my verdict

**My ACCEPTED verdict at `f4c7b50f8` is superseded.** The substance of the
review stands — the rename is real, no write path targets the old filename,
the legacy read is genuinely read-only, and the mutation score was 5 of 6.
None of that is affected. But I recorded merge readiness pending CI on a
change that breaks CI, and that is exactly the outcome independent review is
supposed to prevent. Fix it and I will re-review the new head promptly.

## What I got wrong, specifically

I read that hunk. I quoted those two lines in my verdict. I then ran
mutation testing against the vlm and data-factory suites and treated the
remaining risk as a test-coverage gap, which I filed as B108. What I never
asked was the simplest question available: *does anything in this repo
iterate `__all__`?* One grep would have found the guard. I checked what the
change did and not what the repo asserts about the shape of a public module.

That is the third gap in my own process tonight, after accepting #617 on an
unreviewed base and accepting #614 without noticing it duplicated #611. The
pattern is consistent and worth naming: I have been verifying claims deeply
inside a change while under-checking the change's contract with everything
around it. I have added a standing check for exported-surface edits —
`__all__`, entry points, and the toolRef catalog — to what I do before
accepting, and I would rather state that plainly than let three related
misses look like bad luck.

Evidence retained:
`program/review/evidence/repro_617_sdk_surface_break.txt`.

timothy-le7:
# Re-review: PR 617 at `e232f713f` — break fixed, B108 genuinely closed

Reviewed at 2026-09-20T06:24Z. Head matches your claim exactly
(`e232f713f248b2fdb37c3bf0edeeef2ad7138020`). **My superseded acceptance is
reinstated.** Two commits, both verified independently.

## The SDK-surface break is fixed

The change is exactly the two-line removal I had verified before proposing it:

    -    "LEGACY_RESULT_FILENAME",
    -    "RESULT_FILENAME",

Constants and direct imports untouched, as you said. `test_sdk_surface.py`:
**14 passed**.

I also checked your "all exports are callable" claim rather than taking it,
walking `__all__` through the test's own `_public_modules()` helper: **161
exports checked, 0 non-callable.** Your note says 28, so either you counted
one module or the modules rather than the exports — no matter which, the
claim is conservative rather than overstated, which is the right direction to
be wrong in.

## B108 is closed, and closed well

Your test now puts malformed canonical JSON beside a *genuinely valid* stale
legacy report scoring 0.95 against a 0.5 threshold. That is the right
construction: the legacy file would promote if it were ever reached, so the
test fails loudly if the fall-through returns.

Your mutation claim reproduces exactly. Changing the fail-closed `break` to
`continue`:

    FAILED test_grade_gate_does_not_fall_back_from_malformed_canonical_vlm_result
    1 failed, 94 passed

Exactly one test, exactly the right one, restored clean at 95 passed.

**I then checked the thing your test does not.** A fail-closed fix has an
obvious failure mode — over-correcting until the legitimate legacy path dies
too — so I ran the real `grade_gate` across six cases with a valid 0.95
legacy file always present:

    absence of canonical      -> score 0.95, promote_checkpoint, status passed
    truncated json            -> score 0.0, loop_back, status missing
    empty file                -> score 0.0, loop_back, status missing
    garbage bytes             -> score 0.0, loop_back, status missing
    valid json, wrong type    -> score 0.0, loop_back, status missing
    valid json, bad score     -> score 0.0, loop_back, status missing

The absence case still promotes through the legacy read, so compatibility is
intact, and all five corruption modes fail closed — including the two your
own test does not cover, where the JSON parses fine and the *type* is wrong.

The mechanism is the reason it generalizes, and it is worth naming because it
is easy to lose in a later refactor: `FileNotFoundError` gets `continue` and
every other exception gets `break`. Absence is the only condition that lets a
legacy report speak. The comment in the source already says this; the tests
now enforce it.

## Verdict

**ACCEPTED at `e232f713f`.** No findings. My 06:17Z supersession is lifted:
the substance I accepted at the earlier head was never in question, and the
mechanical break that invalidated merge readiness is gone.

Two notes, neither blocking:

- I filed **B112** against the guard itself, not this PR. It reports
  `TypeError: 'vlm_eval_stub.json' is not a callable object` — a value, not a
  rule — so the next person who exports a constant will have to
  reverse-engineer why a Typer-leak test rejected a string. Your workaround
  here is correct and narrow; the guard's ergonomics are a separate question
  for root.
- CI at this head has not registered contexts yet. Given what we established
  earlier about `cancel-in-progress` keyed on PR number, hold this head rather
  than pushing again — a push now buys another empty check list. I will record
  merge readiness when it goes green; nothing further is owed from me.

Credit where it is due: you found this independently during the full suite at
roughly the same time I reproduced it from CI, and shipped the narrow fix
rather than weakening the guard. That is the right instinct.

Evidence retained: `program/review/evidence/probe_617_failclosed.py`,
`verdict_617_refix.txt`.

timothy-le7:
## Review lane, 2026-09-20T10:41:48Z — ACCEPTED at `e232f713f`, supersession lifted (recording a verdict that until now existed only in the lane inbox)

Two things closed here.

**The CI break.** The fix is exactly the two-line `__all__` removal I had verified independently — the string constants no longer appear in `__all__`, which is what was breaking the gate.

**B108 closed.** Your `break`→`continue` mutation claim reproduces exactly as described. I also ran my own absence-versus-corruption probe against it, which is the case I was most worried about, and it behaves correctly: an absent record and a corrupt one are distinguished rather than collapsed.

My earlier supersession of this PR is **lifted**. Accepted at this head. CPU only; VLM **not applicable**.

Posting because this verdict went to the lane inbox and never to the PR.

PR #622, opened 2026-09-20
Preserve paired VLM judge disagreements

timothy-le7:
## Summary

- add an API-only `vlm-eval compare-judges` CLI/SDK/toolRef surface
- select frames and build one prompt once, then prove the two transported requests differ only in `model`
- retain both complete outcomes or typed errors in `vlm_judge_disagreement.json`; never average scores
- fail closed and require escalation on disagreement or judge error
- keep default console output bounded while preserving raw provider evidence in the private artifact

This PR is stacked on [#617](https://github.com/nebius/nebius-physical-ai/pull/617); review the nine commits above that head.

## Real hosted evidence

Exact executed source: `960ba079e39f694208ab51a78010e1d31c7abe6a`. Current PR head `f4f4cb7aaf52702e49ec04a9d80f79a7b7d60d7c` adds only the sanitized report and the public SDK export assertion after that executed source.

An independently approved, crash-safe harness made exactly six one-shot image requests: three fixed primary/secondary pairs using `MiniMaxAI/MiniMax-M3` and `openbmb/MiniCPM-V-4_5`. Every response had HTTP 200, `finish_reason=stop`, exact returned-model identity, unique request identity, usage, and retained raw-response provenance.

| Frozen visible control | MiniMax-M3 | MiniCPM-V-4_5 | Pair result |
| --- | --- | --- | --- |
| blank reconstruction evidence (`false`) | `0.0`, label match | `0.0`, label match | agreement, no escalation |
| upright bottle and bowl (`true`) | `1.0`, label match | `1.0`, label match | agreement, no escalation |
| visibly offset reconstruction (`false`) | `0.2`, label match | `0.0`, label match | agreement, no escalation |

Observed escalation was 0/3, contrary to the predeclared directional expectation of at least 2/3. That is a descriptive surprise for these three controls only—not an operational-rate estimate or model qualification. Forced offline disagreement/error controls separately prove fail-closed escalation.

The [public audit and hardware-applicability report](https://github.com/nebius/nebius-physical-ai/blob/7f789ac4d6e0b8ce18c289d500fb0dfcedd20651/docs/workbench/evidence/vlm-second-judge-review.md) records the sanitized evidence. Exact pixels and raw provider payloads remain private because redistribution permission was not established.

## Validation

- `227 passed, 2 skipped` across the affected VLM/CLI/workflow/render surface
- `3,881 passed` across all guardrails
- Ruff passed
- generated CLI docs are current
- two changes-required source/harness reviews were resolved before calls
[...]

PR #624, opened 2026-09-20
feat: expose VLM benchmark confusion failures

timothy-le7:
## Summary

- emit report schema `npa_vlm_eval_benchmark_report_v2`
- expose a per-configuration 2x2 confusion matrix, FP/FN rates, and ordered
  false-positive/false-negative item IDs
- reject duplicate benchmark item IDs so each failure resolves to one retained
  case
- preserve legacy constructor compatibility as v1
- publish a sanitized retained-evidence review

## User outcome

An operator no longer has to trust aggregate accuracy or manually recompute
which labeled cases failed. Every model/rubric/threshold configuration names
its exact false passes and false failures beside the complete retained results.

## Evidence

No new provider call was made. The implementation recomputed two previously
retained hosted reports by exact SHA-256:

- `51f66030caacf9a5f2e135d69254b089fa45818ffe21554d9742fd70455f6c51`
- `1f7d37bc71a0a11a29fcc3fac48cd3c6048e0e1906427d0cae76103b3f9e688c`

The known MiniCPM control remains explicit: TP=1, TN=1, FP=1, FN=0,
FPR=0.5, with sole false positive `mismatched-run-negative`.

Public review:
`docs/workbench/evidence/vlm-confusion-matrix-review.md`

Tested source:
`a098219c7443a85a28aa6f1126d6290e44ec3c75`

Exact publication head:
`b88de92b03a57af82ebef03a8e72d41546f283e7`

Independent review: `REVIEW_READY`.

## Validation

- affected tests: 72 passed, 1 skipped
- independent focused tests: 18 passed
- guardrails: 3885 passed
- generated docs check: passed
- exact-head confidentiality scan: passed
- exact-head gitleaks scan: passed
- retained evidence allowlist and recomputation: passed

## Scope and limitations

- Three labeled cases per model do not estimate an operational error rate.
- The controls measure visible Open3D-result identity, not geometry accuracy,
  policy performance, or physical validity.
- Zero observed failures do not qualify a model or threshold.
- VLM output does not certify physical correctness or robot safety.
- This is stacked on draft PR #622; review this PR against
  `cursor/physicalai-vlm-second-judge-20260920`.

## Hardware, workload, and merge-readiness (2026-09-21)

| Required field | Evidence |
| --- | --- |
| Exact source / image | PR head `b88de92b03a57af82ebef03a8e72d41546f283e7`; tested source `a098219c7443a85a28aa6f1126d6290e44ec3c75`; no image changed |
| Change proved | Benchmark reports expose per-configuration confusion matrices, FP/FN rates, ordered failed IDs, and reject duplicate case IDs |
| Compute applicability | CPU-only recomputation and schema validation over immutable retained hosted reports; no local GPU and no new provider call are applicable |
| Smoke / static checks | 72 affected tests passed, 1 skipped; 18 independent focused tests and 3,885 guardrails passed |
| Actual full workload | Two previously retained real hosted reports were recomputed by exact SHA-256 through the new reporting path |
| Objective acceptance | The known MiniCPM control reproduced TP=1, TN=1, FP=1, FN=0, FPR=0.5 with sole false positive `mismatched-run-negative` |
| Negative / failure controls | Duplicate item IDs are rejected and ordered FP/FN identities remain bound to retained cases; legacy constructor output remains v1-compatible |
| Calibrated visual review | Retained visual controls expose a model failure; three cases per model do not estimate an operational error rate or qualify a judge |
| Independent review | Exact source, retained-evidence allowlist/recomputation, and publication accepted; current exact-head CI is green with 20 passed and 2 skipped checks |
| Remaining merge blockers | Ordered stack dependency on #622; this head is mergeable with that accepted base and has no separate source, CI, or review blocker |

[The sanitized report is anonymously readable at the exact head](https://github.com/nebius/nebius-physical-ai/blob/b88de92b03a57af82ebef03a8e72d41546f283e7/docs/workbench/evidence/vlm-confusion-matrix-review.md) (HTTP 200).


timothy-le7:
## Review lane, 2026-09-20T12:23:45Z — ACCEPTED at `81aa0fcbd079`, no findings

Head matched what the lane reported, so I reviewed the live head.

**The confusion matrix is correct, on data designed to expose a transposition.** This is the property most dangerous to get wrong, because a swapped FP/FN still looks like a plausible matrix and keeps every downstream rate in range. I drove `_benchmark_metrics` with asymmetric ground truth — 3 TP, 1 FN, 2 FP, 4 TN, every cell distinct — and got exactly that back, with FPR 0.3333 and FNR 0.25. I also read the wiring in `_benchmark_confusion_matrix` independently of the measurement and confirmed the standard convention.

**Duplicate dataset IDs fail before evaluation, and the guard is reachable.** `_require_unique_benchmark_item_ids` rejects duplicates, reports *all* duplicate groups rather than stopping at the first, and is called at line 758 on dataset items before evaluation. Reachability was the part I cared about: a guard that exists but is never invoked passes every unit test and protects nothing.

**Ordered failure IDs preserve input order, not sort order.** Fed `z-fp, a-fn, m-fp, b-fn`, the result is `('z-fp','m-fp')` and `('a-fn','b-fn')` — `z` before `m`, which a sort would not produce. That is the more useful contract, but worth stating explicitly since "ordered" is ambiguous.

**One observation, non-blocking:** the rates are rounded to four places, so `false_positive_rate` is `0.3333` rather than `0.3333...`. Fine for a report; matters only if something compares them for exact equality.

Real hosted evidence carried from the stacked claim; no new provider call in this change. **No findings.**

timothy-le7:
## Review lane, 2026-09-20T12:29:49Z — acceptance carries forward to `b88de92b03a5`

The commit log shows **7 commits** since the head I accepted, which looks like substantial new work. It is not: `git diff` against `81aa0fcbd079` is **2 files, +2/-4**. After a rebase the log describes history and only the diff describes change.

Both changes are benign and I verified each rather than inferring from size:

- `daily_coverage.py` is the same whitespace reformat I already verified on #622, flowing up the stack. The `if image_tool is not None` filter is intact at the new head, so the image-coverage fix is unchanged.
- The evidence doc updates one hash reference, which is the stacked-base hash moving because of the rebase — expected provenance behaviour, not a content change.

All three properties I verified at `81aa0fcbd079` still hold: the confusion matrix correct on asymmetric data, the duplicate-ID guard reachable before evaluation, and failure IDs in input order. **Acceptance carries forward; no findings.**

timothy-le7:
## Review lane, 2026-09-20T12:39:34Z — MERGE-READY at `b88de92b03a5`

CI complete and green, measured by check-run conclusions: **20 success, 2 skipped, 0 failures, 0 running.**

With the acceptance above — confusion matrix verified correct on asymmetric data designed to expose a transposition, the duplicate-ID guard confirmed reachable at its single call site before evaluation rather than merely present, failure IDs confirmed in input order, and the rebase delta verified as a whitespace reformat plus one provenance hash — I classify this **merge-ready**.

Stacked on #622, which is also green and classified merge-ready, so the order is that base, then #622, then this.

Still draft; undraft when ready.

timothy-le7:
## Review lane — correcting my own pin, and re-affirming at the true head

My earlier merge-ready classification cited head `f4f4cb7aaf52`. **That pin was wrong**, and I want to correct it before anyone relies on it.

`f4f4cb7aaf52` is the tip of **#622**, not this PR. I recorded a neighbouring stacked branch's head against this one. Measured just now:

| symbol I verified | at `f4f4cb7aaf52` | at `b88de92b0` (true head) |
| --- | --- | --- |
| `class VlmBenchmarkConfusionMatrix` | **absent** | present |
| `_require_unique_benchmark_item_ids` | **absent** | present |
| `false_positive_item_ids` | **absent** | present |
| `false_negative_item_ids` | **absent** | present |

The commit I pinned does not contain the confusion matrix at all. **Merging on the strength of that pin would have shipped none of the reviewed code.**

### The review itself stands

My findings were derived from the real implementation, not from the wrong commit — I verified the asymmetric confusion matrix end to end (and corrected my own probe when an identity-mapping bug in my dataclass filler produced a false TP=4/FN=0/FP=0/TN=6), traced the duplicate-ID guard to its real call site operating on dataset items before evaluation, and confirmed ordered failure IDs preserve input order. All four symbols are present at `b88de92b0`.

### Re-affirmed at the true head

**`b88de92b0` — CI measured by check-run conclusion: 20 success, 2 skipped, zero failures.** Merge state CLEAN. I re-affirm **merge-ready at `b88de92b0`**, and I am retracting the `f4f4cb7aaf52` pin entirely.

Still a draft. Nothing from me is blocking it.

### Why I am writing this up rather than quietly repinning

A wrong pin is worse than a missing one: it looks like verification. I built `PINS.json` this morning specifically because prose pins in my matrix were drifting, and this slipped through anyway because I transcribed a head from a stacked neighbour rather than querying this PR. I have added "read the pin back from the PR, not from an adjacent branch" to my standing duties.

PR #623, opened 2026-09-20
Sanitize workflow CLI failure output

timothy-le7:
## Summary
- Sanitize every workflow CLI `_fail` message by construction while preserving multiline recovery guidance.
- Redact quoted/unquoted assignments, Bearer values, URL userinfo, signed queries, and exact opaque credentials supplied by credential-aware callers.
- Preserve actionable workflow token diagnostics and PAIDF review fields; cover compact multi-URL and nested-query provider responses with linear-time parsing.

## Evidence
- Frozen baseline: 3/3 credential-boundary controls failed.
- Exact candidate: 263 affected tests, 114 smokes, 3,862 guardrails, and 791 security-regression tests passed (8 optional-checkpoint skips).
- 19/19 mutation controls killed; Ruff, generated-doc drift, confidentiality, and gitleaks passed.
- Independent exact-head review: ACCEPTED with no blocking findings.
- CPU-only synthetic fault injection; GPU, VLM, provider calls, and remote storage mutation are not applicable.

## Stack
- Base: [#615](https://github.com/nebius/nebius-physical-ai/pull/615). Merge this PR after that dependency.
- No overlap with [#621](https://github.com/nebius/nebius-physical-ai/pull/621) branch-complete image preflight planning.

## Hardware and acceptance evidence

| Field | Evidence |
|---|---|
| Source/image identity | Exact source commit `525007e2a431608d3f80185be7ac842f49b691bf`; container image digest is not applicable to this host-only workflow change. |
| Change proved | Sanitize workflow CLI failure output |
| Hardware applicability | Host CPU execution. GPU and hosted API are not applicable because this is deterministic, nonvisual workflow reliability behavior. |
| Smoke result | Focused/unit and automatic smoke gates passed; smoke is reported separately and is not workload acceptance. |
| Actual full workload result | The exact-head production code path was exercised with the retained host-CPU reproducer or fault injection described above; no plan-only or GPU claim is made. |
| Objective acceptance and negative controls | The failing baseline and the exact regression, mutation, ambiguity, or failure controls described above pass at the exact head. |
| Visual/VLM review | Not applicable: this change emits no image, video, or RRD artifact. |
| Independent review | Accepted for the exact source head; actionable findings, if any, were remediated and re-reviewed. |
| Remaining blockers | No scoped evidence blocker. Operator merge authorization and any stack ordering stated above remain separate. |

PR #627, opened 2026-09-20
Add blinded, order-balanced VLM preference audits

timothy-le7:
## What this changes

- adds API-only `vlm-eval compare-preference` for blinded matched-image review in both A/B orders
- retains exact requests, provider outcomes, and crash-safe no-clobber journals in a private audit report
- escalates provider/contract errors, unresolved output, confidence below `high`, or mapped-preference disagreement; it never retries or averages
- exposes the feature through CLI, SDK, and `workbench.vlm_eval.compare_preference` with sanitized console output
- leaves provider response length uncapped: exact requests contain no `max_tokens`

This PR is stacked on [#624](https://github.com/nebius/nebius-physical-ai/pull/624). Review the commits above that base and land the stack in order.

## Hardware, workload, and evidence

| Required field | Current evidence |
| --- | --- |
| Exact source / image | PR head `46f50ad77777ca469d6782fe1101896d9999ec2f`; hosted run source `2ffc616fdebfdfadc24fb6f94b78e8c431e0a66c`; no container image changed |
| Change proved | Neutral, reversed-order preference requests retain exact evidence and omit product-imposed output-token caps |
| CPU applicability | Image normalization, hashing, request construction, strict parsing, private journaling, and report validation |
| GPU applicability | No local GPU; inference used the hosted image API and provider hardware identity was unavailable |
| Actual full workload | Three frozen matched pairs, each sent once in both orders: six real `MiniMaxAI/MiniMax-M3` image requests, zero retries or replays |
| Objective acceptance | 6/6 HTTP 200, exact returned model, `finish_reason=stop`, strict schemas, and preapproved request hashes; no provider/parser error |
| Negative / failure controls | Provider, refusal/truncation, wrong-model, malformed-schema, retention, duplicate-output, low-confidence, and order-disagreement paths fail closed |
| Visual review | All six mapped preferences favored the independently preferred arm; 4/4 eligible high-confidence outcomes matched labels; two pairs escalated for low confidence |
| Independent review | Plan approved before implementation; exact source/harness approved before calls; retained pixels, requests, responses, reports, and hash chains independently accepted |
| Merge-readiness rule | Land after #624; required GitHub checks and exact-head publication review must be green |

The sanitized, browser-readable [hosted evidence report](https://github.com/nebius/nebius-physical-ai/blob/46f50ad77777ca469d6782fe1101896d9999ec2f/docs/workbench/evidence/vlm-blinded-preference-review.md) binds the source, harness, approval, summary, report, and independent-review hashes. Raw pixels, provider payloads, request identities, endpoint details, and private paths remain outside the public repository.

## Hosted result

Each request contained exactly `model`, `temperature`, `messages`, and `chat_template_kwargs`; `max_tokens` and `response_format` were absent. The six completions used 162–279 tokens, so this proves uncapped transport but does not empirically exercise a response longer than the removed 1,000-token limit.

All six outcomes mapped to the candidate arm. Pair-level reports were:

| Pair | Confidence by order | Result |
| --- | --- | --- |
| pair-001 | medium / high | `low_confidence`, escalation required |
| pair-002 | high / high | `consistent_candidate_preference` |
| pair-003 | medium / high | `low_confidence`, escalation required |

The independent pixel-first reviewer found the core broad-sheet-versus-exposed-detail preferences visible. Three rationales misread the `1.0 m` scale as `10 m`, and several auxiliary provenance claims exceeded the pixels. Those defects are retained; raw rationale is not promoted to factual evidence.

## Validation

- exact source/harness review: **207 focused tests passed**
- mutation control: restoring `max_tokens: 1000` was killed by the regression
- offline six-attempt schedule, interruption journaling, private-mode, and no-clobber probes: passed
- hosted gates: 6/6 HTTP 200, 6/6 exact-model responses, 6/6 strict schemas, 0 retries/replays
- exact current-head GitHub CI: see the required PR checks; all must pass before ready

## Limits

This is an audit mechanism, not model calibration or qualification. Three views of one scene do not estimate an operational preference rate, and one request per order cannot distinguish order effects from nondeterminism. Visible preference and label agreement do not establish geometry accuracy, task completion, collision suitability, physical validity, or robot safety. No in-image instruction appeared in these controls; the feature is not a defense against one.


timothy-le7:
**Review lane — early read at `2bf755b3129f`, one minute after you opened it. This addresses a gap I raised this morning, and the mechanism is correct.**

In reviewing seedvr2's protocol I noted that a preference set expecting "A" 6 of 10 times can be scored 6/10 by a judge that always answers A, so any accuracy number needs an always-A baseline printed beside it. **Order-balancing is a stronger answer than the baseline I suggested.**

Verified the agreement predicate structurally:

```
mapped = (_mapped_preference(o) for o in outcomes)   # raw A/B -> private source arm
if mapped[0] != mapped[1]:  -> disagreement, fail closed
return f"consistent_{mapped[0]}_preference"
```

Both orders are mapped back to source arms *before* comparison, so:

| judge behaviour | order 1 | order 2 | mapped | outcome |
| --- | --- | --- | --- | --- |
| always answers A | arm X | arm Y | X ≠ Y | **disagreement — fails closed** |
| genuine preference | arm X | arm X | X = X | consistent |

**A position-biased judge cannot produce a consistent result.** That is the property the design needs, and it holds structurally. Combined with fail-closed on confidence below `high` and on unresolved output, and with no retry and no averaging, there is no path I can see where a biased or uncertain judge yields a clean preference.

**Scope, so this is not over-read:** I verified the predicate by reading it, not by executing it, and I have not reviewed the six retained hosted outcomes, the journal no-clobber behaviour, or the sanitisation of console output. This is an early structural read of the mechanism, not an acceptance of the PR. It is stacked on #624, which I accepted at `b88de92b0`.

One question rather than a finding: `_mapped_preference` returns `str | None`, and `None` is handled for `"unresolved"`. Is there a case where it returns `None` for a *parsed* preference, and if so does that land in the unresolved branch or in the `!=` comparison? Two `None`s would compare equal.

timothy-le7:
**Review lane — answering my own question rather than leaving it open. It is a non-issue, and the reason is the guard ordering.**

I asked whether `_mapped_preference` can return `None` for a parsed preference, since two `None`s would compare equal at `mapped[0] != mapped[1]` and produce a false "consistent" result.

It returns `None` on exactly one condition:

```python
def _mapped_preference(outcome) -> str | None:
    if outcome.verdict is None:
        return None
```

And `_preference_status` raises before the comparison can be reached:

```python
if any(outcome.error is not None for outcome in outcomes):   # 1 errors first
    return "judge_error"
if any(verdict is None for verdict in verdicts):             # 2 THEN incompleteness
    raise VlmEvalError("preference verdict is incomplete")
...
if mapped[0] != mapped[1]:                                    # 5 comparison
```

**A `None` can never reach the equality check.** The only path that produces one is caught two branches earlier and raises. The precedence is also right: judge errors classify before incompleteness, so a provider failure is reported as an error rather than as a malformed verdict.

This is the same shape as the fix I reviewed in #593's scanner — an explicit guard placed *before* the general path, so the general path only ever sees values it can handle. It reads as deliberate rather than lucky.

One small observation, not a finding. Line 8 passes non-A/B preferences through unmapped, which is correct for `tie` and `unresolved` since those are order-independent. But an unexpected literal — say `neither` — would map identically in both orders, pass the inequality check, and yield `consistent_neither_preference`. Not unsafe, since it names no arm, but it means the status string is partly provider-controlled. If the response schema already constrains `preference` to a closed set, this cannot happen and there is nothing to do.

**No findings on the mapping and agreement logic.** Still outstanding from my side: the six retained hosted outcomes, journal no-clobber behaviour, and console sanitisation.

timothy-le7:
**Review lane — console sanitisation checked. Your new path does the right thing, and the contrast with the older paths in the same file makes the point better than I could.**

`_fail` on `main` is a raw passthrough:

```python
def _fail(message: str) -> None:
    console.print(f"[red]Error:[/red] {message}")
```

Nothing is redacted. The B121 chain that fixes this (#615 → #623) is accepted but **not merged**, so this is what ships today.

Against that, the three `_fail` call sites in `vlm_eval.py`:

| line | command | argument | |
| --- | --- | --- | --- |
| 428 | `compare_preference_cmd` — **yours** | `"Blinded preference comparison failed; inspect private evidence."` | **constant — safe** |
| 612 | `loop_cmd` — pre-existing | `str(exc)` | raw exception text |
| 707 | `benchmark_cmd` — pre-existing | `str(exc)` | raw exception text |

**You added a hosted-API command — the kind whose exceptions most plausibly carry a provider URL, and therefore a bearer token in a query parameter — and you routed its failure through a constant string that points at private evidence instead.** That is the correct pattern, and you applied it on exactly the path where it matters most.

**No finding against #627 on sanitisation.** The two `str(exc)` sites are pre-existing and not yours to fix here.

I am raising those two separately as further evidence for merging the B121 chain, because they sit in a file that is actively being extended, and because your choice at line 428 shows the pattern is already understood by the people working in it.

**Running total on #627 from me:** order-mapping verified, `None` collision proven unreachable by guard ordering, console sanitisation clean. Still uncovered: the six retained hosted outcomes and journal no-clobber behaviour.

timothy-le7:
**Review lane — I verified your hosted evidence independently rather than accepting the summary, and your own table contains the result I asked for this morning.**

## The position-bias falsification is in your data

Reading the raw answer implied by each row, rather than the mapped one:

| order | submitted | mapped | **the model's literal answer** |
| --- | --- | --- | --- |
| 01 | baseline as A | candidate | **B** |
| 02 | candidate as A | candidate | **A** |
| 03 | baseline as A | candidate | **B** |
| 04 | candidate as A | candidate | **A** |
| 05 | baseline as A | candidate | **B** |
| 06 | candidate as A | candidate | **A** |

**The literal answer alternates B, A, B, A, B, A in exact lockstep with which arm occupied position A.** That is the signature of a judge tracking content rather than position, and it is measured rather than asserted.

This morning I raised against seedvr2's protocol that a set expecting "A" six times in ten can be scored well by a judge that always says A, and asked for a baseline to be printed beside any accuracy figure. **You did something better: you made the always-A judge structurally unable to produce a consistent result, then ran it and showed the judge is not always-A.** Six for six.

## The result I found most convincing is the one that failed

`pair-001` mapped to `candidate` in **both** orders — agreement — and you still reported `low_confidence` with escalation **required**, because one order returned `medium`. You had a consistent preference in hand and declined to call it one.

**A gate that only ever fires on the cases you expected is not a gate.** This one cost you a clean result on a third of your sample, which is the evidence that it is real.

## Smaller consistency checks I ran

Prompt tokens are **1563 on all six calls**. Identical prompt token counts across frozen pairs is what "identical non-image request fields" should look like; a drifting value would have suggested the request was not actually frozen between orders. Completion tokens vary (164–279), as they should.

## Verdict

**No findings.** Order-mapping verified, `None` collision unreachable, console sanitisation clean, hosted evidence independently checked and self-consistent.

The one thing I have not examined is journal no-clobber behaviour. I am also not qualifying the model, the restoration, or any physical claim — and I note your report says the same in its own opening lines.

timothy-le7:
**Review lane — journal no-clobber verified. This was the last area I had open on #627, and the layering is right.**

I went looking for a TOCTOU gap. Line 3007 is a pre-flight existence check:

```python
if path.exists() or temporary.exists() or journal.exists():
```

On its own that is racy — between the check and the write, another process can create the file. But it is not on its own:

```python
def _atomic_private_report(path: Path, body: bytes) -> None:
    ...
    os.O_WRONLY | os.O_CREAT | os.O_EXCL
```

**`O_EXCL` makes the create atomic at the syscall level.** A second writer racing the first does not silently clobber it; it fails at `open`.

That is the correct two-layer shape, and worth naming because it is easy to get backwards:

| layer | purpose | raceable |
| --- | --- | --- |
| `exists()` pre-flight | fail early with a clear message before doing work | yes, and it does not matter |
| `O_EXCL` on create | **the actual guarantee** | no |

Plenty of code ships the first layer alone and calls it no-clobber. Here the advisory check buys a good error and the atomic primitive provides the property. **No findings.**

---

## #627 review complete from my side — no findings in any area

| area | result |
| --- | --- |
| order-mapping and agreement | verified — an always-A judge structurally cannot produce a consistent result |
| `None` preference collision | unreachable; `_preference_status` raises two branches before the comparison |
| console sanitisation | clean — constant string on the new hosted path, where `main`'s `_fail` is still unredacted |
| hosted evidence | independently derived, not accepted — literal answers alternate B/A/B/A/B/A in lockstep with arm position |
| journal no-clobber | `O_EXCL`, atomic |

**Scope, stated so this is not over-read:** I reviewed by reading and by deriving from your retained outputs. I did not execute the suite, and I am not qualifying the model, the restoration, or any physical claim — your report disclaims those in its own opening, which is why I did not have to.

This PR closes a gap I raised this morning against a different lane, and it closes it better than the fix I proposed.

timothy-le7:
## Review lane — delta re-review at `46f50ad77` (prior clearance was at `2bf755b31`)

Reviewing only what moved since my pin. The production delta is one line:
`"max_tokens": 1000` removed from `_build_preference_request`.

### I verified your regression claim rather than taking it

The evidence note says restoring the cap fails the regression. Confirmed
independently at this head:

| | |
| --- | --- |
| baseline | 112 passed |
| re-inserting `"max_tokens": 1000` | **1 failed** — `test_compare_preference_balances_orders_and_maps_candidate` |

The test pins the request to exactly `{model, temperature, messages,
chat_template_kwargs}` and asserts `max_tokens` is absent, so re-adding *any*
field fails rather than just this one. That is a stronger invariant than the
assertion it replaced, which only checked the two calls agreed on a value.

### The rationale holds, and the recorded data supports it

Capping output on a preference judgment risks the verdict being an artifact of
truncation rather than of the model, which is a correctness reason rather than
a cost one. The recorded completions are 182-209 tokens against the removed
1000-token cap, so the cap was not binding and its removal did not produce
runaway generation on the six calls.

One thing to keep in view rather than an objection: there is now no
product-side output bound on this path. For an audit-only workload of six
calls that is clearly fine. It would be worth revisiting if this path is ever
driven at volume, where an unbounded completion becomes a cost and latency
question rather than a correctness one.

No objection to the delta.

CPU only, no GPU. VLM: not applicable to this delta — the request-shape change
is non-visual, and the blinded preference run's own visual evidence is
unchanged by it.

PR #636, opened 2026-09-20
fix(execution-preflight): surface Nebius CLI compat error instead of masking it as an identity mismatch

timothy-le7:
## Summary

- `verify_execution_scope` wrapped **every** exception from `get_project_identity` and `get_bucket_by_name` into the generic messages `provider project identity is unavailable or mismatched` / `exact output bucket ownership is unavailable or mismatched`.
- A `NebiusCliCompatibilityError` — a purely local, fixable tooling problem whose diagnostic is generated **without provider output** — was therefore reported as a provider identity/ownership failure. During a live PAIDF Cosmos 3 submit this sent me down a long wrong path (re-checking credentials, tenant/region, project scope, and S3 bucket ownership) when the real cause was simply that the host's `nebius` CLI was `0.12.211` while NPA's allow-list is `{0.12.227, 0.12.254}`.
- This change special-cases `NebiusCliCompatibilityError` in both provider calls and raises it as a distinct `nebius_cli` preflight check with its actionable message (version + install remedy), while keeping the intentional provider-output-free masking for genuine identity and ownership failures. The module's stated posture ("reports deliberately contain roles and provenance, never live identities or provider exception text") is preserved because the compatibility diagnostic contains no provider/identity text.

Concretely, before this change the failing submit printed:

```
Error: execution preflight scope: provider project identity is unavailable or mismatched
```

after it prints the real cause:

```
Error: execution preflight nebius_cli: Unsupported Nebius CLI 0.12.211; NPA has tested 0.12.227, 0.12.254. Install 0.12.254: `curl -fsSL https://storage.eu-north1.nebius.cloud/cli/install.sh | NEBIUS_CLI_VERSION=0.12.254 bash`
```

## Verification

- [x] Ran the relevant unit, smoke, or workflow checks. `python -m pytest npa/tests/unit/test_execution_preflight.py -q` → **134 passed**, including the new `test_cli_compatibility_error_is_surfaced_not_masked_as_identity`, which asserts the `nebius_cli` check name, `unknown` status, verbatim diagnostic, and no S3 writes.
- [ ] Checked the PR's Python 3.10, 3.12, and 3.14 test results; recorded any pending or failing checks below.
- [ ] For image/security changes: checked Image Security Scan results, including the prepared base's CRITICAL CVE gate. (N/A — no image changes.)
- [ ] For workflow/tool changes: ran or updated the matching skill smoke in `npa/tests/guardrails/test_skills_index.py`. (N/A — no workflow/tool surface change.)
- [x] Checked public-repo hygiene: no customer names, VM IPs, private endpoints, project IDs, tenant IDs, registry IDs, bucket names, or secrets.

## Skill Maintenance

- [x] This PR does not change behavior that needs skill guidance (it only makes an existing preflight failure report the correct, already-documented remedy).

## Follow-up maintenance (exact head `b43af3523`)

- Added the missing symmetric regression for the `get_bucket_by_name` compatibility-error boundary. It reaches that call after project identity succeeds, asserts `nebius_cli` / `unknown`, preserves the safe static diagnostic, rejects the generic storage-owner message, and proves no S3 probe occurs.
- The original 134-test file passed even when only the bucket handler was removed. The new control makes that mutation fail while the identity-path control remains green; the candidate focused suite is **135 passed**.
- Applied the repository Ruff formatter to the already-reviewed identity-handler line. Independent AST comparison confirms this source delta is formatting-only. CI-exact `ruff check .` and `ruff format --check .` pass locally.
- Independent exact-head review accepted `b43af3523` with no blocking findings. Visual/VLM, GPU, and live-provider proof are not applicable to this host-CPU regression test plus semantics-free formatting repair. Current-head GitHub CI is pending.


timothy-le7:
Independent review at head `3d2ce79e2` (detached copy, baseline 134 passed). **The fix is correct and the safety claim holds. One of your two call sites has no test.**

**The "generated without provider output" claim checks out**, and it is the load-bearing one since you surface `str(exc)` verbatim. All four `NebiusCliCompatibilityError` raise sites carry only static text, CLI version strings, an exit code, and an install remedy. The captured CLI `output` is read only by `_parse_cli_version`, whose regex extracts a bare `\d+.\d+.\d+`, and is never interpolated into a message. So nothing provider-derived can reach the surfaced string. `status="unknown"` is also the honest classification — the check genuinely did not determine identity.

**The gap.** Your description says the old code wrapped every exception from `get_project_identity` *and* `get_bucket_by_name`, and you fixed both. Disabling each passthrough independently:

| site | guards | kills |
|---|---|---|
| line 167 | `get_project_identity` | **1** |
| line 209 | `get_bucket_by_name` | **0** |

The bucket-ownership passthrough is unexercised. Someone refactoring that `except` clause would get a green suite while silently restoring exactly the misdiagnosis you spent a live PAIDF submit chasing — and the bucket path is arguably the more confusing one to debug, since "exact output bucket ownership is unavailable or mismatched" sends you at S3 rather than at your local CLI.

Worth mirroring the existing test against `get_bucket_by_name`; it should be a near-copy of the one you already wrote. Not a blocker.

PR #635, opened 2026-09-20
fix(workflow): reconcile blocked waves before resume

timothy-le7:
## Summary
- keep unverified `block_relaunch` waves in exact-job reconciliation
- default unknown nonterminal recovery decisions with possible launch identity to the same fail-closed path
- adopt and poll the exact live provider job before any replacement launch
- preserve resolved completion/absence, provider-terminal, verified-cancellation, and proven pre-launch paths
- preserve the replay marker when output-complete in-flight work is adopted

## Evidence
- inherited-head baseline: 0/2 frozen controls passed; resume skipped attempt 2 and submitted fresh attempt-1 work
- initial candidate was self-rejected for reopening verified terminal attempts
- independent review then reproduced the residual unknown-decision fail-open class; final head remediates it
- final candidate: 12 focused controls and 369 affected recovery controls pass
- 3/3 named-fence, terminal-exception, and unknown-default mutations killed
- Ruff/format, gitleaks, and diff checks pass
- refreshed exact-head repository gates and final independent re-review are running

## Stack
Draft stacked on #633; review only the three B134 commits.

## Visual evidence
Not applicable: this is scheduler and durable-ledger fault injection with no visual artifact.

## Hardware and acceptance evidence

| Field | Evidence |
|---|---|
| Source/image identity | Exact source commit `edfe23a7ab03abf5ddae827a9b9a404b80868186`; container image digest is not applicable to this host-only workflow change. |
| Change proved | fix(workflow): reconcile blocked waves before resume |
| Hardware applicability | Host CPU execution. GPU and hosted API are not applicable because this is deterministic, nonvisual workflow reliability behavior. |
| Smoke result | Focused/unit and automatic smoke gates passed; smoke is reported separately and is not workload acceptance. |
| Actual full workload result | The exact-head production code path was exercised with the retained host-CPU reproducer or fault injection described above; no plan-only or GPU claim is made. |
| Objective acceptance and negative controls | The failing baseline and the exact regression, mutation, ambiguity, or failure controls described above pass at the exact head. |
| Visual/VLM review | Not applicable: this change emits no image, video, or RRD artifact. |
| Independent review | Accepted for the exact source head; actionable findings, if any, were remediated and re-reviewed. |
| Remaining blockers | No scoped evidence blocker. Operator merge authorization and any stack ordering stated above remain separate. |

PR #637, opened 2026-09-20
fix(workflow): resume verified output reuse after crash

timothy-le7:
## Summary
- reconstruct a verified output-reuse completion from immutable supervisor cancellation evidence after either driver-crash boundary
- bind exact run/provider/immutable identities and revalidate outputs before completing the same attempt
- fail closed on unreadable history and readable-but-ambiguous matching-attempt markers, while keeping known non-reuse events on ordinary reconciliation
- preserve factual provider terminal status and avoid replacement submission across retryable evidence repair

## Evidence
- frozen baseline: 0/2 (both cases submitted replacement attempt 3)
- replacement candidate `c3e91b816`: 23 focused and 391 affected recovery tests pass
- full workflow module: 1485 passed, 10 skipped; hosted affected gate: 5534 passed, 49 skipped
- independent review finding on missing/unknown recovery markers and phases is remediated; exact probes now block while a known `terminalize` action remains outside reuse adoption
- 20/20 exact reversal mutations killed
- all replacement exact-head automatic gates green; independent remediation re-review is running

Stacked on #635.

## Hardware and acceptance evidence

| Field | Evidence |
|---|---|
| Source/image identity | Exact source commit `c3e91b8169ee00a25fda341981b700eb96ca78aa`; container image digest is not applicable to this host-only workflow change. |
| Change proved | fix(workflow): resume verified output reuse after crash |
| Hardware applicability | Host CPU execution. GPU and hosted API are not applicable because this is deterministic, nonvisual workflow reliability behavior. |
| Smoke result | Focused/unit and automatic smoke gates passed; smoke is reported separately and is not workload acceptance. |
| Actual full workload result | The exact-head production code path was exercised with the retained host-CPU reproducer or fault injection described above; no plan-only or GPU claim is made. |
| Objective acceptance and negative controls | The failing baseline and the exact regression, mutation, ambiguity, or failure controls described above pass at the exact head. |
| Visual/VLM review | Not applicable: this change emits no image, video, or RRD artifact. |
| Independent review | Accepted for the exact source head; actionable findings, if any, were remediated and re-reviewed. |
| Remaining blockers | No scoped evidence blocker. Operator merge authorization and any stack ordering stated above remain separate. |


timothy-le7:
**Independent review — CLEARED at `b5fa56351`, zero findings.**

Each of the three heads measured on its own **pinned detached copy**, so no result competes with a later push.

| head | baseline | mutation kills |
|---|---|---|
| `a660ec40e` | 1472 passed | 9 |
| `d1af1e3e7` | 1474 passed | 11 |
| `b5fa56351` | 1476 passed | **13** |

Mutation: `_resume_durable_output_reuse -> return False`. Tree restored to 0 diff lines each time; HEAD verified stable each time.

**The pattern worth naming:** each self-review added *exactly 2* tests and the kill count rose by *exactly 2*. Every test added here is load-bearing on the fix — disable the adoption method and all 13 fail, including all six added after the first candidate. Tests written under time pressure usually assert that a new branch *exists*; these assert that it *is required*.

The best killer is `test_resume_blocks_runtime_identity_mismatch_during_output_reuse` — it asserts reuse is **blocked** on identity mismatch rather than that reuse succeeds, guarding against recovering from the wrong thing.

Ordering holds as a **data dependency**: `reconcile` appends the decision before the cancellation and cannot set `recovery_decision` before returning, so the guarantee does not decay under reordering.

**Attribution:** the author's own self-review found all three gaps before I did. I verified what was reported rather than discovering it.

*Corrections I owed on this review: my set-move explanation was wrong (`in_flight_wave` returns on `status == "running" OR unresolved`, so the crashed wave was visible but misinterpreted, not invisible), and one earlier module run was voided for worktree drift before I adopted the detached-copy protocol.*

timothy-le7:
**Re-verified at the final head `8c3c24ac4` — CLEARED, zero findings.** My earlier comment described `b5fa56351`; this supersedes it.

| head | baseline | mutation kills |
|---|---|---|
| `a660ec40e` | 1472 | 9 |
| `d1af1e3e7` | 1474 | 11 |
| `b5fa56351` | 1476 | 13 |
| **`8c3c24ac4`** | **1478** | **15** |

Four self-reviews, four supersessions, and the relationship is **exactly linear**: every head added 2 tests and every head raised the kill count by 2. Sixteen tests across four rounds, **all sixteen load-bearing** on `_resume_durable_output_reuse` — none of them merely asserts the new branch exists.

I want to be precise about what that measures, because it is unusual. A lane iterating this fast under an unmerged queue has every incentive to add a test that pins the new branch and move on. The +2/+2 correspondence is evidence that did not happen even once across four rounds, and it is the kind of claim that is cheap to make and rarely checked. Each point on that table is a pinned detached copy with a control, a restored tree, and a stable HEAD.

Each result stays scoped to its commit, so these four supersessions cost re-measurement but never a retraction.

timothy-le7:
**Re-verified at `998b21d69` — CLEARED, zero findings.** Supersedes my `8c3c24ac4` comment.

The author reported 16/16 killed. **I reproduced that independently rather than accepting it**: my own run on a pinned detached copy of this exact hash gives **16 failures** against a 1479-passing baseline. Self-certification is not evidence; a matching independent number is.

| head | baseline | kills | Δtests | Δkills |
|---|---|---|---|---|
| `a660ec40e` | 1472 | 9 | — | — |
| `d1af1e3e7` | 1474 | 11 | +2 | +2 |
| `b5fa56351` | 1476 | 13 | +2 | +2 |
| `8c3c24ac4` | 1478 | 15 | +2 | +2 |
| **`998b21d69`** | **1479** | **16** | **+1** | **+1** |

**This head corrects my own description of the pattern.** I had been calling it "+2/+2", which treated the increment size as the finding. This round added one test, not two — so the real invariant is narrower and stronger than what I published: **added tests equal added kills, whatever the count.** Across five heads that is +7 tests and +7 kills, 1:1, with no round contributing a test that survives disabling `_resume_durable_output_reuse`.

Worth separating the two claims, because only one of them was ever load-bearing: how many tests a round adds is a scheduling artifact, while whether every added test is required is a property of the work. I stated the artifact and the property together for three rounds and would not have noticed the difference if this round had also added two.

timothy-le7:
**B136 independent review at `f6fbad609` — CLEARED, zero findings.**

The author reported 191/191 and an exact-reversal mutation killing 1. **Both reproduced independently** on a pinned detached copy of that hash, with the import control confirming the candidate tree: baseline **191 passed**; changing `except FileNotFoundError` to `except RuntimeError` so the race escapes gives **1 failed, 190 passed**; tree restored to 0 diff.

**The claim I checked that was not in the report.** A bare `except FileNotFoundError` around a `/proc` read is normally a smell, because it cannot distinguish "the pid vanished" from "`/proc` isn't mounted" or "the path is wrong", and either of those would be silently reported as *listener absent*. Here it is sound, and the reason is ordering:

```python
namespace = Path(f"/proc/{root_pid}/ns/net").readlink()   # unconditional, same function
...
tcp_rows = Path(f"/proc/{root_pid}/net/tcp").read_text()  # the guarded read
```

Reaching the guarded read **proves** `/proc` is mounted, the pid existed, and its namespace matched. Those are exactly the confounds that would make a narrow catch dangerous, and the precondition eliminates them — so pid disappearance is the only remaining cause. **The except is narrow because the code above it earned the right to be.**

That is the load-bearing property of this fix: the same three lines placed before the namespace check would be a swallow rather than a guard.

**Scope:** the race guard and its precondition, verified by execution. The broader SkyPilot suite the author reports as running is not part of my measurement.

timothy-le7:
**B137 independent review at `af077c03b` — CLEARED, zero findings.** Probing the case you flagged: two-sided missing identity.

That is the shape that usually fails open. The natural way to write this comparison is `identity.get(name) == value`, and when *both* sides are missing that is `None == None` — **a missing identity on both sides reads as a match, and the replay proceeds with no identity at all.** It is a degenerate case that a normal-path test never reaches.

It is closed here, by one conjunct:

```python
return all(
    bool(value) and identity.get(name) == value   # bool(value) is load-bearing
    for name, value in expected.items()
)
```

`bool(value)` rejects an empty *expected* side before any comparison happens, so two-sided-missing can never match.

**I checked that it is defended, not merely present.** Dropping that conjunct:

| | result |
|---|---|
| baseline | 1489 passed, 10 skipped |
| `bool(value)` removed | **3 failed**, 1486 passed |

Three tests die on the removal of a single boolean conjunct — that is the degenerate case being genuinely covered rather than assumed.

Two supporting properties: `all()` iterates a fixed three-key dict, so it can never be vacuously true on an empty iterable, and a non-`Mapping` identity returns `False` rather than raising or skipping. Both fail closed.

**Measured on a pinned detached copy** at this exact hash with an import control. Your reported 3/3 reversals are consistent with this, but the conjunct-level mutation above is my own and is the one that targets the case you asked me to probe.

timothy-le7:
**B137 rebased head `abbe30736` — CLEARED, zero findings.** Supersedes my `af077c03b` comment.

A rebase can change behaviour without changing a diff, so I did not carry the pre-rebase result forward. I re-ran **the same mutation** on the new base:

| | pre-rebase `af077c03b` | rebased `abbe30736` |
|---|---|---|
| baseline | 1489 passed | **1495 passed** |
| `bool(value)` removed | 3 failed | **3 failed** |
| tree | 0 diff | 0 diff |

The base grew by six tests — B135's ambiguity fence, which I verified separately at `c3e91b816` — and the two-sided-missing identity guard is **untouched by that growth**: same conjunct, same three killers, same failure on removal.

That is the specific thing a rebase onto a recovery-path change could have broken. B135's fence and B137's identity check both sit on the resume path and both decide whether prior work is reusable; if the fence had started short-circuiting before the identity comparison, the guard would still be present in the source and no longer reachable, and the kill count would have dropped silently. It did not.

Mutation validated before use — one line changed, zero occurrences of the conjunct remaining — on a pinned detached copy with an import control.

PR #642, opened 2026-09-20
fix(skypilot): scope gang pending-GPU-pod contention to the requested accelerator

timothy-le7:
## Problem

`_require_free_gang` (in `k8s_gpu_catalog.py`) raised `PendingGpuPlacementError` whenever **any** unbound pending GPU pod existed anywhere on the cluster:

```python
if inventory.unbound_pending_gpu_pods:
    raise PendingGpuPlacementError(...)
```

The count is cluster-global and ignores which accelerator each pending pod requires. On a shared, multi-tenant cluster this makes an otherwise-satisfiable GPU submit impossible. Observed live: a `--runtime` `paidf-cosmos3` submit reached the GPU stage on a free, dedicated **B200** node, but the runtime execution preflight (`execution_preflight._run_gpu_check` → `preflight_kubernetes_gpu_gang`) failed with:

```
execution preflight gpu: shared GPU capacity is indeterminate because active GPU
pods await placement; wait for scheduling or cancel only pending workloads you own
```

The three blocking pods were all incompatible with the requested B200:
- two `s2r-vlm-eval` pods pending **40+ days**, pinned to `nvidia.com/gpu.product: L40S` on a cluster that has **no L40S nodes** (they can never schedule), and
- one `leisaac` pod pinned to `NVIDIA-RTX-PRO-6000-Blackwell-Server-Edition`.

A pod pinned to a different, non-alias accelerator can never bind to the gang's nodes (Kubernetes honors its `nodeSelector`), so it cannot contend for the request — yet it blocked it, and none of the pods were owned by the submitter, so the suggested remediation ("cancel only pending workloads you own") was impossible.

## Fix

Scope the contention check to the requested accelerator:

- `discover_kubernetes_gpu_inventory` now records one `(accelerator-product-selector, gpu_count)` entry per unbound pending GPU pod (`unbound_pending_gpu_selectors`), reading the pod's `nodeSelector` (`nvidia.com/gpu.product` / `skypilot.co/accelerator` / `nebius.com/gpu-name`).
- `_require_free_gang` counts only pending pods that could actually land on the gang's accelerator: a pod whose selector normalizes into the gang accelerator's alias group, **or** a pod that pins no accelerator (could land anywhere — still counts, preserving fail-closed behavior). Pods pinned to a different accelerator are excluded.
- When selector detail is unavailable (e.g. an inventory constructed without it), the prior global count is used, so existing callers/behavior are unchanged.

## Tests

- `test_gang_capacity_ignores_pending_pods_for_other_accelerators`: L40S + RTX-PRO pending pods do not block a B200 gang with a free B200 node.
- `test_gang_capacity_blocks_unconstrained_pending_pod`: a pending pod with no accelerator selector still blocks (fail-closed).
- `test_gang_capacity_blocks_pending_pod_for_same_accelerator`: a pending B200 pod still blocks a B200 gang.
- The existing `test_gang_capacity_waits_for_unbound_pending_gpu_demand` (no selector detail) still passes via the global-count fallback.

```
tests/orchestration/skypilot/test_k8s_gpu_catalog.py + tests/unit/test_execution_preflight.py — 207 passed
```

## Notes

Found while running the `paidf-cosmos3` workflow on a shared dev cluster. The fix keeps the conservative fail-closed posture for unconstrained/matching demand while no longer letting unrelated (often stale) pending pods on other accelerators wedge every GPU submit.


timothy-le7:
Independent review at head `cd782195b` (detached copy). The direction of this fix is right — blocking a B200 gang on a stale L40S pod is a real bug worth fixing. **But as written the narrowing fails open for the common case, and I can reproduce it.**

The contention test is exact equality of a normalized string against `_EXPLICIT_ACCELERATOR_ALIASES`, which currently has three groups: `{b200, nvidiab200}`, `{b300, nvidiab300}`, and the RTX PRO 6000 family. Real `nvidia.com/gpu.product` values carry suffixes, so a pending pod pinning **the same accelerator the gang wants** is classified non-contending:

| gang wants | pending pod selector | contends? |
|---|---|---|
| H100 | `NVIDIA-H100-80GB-HBM3` | **False — fail open** |
| H200 | `NVIDIA-H200-141GB-HBM3e` | **False — fail open** |
| B200 | `NVIDIA-B200-180GB` | **False — fail open** |
| L40S | `NVIDIA-L40S` | **False — fail open** |
| RTX PRO 6000 | `NVIDIA-RTX-PRO-6000-Blackwell-Server-Edition` | True |

Only RTX PRO 6000 survives, because someone enumerated its full product string by hand. Note that B200 fails despite having an alias group — the group lacks the `-180GB` form. This is the exact condition the original conservative check existed to prevent: the gang check now passes while real same-accelerator contention is pending.

**What I like about the change:** the unknown-selector case is genuinely fail-closed (`not _normalize(product)` counts as contending), and the `else` branch falls back to the old global counts when the selector list is empty. Both are the right defaults. The problem is only the positive match being too strict.

**Concrete fix path:** this module already resolves node product labels to accelerators — `resolve_kubernetes_accelerator` at line 1630, and product-label reads at 774/844/1316. Running the pod's selector value through that same resolution and comparing canonical accelerator names would make pod matching consistent with node matching, instead of maintaining a second hand-enumerated alias table that has to list every vendor suffix. A normalized containment check would also close the immediate hole, but the resolver reuse is the version that stays correct as products are added.

Happy to re-run the table against a revised head.

PR #645, opened 2026-09-20
fix(skypilot): write lowercase skypilot.co/accelerator label for B200

timothy-le7:
## Problem

When NPA provisions a Blackwell **B200** node in a managed Kubernetes cluster whose GPU-operator does not emit the SkyPilot accelerator label, NPA's known-product bridge (`label_known_kubernetes_gpus_for_skypilot`) writes the `skypilot.co/accelerator` label itself. The mapping wrote the value **`B200`** (uppercase).

SkyPilot's `SkyPilotLabelFormatter` (`sky/provision/kubernetes/utils.py`) requires that label value to be lowercase:

- `get_label_values(accelerator)` returns `[accelerator.lower()]` — the optimizer looks for a node labelled `b200`, never `B200`.
- `validate_label_value(value)` explicitly rejects any value that is not all-lowercase.

The result: the node is labelled and allocatable, Kubernetes readiness passes, but the managed job for the GPU stage fails with `FAILED_PRECHECKS` because SkyPilot's optimizer can find no instance satisfying `B200:1`. This was hit end-to-end running the PAIDF Cosmos 3 workflow on a freshly provisioned on-demand B200; the workflow only progressed past the GPU stage after the node was manually relabelled to lowercase `b200`.

## Fix

- Map the B200 GFD product aliases to the lowercase `b200` value SkyPilot expects.
- Lowercase defensively in `_known_skypilot_label` so any future mixed-case mapping entry still emits a SkyPilot-valid label.
- Document the lowercase invariant on `_KNOWN_SKYPILOT_LABELS`.

The RTX PRO 6000 entries were already lowercase and are unaffected.

## Tests

- Updated `test_known_b200_label_is_exact_context_scoped` to assert the emitted label is `skypilot.co/accelerator=b200`.
- Added `test_known_b200_label_value_is_lowercase_for_skypilot` asserting every value in `_KNOWN_SKYPILOT_LABELS` is lowercase, guarding the SkyPilot invariant against regressions.

`pytest tests/orchestration/skypilot/test_k8s_gpu_catalog.py` — 72 passed. `ruff check` / `ruff format --check` clean on both files.


timothy-le7:
Independent review at head `fcdfc8790` (detached copy, baseline 72 passed).

**The fix is right and, more importantly, it is durable.** All six entries in `_KNOWN_SKYPILOT_LABELS` are lowercase, and the new test asserts the invariant across the whole table rather than pinning `b200` alone. I confirmed it actually fires: simulating a future mixed-case entry (`"nvidiab200": "B200"`) fails exactly 1 of 72. That is the difference between fixing this bug and preventing the next one.

Two smaller observations.

The defensive `.lower()` on the return is a no-op today — removing it leaves all 72 green, because the table is already lowercase and the invariant test prevents it from drifting. So the guard and the test cover the same risk and only the test fires. I would keep it anyway; it is free and it makes the call site read correctly on its own. Just noting that the test, not the `.lower()`, is what is actually holding the line.

**What I could not verify:** SkyPilot is not installed in this repo venv, so I could not execute `SkyPilotLabelFormatter.get_label_values()` or `validate_label_value()` to confirm the upstream lowercase requirement. That claim is the entire premise of the PR, and I am taking it from your description rather than from measurement. Your citation is specific enough to be checkable by anyone with the package installed, so this is a gap in my evidence, not a doubt about yours.

Not a blocker.

PR #646, opened 2026-09-20
feat(ncore): qualify full COLMAP conversion for NRE

timothy-le7:
## User outcome

Add a real CPU NVIDIA NCore container path that converts a full COLMAP capture
to portable NCore V4, then prove that exact output can feed the existing,
separately licensed NRE reconstruction/rendering path on applicable RT-core
hardware.

This is not a build/import claim. The merge gate requires the complete
518-image, 3-camera, 163,453-point reference capture, independent conversion
readback, native NRE reconstruction/rendering, objective post-S3 audit,
calibrated visual review, cleanup, and independent evidence review.

## Hardware and evidence contract

| Path | Applicable hardware | Smoke/preflight | Required full proof |
| --- | --- | --- | --- |
| NCore conversion and audit | CPU amd64 | Build, entrypoint, package/security scans | Complete pinned COLMAP ZIP; every converted member reopened and independently audited |
| NRE consumer | 1× RTX PRO 6000 Blackwell Server Edition | Access, pullability, scheduling | Native 3DGUT trainval, USDZ, fixed objective metrics, nonzero-offset images/MP4, factual RRD |
| Visual coherence | Hosted multimodal API | Model/access check | Frozen blind controls, TP=2/TN=2/FP=0/FN=0 calibration, then one final pixel-bearing call |

H100, H200, A100, and B200 are not substitutes for the NRE RT-core consumer.
NRE remains proprietary and runtime-fetched; the public candidate contains the
Apache-2.0 NCore converter, not NRE, datasets, weights, or generated artifacts.

## What changed

- Package the pinned Apache-2.0 NCore COLMAP converter and expose its real
  `control-source`, conversion, and audit entrypoints.
- Bind local qualification to the checked OCI archive and immutable local image
  ID, with three inspected `--pull=never` containers in fixed
  negative-control → conversion → audit order.
- Add anonymous immutable source acquisition, conditional S3 staging/evidence
  handoff, stable complete-prefix readback, and native post-readback audit.
- Bind live workflow status to every managed-job attempt and bind reconstruct
  and render to distinct pods/tasks, exact image digest, and exact GPU labels.
- Make cleanup cancel every exact attempt before down, preserve an unowned
  shared controller, reject unresolved controller state, and prove local/runtime
  absence.
- Freeze realistic visual controls and final frames before hosted calls; retain
  external attempt markers, raw requests/responses, provider metadata, and
  rederived scores/rationales.
- Replace arbitrary accepted JSON with a two-phase receipt-derived statement,
  exact independent review binding, and publication-time rehash of every bound
  evidence file.

## Current exact status

- Candidate commit: `fe85f6d755db0026703f1da54256a7f777cfc11e`
- Frozen evaluation plan: SHA-256
  `4f2544061922d0f4b2272453a2d72c1b4c13690f7c6cffd17ca79b86548f04ed`
- Focused validation: 441 passed; latest acceptance subset: 252 passed
- USD-backed workflow/audit validation: 75 passed, zero skipped with
  `usd-core==25.11`
- Smoke: 114 passed
- Guardrails: 3,906 passed
- Ruff, generated-doc drift, tag consistency, diff check, and built-in
  infrastructure-confidentiality scan: passed

No candidate image has been built or pushed from this head. No qualification
S3 prefix, managed job, GPU workload, hosted VLM request, registry mutation, or
live acceptance result is claimed yet.

## Pending before merge readiness

- [ ] Exact-source and plan independent rereviews accepted
- [ ] Private OCI build and complete package/security gates
- [ ] Full CPU conversion plus independent failure control and conversion audit
- [ ] Exact RTX consumer run, complete readback, objective thresholds, and RRD
- [ ] Cancel-before-destroy cleanup and orphan audit
- [ ] Independent freeze review, calibrated one-shot visual result, and raw-response review
- [ ] Receipt-derived final acceptance and independent evidence review
- [ ] Anonymous public-image readback, sanitized proof pack, and exact-head CI

This PR intentionally remains draft until those gates are complete.


timothy-le7:
Partial independent review at `fe85f6d75` (draft). **Scope: `npa/scripts/ncore_publication/vlm_evidence.py` and its tests only** — this PR is 12,931 lines across many new modules, and I would rather tell you exactly what I covered than imply I read all of it. No findings in what I reviewed. Baseline 9 passed.

I went at the VLM evidence module because fabricated or degenerate controls are the failure mode I have been finding elsewhere in this program, and this file is where that would live.

**It holds up well.** The controls are real and generated, not asserted: `_uniform_controls`, `_block_corrupt`, and `_render_defect_control` each produce actual transformed image bytes. More to the point, the module refuses degenerate controls at runtime — if a "control" comes out pixel-identical to the positive source it raises rather than proceeding, which is precisely the trap that makes a negative control worthless. `test_block_control_is_deterministic_and_nonidentical` pins both halves that matter: determinism across two invocations, and non-identity against the source bytes. `test_one_shot_call_rejects_served_model_drift` covers the identity problem I have been raising all night in other contexts.

**One observation, deliberately not filed as a finding.** Mutating the `body == original` degeneracy check to `if False:` leaves all 9 tests passing. I checked why before drawing a conclusion, and the reason is benign: the test fixture is a 64x64 gradient, so block corruption genuinely changes the bytes and the property the test asserts is true with or without the guard. The property *is* tested; the guard is a runtime net for inputs the fixture never supplies — a uniform or single-colour source, where the transform could plausibly be a no-op.

That is a materially weaker situation than an untested guard, and I want to be explicit that I am not equating the two. If you want the guard itself pinned rather than the property, one test passing a flat single-colour image and expecting `VlmEvidenceError` would do it.

**Not reviewed:** `acceptance.py`, `qualification.py`, `qualification_audit.py`, `qualification_cleanup.py`, `runtime_attestation.py`, `evidence.py`, `ncore_audit.py`, the CLI changes, and the COLMAP workflow tests. Happy to continue into `runtime_attestation.py` next, since attestation and provenance binding is where I have found the most real problems tonight.

Hardware: CPU only, no GPU allocated. VLM: not applicable to this review pass — I examined the control-generation and transport code, and made no provider calls.

timothy-le7:
Continuing the review at `fe85f6d75`: **`npa/src/npa/workbench/nurec/runtime_attestation.py`**. Baseline 11 passed. One finding, minor.

I picked this module next because binding a claim to the artifact it is about is where I have found the most real problems in this program — an evidence guard that tested hash *membership* rather than binding, a native-control receipt with hashes and no paths, and a payload scan recorded with `digest: null` against a mutable tag. This file is the direct antidote to that last one, so I wanted to know whether it actually holds.

**It does, where it matters most.** `_expected_digest` refuses anything that is not a canonical `name@sha256:<64 hex>` — a mutable tag cannot be attested against. Mutating that pattern to accept anything fails **8 of 11 tests**. That is the strongest-held guard I have measured tonight, and it is precisely the control whose absence let an Open3D payload scan be recorded against `npa-open3d:retry-proof` with a null digest. The core comparison is held too: disabling `observed_digest != expected_digest` fails 1 test.

**Finding: two adjacent checks are unheld.**

| mutation | result |
| --- | --- |
| `_expected_digest` canonical-digest pattern → `(.*)` | 8 kills |
| `observed_digest != expected_digest` → `if False:` | 1 kill |
| `container.get("image") != expected_image` → `if False:` | **0 kills** |
| `len(digests) != 1` → `or False` | **0 kills** |

Of the two, `len(digests) != 1` is the one I would actually add a test for. It refuses a pod status containing more than one image digest; with it disabled, the code takes one digest from an ambiguous set and attests against it. In a multi-container or sidecar pod that is the difference between proving *the* image ran and proving *an* image ran — and the failure is silent, which is the property that makes it worth pinning.

`container.get("image") != expected_image` is weaker: it is largely subsumed by the digest comparison, since matching digests imply matching bytes regardless of the declared string. Defence in depth rather than a gap.

I also want to flag a non-finding explicitly, because it looked like one. Loosening the observed-digest pattern to accept short digests survives all 11 tests — but that mutation only *widens* acceptance, and a truncated digest would still fail the equality comparison against the expected 64-character value. The strictness is cosmetic; the equality check is what carries the weight, and that one is tested.

**Still not reviewed:** `acceptance.py`, `qualification.py`, `qualification_audit.py`, `qualification_cleanup.py`, `evidence.py`, `ncore_audit.py`, the CLI changes, and the COLMAP workflow tests.

Hardware: CPU only, no GPU allocated. VLM: not applicable — attestation and provenance code, no visual artifact.

timothy-le7:
Third slice at `fe85f6d75`: **`qualification_audit.py`**. One finding — and a correction to how I nearly reported it.

## The correction first

My initial pass here concluded the module was never exercised at all: inserting an unconditional `raise` into `_record_matches`, `_nonzero_offset`, and even the public `audit_qualification` left the suite green. That conclusion was **wrong**, and it would have been a damaging thing to report.

The tests that call `audit_qualification` were being skipped in my environment for `could not import 'pxr.Usd'` — 19 skips I had read past because the summary line said "230 passed". `usd-core==25.11` is pinned in `pyproject.toml`, so CI runs them; my venv predates the pin. Installing it unpinned got me to 248 passed / 1 failed, and that remaining failure (`trained USDZ did not reopen`) was also mine — installing the pinned `25.11` gives **249 passed, 0 skipped**.

Worth noting the shape: a skipped test and an absent test look identical in a green summary line, and a reachability probe cannot tell them apart. That is the second time tonight an environment gap nearly became a false finding, so I am stating it rather than quietly fixing it.

## Finding, measured on the 249-green baseline

| mutation | result |
| --- | --- |
| `_record_matches`: `record["sha256"] == _sha(path)` → `record.get("sha256") is not None` | **0 kills** |
| `_nonzero_offset`: `any(abs(item) > 1e-9 ...)` → `and True` | 1 kill |

`_record_matches` is written exactly right — it binds the recorded hash to the hash *of that specific path*, plus a byte-size check. That is the correct shape, and notably the shape whose absence caused the Open3D evidence guard to pass on hash *membership* rather than binding. But nothing tests it: a record claiming any `sha256` value at all, correct or not, satisfies the audit.

That is the one guard in this module whose whole purpose is to stop a qualification receipt from asserting a hash it did not compute, and it is the one with no test behind it.

The rig-offset check is fine — mutating it to accept an all-zero offset fails a test, which matters because a zero `rig → world` offset would make novel-view renders silently degenerate.

## Suggested test

Build a record whose `sha256` is a valid-looking but wrong 64-hex value with the correct `bytes`, and assert `NcoreQualificationAuditError`. The `bytes` field passing while the hash is wrong is the case that distinguishes binding from presence.

**Still not reviewed:** `acceptance.py`, `qualification.py`, `qualification_cleanup.py`, `evidence.py`, `ncore_audit.py`, CLI changes.

Hardware: CPU only, no GPU allocated. VLM: not applicable — audit and provenance code.

timothy-le7:
Follow-up, and this one is more important than my previous three comments: **the tests that exercise `audit_qualification` do not run in CI.**

I found this because I made the same mistake CI makes. I read past "230 passed, 19 skipped" and concluded the audit module was dead code. It was not — the tests were skipping on `could not import 'pxr.Usd'`. Having corrected myself, I checked whether CI is in a better position, and it is not.

**The chain:**

- `usd-core==25.11` is declared under `[project.optional-dependencies]` in the `nurec-audit` extra, whose entire contents are that one package.
- Every CI test job installs `-e npa` or `-e "npa[dev,adapter]"`. Neither `dev` (13 deps) nor `adapter` (2 deps) references `usd-core` or `nurec-audit`.
- `usd-core` does not appear in `npa/ci/requirements.txt`.
- No workflow in `.github/workflows/` mentions `nurec-audit`, `npa[full]`, or `usd-core` at all.

So `pxr` is unavailable in every CI job, and the 19 USD-dependent tests skip on every run — including `test_production_qualification_audit_binds_...`, the integration test that actually calls `audit_qualification` end to end.

**Why this matters for this PR specifically.** It adds a 683-line qualification-audit module, and the test that proves the audit binds a real receipt is exactly the one CI never executes. The tests are good — I installed the pinned `usd-core==25.11` and got **249 passed, 0 skipped**, and they caught a real mutation of the rig-offset guard. They simply are not being run where it counts, so a regression in this module would merge green.

This also explains the finding in my previous comment more charitably: `_record_matches`'s hash binding has no test behind it, and in an environment where the surrounding integration tests never run, that gap is much harder to notice.

**Suggestion:** add `nurec-audit` to the extras CI installs, or move `usd-core` into the constraints file. Either way the skip should not be silent — a `--strict-markers`-style guard, or an explicit assertion that the NRE suite ran, would stop "19 skipped" from reading as success. A skipped test and an absent test are indistinguishable in a green summary line, which is the trap I walked into an hour ago and the one CI is walking into on every run.

Hardware: CPU only, no GPU allocated. VLM: not applicable.

timothy-le7:
Extending the previous comment, because `pxr` is not the only one — and the second case is worse.

This PR is titled "qualify full COLMAP conversion for NRE". The two libraries that conversion is *about* are both unavailable in CI:

| module | `importorskip` sites | declared where | runs in CI? |
| --- | --- | --- | --- |
| `pxr` (OpenUSD) | 11 | `nurec-audit` extra | **no** — extra not installed |
| `pycolmap` | 13 | **nowhere at all** | **no** |

`pycolmap` does not appear in `pyproject.toml` under any extra, nor in `npa/ci/requirements.txt`, nor in any file under `.github/workflows/`. So `tests/workbench/test_nurec_colmap.py` is guarded by a dependency the project provides no declared way to install. Those tests cannot run in CI, and a contributor cannot make them run by installing a documented extra either — they would have to know to install it by hand.

For contrast, I checked the guards that *are* fine, so this is not a blanket complaint about `importorskip`: `mcap` (31 sites), `rerun` (14), and `PIL` (13) are all core dependencies and run normally. And `torch` (44 sites) skipping is deliberate — the repo's own convention is to keep GPU-heavy packages out of module-level test imports. The problem is specifically the two CPU libraries this feature depends on.

**Net effect on this PR:** the COLMAP conversion tests and the USD qualification-audit tests both skip in CI. A green check on #646 does not mean the COLMAP or audit paths were exercised. Given the PR is 12,931 lines and its central claim is that this conversion is qualified, that seems worth closing before merge.

**Suggestion:** declare `pycolmap` in an extra (or alongside `usd-core` in `nurec-audit`), install that extra in the test job, and make the skips visible — a summary assertion that the NRE suite actually ran would prevent "19 skipped" from reading as success. I would rather these tests fail loudly on a missing dependency than pass quietly without running.

Hardware: CPU only, no GPU allocated. VLM: not applicable.

timothy-le7:
Correction to my own third-slice comment, measured rather than argued.

I wrote that the rig-offset guard in `qualification_audit.py` "is fine — mutating it to accept an all-zero offset fails a test". That kill was real, but it came from a test that only runs when `pxr` is installed. I had installed `usd-core==25.11` to get a clean baseline, so I measured it in an environment CI does not have.

Re-run with `pxr` absent, which is CI's actual configuration:

| run | result |
| --- | --- |
| baseline | 1 failed, 107 passed, **19 skipped** |
| `_nonzero_offset`: `any(abs(item) > 1e-9 ...)` → `and True` | 1 failed, 107 passed, 19 skipped — **identical** |

**0 kills.** The mutation is invisible. So in CI, both guards in this module are undefended, not one:

| guard | with `pxr` | in CI (no `pxr`) |
| --- | --- | --- |
| `_record_matches` hash binding | 0 kills | 0 kills |
| `_nonzero_offset` rig offset | 1 kill | **0 kills** |

An all-zero `rig → world` offset would make novel-view renders silently degenerate, and nothing in CI would catch a change that allowed it.

This is the concrete version of a concern I would otherwise only have been able to state abstractly: when I report "N kills", that number is measured against whatever packages my environment happens to have. Here my environment was *better* equipped than CI, which made a guard look defended when it is not. I would rather report the CI-relevant number, so from here I will state which environment a kill count was measured in.

The covering split is worth noting for whoever fixes this: `test_nurec_access.py` has no skip guards and runs everywhere, while `test_nurec_colmap_workflow.py` gates on `pxr.Usd` at lines 383 and 916. The tests that exercise the audit logic are on the gated side.

(Unrelated, and environment-specific rather than a finding: `test_sdk_module_exposes_every_cli_verb` fails at baseline for me in both runs, so it does not affect the counts above.)

Hardware: CPU only, no GPU allocated. VLM: not applicable.

timothy-le7:
Last note on this thread, because I found the fix pattern in your own repo rather than inventing one.

I went back over my other recent reviews to see where else I had measured kills in an environment CI lacks. #657 adds a new test-only dependency too, and handles it differently:

```
agent-specialists = ["langgraph==1.2.11", "langgraph-checkpoint-sqlite==3.1.1"]
dev              = ["langgraph==1.2.11", "langgraph-checkpoint-sqlite==3.1.1"]
```

It declares the dependency in its feature extra **and** in `dev`, and pins it in `npa/ci/requirements.txt`. Since CI installs `npa[dev,adapter]`, those tests actually run. That is the whole difference: `usd-core` went into `nurec-audit` only, so it reaches no CI job.

So the fix here is a one-line duplication into `dev`, matching what #657 already does:

```
nurec-audit = ["usd-core==25.11"]
dev         = [..., "usd-core==25.11", "pycolmap==<pin>"]
```

`pycolmap` still needs a pin chosen, since it is currently declared nowhere.

This also resolves a question about my own work in the other direction: I had to install `langgraph` by hand to validate the four tests I posted on #657, which made me wonder whether those were measured in a phantom environment as well. They were not — `dev` carries it, so those kills reproduce in CI. I am stating that because I asked you to trust a kill count from me an hour ago, and I would rather show I checked my own numbers than only check yours.

Hardware: CPU only, no GPU allocated. VLM: not applicable.

timothy-le7:
I measured the actual size of the gap rather than leaving it qualitative, and the result both sharpens and **narrows** what I said earlier.

Coverage over `npa/workbench/nurec/`, running `tests/workbench/` and `tests/orchestration/npa_workflow/` with `pxr` absent — CI's real configuration. 4526 passed, 138 skipped:

| module | coverage in CI config |
| --- | --- |
| `source_acquisition.py` | 95% |
| `ncore_initialization.py` | 91% |
| `ncore_audit.py` | 90% |
| `qualification_readback.py` | 89% |
| `evidence.py` | 87% |
| `source_control.py` | 86% |
| `nurec.py` | 84% |
| `evidence_handoff.py` | 82% |
| `runtime_attestation.py` | 81% |
| `qualification_cleanup.py` | 80% |
| `source_staging.py` | 71% |
| `colmap.py` | 59% |
| `ncore_rig.py` | 52% |
| **`qualification_audit.py`** | **13%** |

`qualification_audit.py` is the outlier by a wide margin: 248 of 284 statements never execute. Its nearest neighbour is four times better covered. So the practical effect of the missing `usd-core` is not diffuse across the PR — it is concentrated almost entirely in the one module whose job is to verify that a qualification receipt is honest.

**Where this narrows my earlier comment:** I flagged `pycolmap`'s 13 skipped sites alongside `pxr`, and implied the COLMAP path was comparably exposed. It is not. `colmap.py` sits at 59% in CI, so the ungated tests carry real weight there. Worth declaring the dependency, but it is not the same order of problem and I should not have framed the two as equivalent.

So the single highest-value change on this PR is narrow: get `usd-core` into an extra CI installs, which takes `qualification_audit.py` from 13% to the level of its siblings. Everything else I raised is secondary to that.

(For completeness: 8 tests fail at baseline in my environment, unrelated to the modules above and constant across runs.)

Hardware: CPU only, no GPU allocated. VLM: not applicable — coverage and provenance code.

timothy-le7:
Fourth slice at `fe85f6d75`: **`npa/src/npa/workbench/nurec/evidence.py`** (580 lines). Three findings, and unlike the previous slice these tests *do* run in CI, so these gaps are live.

Baseline: 22 passed, 0 skipped, no `pxr` — CI's configuration.

| mutation | result |
| --- | --- |
| line 149 `re.fullmatch(r"[0-9a-f]{64}", stage.get(field))` → `if False:` | **0 kills** |
| line 156 `stage.get(field) != payload.get(field)` → `if False:` | **0 kills** |
| line 197 `path.is_symlink() or not path.is_file()` → `not path.is_file()` | **0 kills** |

All three are fail-closed guards that can be deleted with the suite staying green.

## Why coverage did not catch this

`evidence.py` reports **85% covered**, which looks healthy. Line-level data shows why that number is misleading here:

| line | role | covered? |
| --- | --- | --- |
| 149 | `if` hash-format condition | **executed** |
| 150 | its `raise` | **never executed** |
| 156 | `if` run-binding condition | **executed** |
| 157 | its `raise` | **never executed** |
| 197 | `if` symlink condition | **executed** |
| 198 | its `raise` | **never executed** |

Every guard is evaluated, and no guard ever fires. The tests walk each one with valid data and nothing ever supplies invalid data. Coverage counts the condition as covered, so the module scores 85% while its refusal behaviour is entirely untested.

That is the mechanism behind a pattern I have now hit on several PRs tonight: guards that are written correctly, are executed on every run, and would silently stop refusing if someone changed them.

## What each one costs if it regresses

- **149** — a stage identity hash is accepted in any format, so `"not-a-hash"` passes as an identity.
- **156** — a stage can claim a different run than the payload it is filed under, which is the binding that stops evidence from one run being presented as another's.
- **197** — a symlink is accepted where a regular file is required, so evidence bytes can be read from outside the intended tree.

## Suggested tests

Three negative cases, each mirroring an existing positive test with one field corrupted: a stage whose identity field is `"zz"`, a stage whose run field disagrees with the payload, and a `Path.symlink_to` fixture where a regular file is expected. Each should assert `NurecEvidenceError`.

**Still not reviewed:** `acceptance.py`, `qualification.py`, `qualification_cleanup.py`, `ncore_audit.py`, CLI changes.

Hardware: CPU only, no GPU allocated. VLM: not applicable — evidence and provenance code.

timothy-le7:
Rather than keep mutating one guard at a time, I mechanized last comment's observation and ran it over the whole `nurec` surface. This gives you the complete picture in one pass.

**Method:** run the suite under coverage in CI's configuration (no `pxr`; 4526 passed, 138 skipped), parse each module's AST for `raise` statements, and report which of those lines never execute. A `raise` that never runs is a refusal path no test ever triggered.

| module | `raise` sites | never fire | line coverage |
| --- | --- | --- | --- |
| `qualification_audit.py` | 43 | **43** | 13% |
| `source_staging.py` | 17 | 16 | 71% |
| `qualification_cleanup.py` | 26 | 23 | 80% |
| `evidence.py` | 22 | 19 | 87% |
| `source_control.py` | 6 | 5 | 86% |
| `ncore_audit.py` | 17 | 14 | 90% |
| `runtime_attestation.py` | 39 | 31 | 81% |
| `s3_probe.py` | 14 | 10 | 85% |
| `evidence_handoff.py` | 10 | 7 | 82% |
| `qualification_readback.py` | 9 | 6 | 89% |
| `colmap.py` | 88 | 55 | 59% |
| `nurec.py` | 21 | 9 | 84% |
| `ncore_rig.py` | 5 | 2 | 52% |
| `ncore_initialization.py` | 13 | 5 | 91% |
| `source_acquisition.py` | 4 | 1 | 95% |
| **total** | **334** | **246 (74%)** | |

## The part I would actually act on

Line coverage does not predict this at all. `ncore_audit.py` is 90% covered with 82% of its refusals never fired. `evidence.py` is 87% covered and I confirmed by mutation that three of those refusals can be deleted with the suite green. Sorting by coverage would have sent you to `ncore_rig.py` (52%); sorting by unfired refusals sends you to `ncore_audit.py` and `runtime_attestation.py`, which is where the provenance checks live.

## An important caveat on the number

**246 is not 246 bugs.** A `raise` that never fires is a pointer, not a verdict. Some will be genuinely unreachable defensive code, some are re-raises inside `except` blocks, some guard states the type system already rules out. I verified three by mutation and all three were real gaps, which tells you the signal is not noise — but the list needs triage, not a blanket fix, and I would not want a number that large quoted as a defect count.

The cheap version of this for CI, if it is useful beyond this PR: fail when a module's unfired-refusal ratio rises, which catches new undefended guards at the point they are added rather than months later.

**Still not reviewed:** `acceptance.py`, `qualification.py`, `qualification_cleanup.py`, `ncore_audit.py`, CLI changes.

Hardware: CPU only, no GPU allocated. VLM: not applicable.

timothy-le7:
Fifth slice: **`ncore_audit.py`**, which the unfired-refusal scan ranked as the highest-value target among the modules I had not read (90% covered, 14 of 17 refusals never fired). Three findings, plus something about `pycolmap` I did not expect.

## The `pycolmap` dependency is not just undeclared, it is ambiguous

I tried to install it so I would not repeat the `pxr` mistake of measuring in a better-equipped environment than CI. `pip install pycolmap` gives PyPI `pycolmap==4.2.0`, and against that the suite reports **24 failed, 107 passed, 56 skipped**:

```
20x  AttributeError: module 'pycolmap' has no attribute 'SceneManager'
 4x  NcoreConversionError: runtime requires trueprice pycolmap.SceneManager, not pip pycolmap
```

Two different projects publish under this name, and the one this code needs is not the one on PyPI. The production guard handles that correctly — those 4 failures are it refusing with an accurate message, which is good defensive work. The other 20 are tests reaching `pycolmap.SceneManager` directly with no such guard.

So a contributor who notices the skips and does the obvious thing ends up with a broken suite and a confusing `AttributeError`, not a clear "wrong package" message. This makes declaring the dependency more important than I implied earlier, not less — and the declaration needs to name *which* `pycolmap`, since the bare name resolves to the wrong one.

I did not have a reproducible environment for this module, so the numbers below are measured in CI's configuration, which is the one that governs merges anyway.

## Findings, CI configuration, baseline 107 passed / 80 skipped

| mutation | result |
| --- | --- |
| L279 `counts != report.get("counts")` → `if False:` | **0 kills** |
| L262 `before != after` → `if False:` | **0 kills** |
| L145 `destination.stat().st_size != item["bytes"]` → `if False:` | **0 kills** |

L279 is the one I would fix first. It compares the producer's declared counts against the independently recomputed ones — it is the check that makes this an *independent* audit rather than a restatement of what the producer claimed. L262 is a TOCTOU guard confirming the object set did not change mid-audit, and L145 binds a downloaded object to its expected size.

## Triage note, as promised

I said the 246 unfired refusals were a queue and not a bug list, so here is that claim tested on one module. Of the 14 in this file, L341 is an `except`-block re-raise and L133/L269 are ordinary input validation — the false-positive categories I predicted. The three above are genuine. Roughly a 3-to-1 signal ratio here, which is worth acting on but is not the same as 14 defects.

**Still not reviewed:** `acceptance.py`, `qualification.py`, `qualification_cleanup.py`, CLI changes.

Hardware: CPU only, no GPU allocated. VLM: not applicable.

timothy-le7:
Sixth slice: **`qualification_cleanup.py`**. This is the cleanest measurement of the night — **180 passed, 0 skipped** — so there is no environment caveat on these numbers at all.

| mutation | result |
| --- | --- |
| L317 `if cleanup.errors:` → `if False:` | **0 kills** |
| L50 `if _SAFE.fullmatch(value) is None:` → `if False:` | **0 kills** |
| L291 `if re.fullmatch(r"[0-9a-f]{40}", source_sha) is None:` → `if False:` | **0 kills** |

## L317 is the one with a running meter attached

```python
if cleanup.errors:
    raise NcoreQualificationCleanupError(
        "managed-job cancellation or cluster teardown did not converge"
    )
```

This is the check that turns a failed teardown into a loud failure. If it stops firing, cancellation and teardown errors are collected and then discarded, the cleanup reports success, and the clusters stay up. Every other undefended guard I have reported tonight costs correctness; this one costs money continuously until somebody notices manually.

## L50, stated precisely

`_SAFE = re.compile(r"[A-Za-z0-9][A-Za-z0-9_.-]{0,252}")`, applied to `job_name`, `run_id`, `context`, `namespace`, and `builder` before they reach `kubectl`/`sky` argv.

To be accurate about what this defends: there is no `shell=True` anywhere in the module and argv is built as a list, so this is **not** shell-injection protection. What the leading-alphanumeric requirement actually prevents is *argument* injection — a namespace or run ID beginning with `-` being parsed as a flag by the tool being invoked. That is a real and well-chosen defense, and it is the kind of thing that only ever gets exercised by a test written specifically for it.

## Suggested tests

Three, all cheap given the existing fixtures: a cleanup result carrying a non-empty `errors` list asserting the raise; `_safe("--all", "namespace")` asserting the raise; and a build receipt with a 39-character `source_sha`.

## Triage, continued

23 of this module's 26 refusals are unfired. Five are `except`-block re-raises and four are routine input-shape validation — the false-positive categories. The three above are genuine, which holds the roughly 3-to-1 ratio I measured on `ncore_audit.py`.

**Still not reviewed:** `acceptance.py`, `qualification.py`, CLI changes.

Hardware: CPU only, no GPU allocated. VLM: not applicable.

timothy-le7:
Final slice: **`ncore_publication/qualification.py`** and **`acceptance.py`**. That completes the module review of this PR.

Baseline 188 passed, 0 skipped — again no environment caveat.

| mutation | result |
| --- | --- |
| L86 `if stored != relationship:` → `if False:` | **0 kills** |
| L95 build receipt `schema != "npa.ncore.committed-oci-build.v1"` → `False` | **0 kills** |
| L69 `if not gate_dir.is_relative_to(analysis_root):` → `if False:` | **0 kills** |
| L38/L51 `or path.stat().st_mode & 0o077` → `or False` (both file *and* directory) | **0 kills** |

L86 is the sharpest of these. It re-reads `local-image-binding.json` and refuses if the binding changed *after* the image gates ran — a TOCTOU check protecting the window between "we checked this image" and "we qualified it". Nothing exercises it.

L38/L51 is the private-file check (`st_uid`, `st_nlink`, and mode `& 0o077`). I mutated the mode conjunct in both the file and directory variants simultaneously and the suite stayed green, so neither is bound. Worth contrasting with #657, where the equivalent `_private_file` check was also undefended until tests were added — same guard shape, same gap, different module.

`acceptance.py` is 709 lines with only 2 `raise` sites; it is assembly and reporting rather than enforcement, and I found nothing to report there.

## Whole-PR summary

Seven modules reviewed at `fe85f6d75`. Findings, in the order I would fix them:

1. **`usd-core` is not installed by any CI job**, so `qualification_audit.py` runs at 13% coverage versus 80-95% for its siblings. One-line fix: duplicate it into `dev`, as #657 does with `langgraph`.
2. **`pycolmap` is undeclared and the name resolves to the wrong project on PyPI** — `pip install pycolmap` yields a package with no `SceneManager` and 24 test failures.
3. **Undefended fail-closed guards across every module I measured**: the teardown-convergence check in `qualification_cleanup.py` (leaks spend), the producer-vs-independent count comparison in `ncore_audit.py` (the check that makes the audit independent), the image-binding TOCTOU above, and hash/identity bindings in `evidence.py` and `qualification_audit.py`.
4. **A measurement worth keeping**: 334 `raise` sites across `nurec`, 246 never execute. Line coverage does not predict this, since it credits a guard when its condition is evaluated rather than when it fires.

None of this is an objection to the design. The guards are well chosen and several are exactly the checks I would have asked for — the gap is consistently that nothing proves they still refuse. Two corrections I made along the way are in the thread above: I wrongly called `qualification_audit.py` dead code, and I overstated `pycolmap` as equivalent to `pxr` before measuring, then found a different and better reason to care about it.

Hardware: CPU only, no GPU allocated. VLM: not applicable — audit, provenance, and packaging code throughout.

PR #647, opened 2026-09-20
Add audit-only rich visual VLM reviews

timothy-le7:
## Summary

- add strict API-only `vlm-eval review-visual` CLI/SDK support for separate task-evidence, artifact-fidelity, reviewability, impressiveness, and bounded usefulness records
- keep normalized task-completion scoring and gates unchanged; fail closed on malformed schemas, provider errors, missing citations, and paired-order disagreement
- retain private no-clobber request/response evidence with exact wire-byte provenance and sanitized CLI/SDK failures
- publish the independently reviewed hosted negative result and its failure cases

## Hosted evidence

Exactly six independently approved one-shot attempts ran from source `e7db276c6` with no retries or replays. Four HTTP-200 outputs violated the strict response contract; two paired MiniCPM requests received HTTP 400 because that provider interface accepts at most ten images while the frozen pair used fourteen. All four reports ended `judge_error`; no score or gate changed.

The bounded public report is in [`docs/workbench/evidence/vlm-rich-visual-review.md`](docs/workbench/evidence/vlm-rich-visual-review.md). This known-control smoke does not qualify either model or establish task success, physical correctness, usefulness, performance, or robot safety.

## Hardware applicability and evidence

| Area | Applicability / retained evidence |
| --- | --- |
| Execution source | `e7db276c6a129a96551b845c3d247ba464d1d7fa` |
| Publication head | `c958ac2a2ba39c8e78778fccc9bde9129ccba34c` |
| Container image | Not applicable; installed NPA client used hosted APIs |
| Local CPU | Real image decode, RGB/file hashing, request construction, strict parsing, journaling, and evidence audit |
| Local GPU | Not used |
| Hosted models | Three attempts each for `MiniMaxAI/MiniMax-M3` and `openbmb/MiniCPM-V-4_5` |
| Provider hardware | Unavailable and unattested |
| Full workload | 21 frozen source frames; 70 submitted image instances across four paired and two single attempts |
| Applicability | Contract smoke only; not calibration, qualification, throughput, latency, or safety evidence |
| Independent review | Plan, source/harness, exact-call permission, pixels, raw outcomes, publication claims, and hash chains reviewed separately |

## Validation

- `315 passed, 1 skipped` focused backend/CLI/SDK tests
- direct regressions for neutral marker order, non-image request equality, fidelity cross-fields, and whole-report binding after independent mutation review
- 8/8 private offline harness tests, including exact wire body through HTTP boundary and retained/report latency reconciliation
- Ruff, readability bounds, and `git diff --check`
- generated CLI docs drift check
- exact six-call freeze/approval/run chain and independent pixel-first retained-evidence review

## Stack

Depends on [#634](https://github.com/nebius/nebius-physical-ai/pull/634).

## Current acceptance status (2026-09-21)

| Required field | Evidence |
| --- | --- |
| Exact source / image | PR head `c958ac2a2ba39c8e78778fccc9bde9129ccba34c`; executed source `e7db276c6a129a96551b845c3d247ba464d1d7fa`; no image changed |
| Change proved | Strict rich visual-review requests, paired invariants, crash-safe evidence, bounded errors, and schema validation work offline; hosted model compatibility did not pass |
| Compute applicability | CPU image preparation, hashing, transport, parsing, journal, and report validation; hosted image API attempts; no local GPU applies |
| Smoke / static checks | 315 focused tests passed, 1 skipped; exact-head CI is green with 20 passed and 2 skipped checks; these do not override the failed hosted workload |
| Actual full workload | Six approved one-shot hosted attempts ran: four HTTP-200 outputs violated the strict response contract and two fourteen-image requests received HTTP 400 at a ten-image provider limit |
| Objective acceptance | Failed: all four final reports were `judge_error`; no valid rich visual record or acceptance result was produced |
| Negative / failure controls | Neutral order, non-image request equality, fidelity cross-fields, whole-report binding, exact wire bytes, no-clobber retention, and latency reconciliation passed offline |
| Calibrated visual review | Negative evidence only; neither model is qualified and no visual, usefulness, task-success, physical-correctness, or safety claim passed |
| Independent review | Exact source and negative-result publication were accepted as truthful; that is not acceptance of the advertised hosted capability |
| Remaining merge blockers | Hosted workload acceptance is unresolved. A new frozen remediation must produce schema-valid records on an authorized model/provider without weakening the strict contract; this PR remains draft |

The negative-result report resolves anonymously (HTTP 200). Failed responses and limits are preserved rather than relabeled as acceptance.


timothy-le7:
Independent exact-head review at `1114d668e` (matches the review request, so this is attributable). Baseline 124 passed.

**Audit-only holds.** I checked it structurally rather than trusting the label: `visual_review.py` makes no assignment to score, gate, threshold or pass state; the core symbols it uses are all read-only or constructive (frame selection, evidence builders, `_sha`, HTTP post); and `review_visual(` has no call site in `npa/src/npa` outside its own definition and the SDK wrapper. It is reachable only through the separate `review-visual` CLI command, never from `run`/`loop`. The separation is real.

**Fail-closed is three-for-four.** Neutralizing individual refusals and counting kills:

| guard | line | kills |
|---|---|---|
| strict schema field match | 1176 | **18** |
| neutral frame marker order invalid | 1747 | 0 |
| paired requests differ beyond image ordering | 1754 | 0 |
| fidelity status requires empty issue list | 897 | 0 |
| `deployment_status == "audit_only"` | 810 | 0 |
| `score_gate_affected is False` | 811 | 0 |
| `normalized_task_completion_score is None` | 812 | 0 |
| whole-report binding `report != expected` | 801 | 0 |

Malformed schemas are strongly defended. **Order disagreement is not** — both order guards can be neutralized with every test still green, and they are the mechanism behind the paired A/B reversal the design rests on.

On 810-812: line 801 rebuilds the expected report and compares the whole object, which subsumes those field checks, so I am not filing them as a plain gap. But 801 is itself unexercised, so it is two undefended layers rather than one defended one. Good layering, no test behind either.

**Limit:** 0 kills means no test exercises that refusal, not that the guard is wrong. I read all four and they look correct; this is absence of proof, not evidence of a defect. Given a reversed-order paired review is the core of B030, a test that reverses image order and asserts the refusal seems worth having before this moves from retained negative to qualification.

timothy-le7:
Remediation verified independently at `c958ac2a2` (matches live head). I reran exactly the four mutations from my earlier review; baseline is now 126 passed.

| guard | before | after |
|---|---|---|
| `neutral frame marker order is invalid` | 0 | **1** |
| `paired requests differ beyond image ordering` | 0 | **1** |
| `fidelity status requires an empty issue list` | 0 | **1** |
| `visual review report is not internally bound` | 0 | **1** |

Four for four — the gap is closed. The whole-report binding regression is the one I most wanted: forging only `attempt_count` slips past the explicit field predicate and has to be caught by the binding comparison, which is exactly the layered case that previously had no test behind either layer.

One clarification for the record rather than a complaint. The note describes four direct regressions; the diff adds two new test functions and delivers the other two as `pytest.raises` assertions inside existing tests, which is why the file moved 124 → 126 rather than 128. Coverage is real either way since I verified by mutation, but someone reconciling "four regressions" against "+2 tests" later would otherwise think something was missing.

No remaining findings from me on this head.

PR #648, opened 2026-09-20
Add remote development desktop and shared mobile Codex chat

timothy-le7:
## Summary

Add a persistent Linux development desktop and a mobile Codex interface under `npa tools desktop`. The desktop runs full VS Code and Codex on an existing VM. Its noVNC viewer requests device pixels, uses lossless Tight encoding, and offers workspace sizing, clipboard, and full-screen controls for readable Retina rendering. Setup disables XFCE compositing and GTK animations; the new `optimize` command applies those preferences to a running desktop without changing resolution or restarting applications.

The new `chat-setup --connect-vscode` command serves `/chat/` through the desktop's authenticated HTTPS gateway. Phone and desktop users can search sessions, read paginated history, send prompts, steer or stop a running turn, and answer supported approvals and questions. VS Code and the browser share one Codex runtime through a private Unix socket. The iPhone layout includes safe-area and keyboard handling, while reconnect recovery preserves the selected conversation and duplicate submission protection prevents repeated prompts.

Public access uses a dedicated gateway, trusted short-lived IP certificates, and automatic renewal. Desktop, chat, assets, history, and controls require authentication; VNC, websockify, and the chat backend stay on loopback. No VPN is needed when the operator supplies a publicly routed external address. Credentials remain in private runtime files and are never embedded in web assets or URLs. Browser mutations require the configured HTTPS origin.

Setup preserves existing credentials and running applications. Sessions held by older independent Codex clients are readable but cannot be controlled until released by those clients. Reload the connected VS Code window after its work finishes. Specialized MCP and dynamic-tool requests remain in VS Code. This integration uses the experimental Codex app-server API and development executable setting; both clients need revalidation after extension upgrades.

The implementation lives in `npa.tools.desktop`, with operator documentation and a matching skill. Cloud address allocation and backup repository provisioning remain explicit operator steps.

## Verification

- Desktop and mobile HTTP/control tests: 44 passed. Onboarding smoke: 114 passed.
- Full default suite, including guardrails: 24,144 passed, 154 skipped, and one xpassed in an isolated Python 3.12 environment.
- Required security regressions: 799 passed in an isolated Python 3.12 environment with the official CPU Torch 2.13.0 wheel.
- Ruff, JavaScript syntax, generated CLI documentation drift, wheel build and inclusion of all chat assets passed.
- Diff-scoped confidentiality and Gitleaks passed across the complete PR commit range. No live infrastructure identifiers, credentials, private endpoints, session exports, or screenshots are included.
- Live mobile validation: iPhone-sized Chromium and WebKit rendered without horizontal overflow or JavaScript errors. Two browser clients exchanged real prompts and replies in the same session. The IDE transport received completion after mobile steering and interruption. Restarting the web service preserved the Codex engine and automatically recovered the selected chat. Duplicate submission protection was exercised with a delayed response.
- Live gateway validation: trusted TLS; unauthenticated protected routes returned 401; authenticated routes returned 200; foreign origins were rejected. Direct public access was verified using the physical network interface without the overlay network.
- Desktop validation: repeated setup preserved the session and credentials. Retina rendering used device-pixel framebuffers with selectable workspace sizing. A 24-input click/key probe measured median input-to-frame delay falling from 84 ms to 66 ms after disabling compositing and GTK animations (95th percentile: 96 ms to 76 ms). Lower compression did not improve the result, so lossless encoding and the current resolution were retained. The desktop process remained running and the original application focus was restored after the probe.
- Recovery validation: a full disk snapshot reached READY; hourly encrypted backups include consistent Codex SQLite copies. Independent selected-file restores matched hashes and the restored session database passed its consistency check. A complete replacement-VM boot was not rehearsed.

## Skill Maintenance

Added the development-desktop skill and index smoke entries, documented the mobile setup and legacy-client migration, updated CLI namespace guidance, and linked the operator guide and generated command reference.


timothy-le7:
Independent security review of the chat gateway at head `d11d67a64`, focused on the exposure surface a remote desktop plus mobile chat introduces.

**The authentication is well built.** `_authorized` requires the `Basic ` scheme, decodes with `validate=True` so malformed base64 is rejected rather than coerced, and compares with `hmac.compare_digest` rather than `==`. The server binds `127.0.0.1` only. I probed it with nine adversarial inputs rather than reading it — empty header, `Basic ` with no payload, lowercase scheme, non-base64 payload, embedded whitespace, a correct prefix that is one character short, and a correct password with a trailing character — and all nine were refused, with the correct credential accepted. `_write` uses `0o600` and an atomic temp-then-`replace`, and the credentials come from `secrets.token_urlsafe`.

Two things worth fixing, neither a live vulnerability:

**1. No emptiness guard on the password.** `_guard` does `password = Path(config["password_file"]).read_text().strip()` with no check that the result is non-empty. If that file is ever empty or whitespace-only, `Basic ` + base64 of `username:` authenticates — I confirmed `_authorized(header, "alice", "")` returns `True` for exactly that input.

I want to be straight about reachability, because it is low: the file is written atomically via temp-then-replace, so a truncated write cannot produce an empty file, and a missing file raises rather than authenticating. It takes operator error or external corruption to get there. But the guard is one line, the failure is silent, and it converts a corrupted state into an open door rather than a closed one.

**2. `_write` chmods after writing.** The order is `temporary.write_text(content)`, then `temporary.chmod(mode)`, then `replace`. Between the first and second call the temp file exists with the process umask applied — potentially world-readable — while already containing the credential. Opening with `os.open(path, O_WRONLY|O_CREAT|O_EXCL, mode)`, or chmodding the empty file before writing, closes the window. This matters more here than in most places because `_write` is the function used for the password and keyring secrets specifically.

Also noting an asymmetry you may have intended: `secrets.token_urlsafe(6)` at line 80 versus `token_urlsafe(24)` at line 401 and `token_hex(32)` at line 81. 48 bits is not a practical online-guessing risk behind a loopback bind, so I am not flagging it as weak — but the inconsistency is unexplained in the code, and line 469 refers to a `public-password`, so it is worth confirming the 6-byte one is not the internet-facing credential.

No blocking findings.

PR #650, opened 2026-09-20
fix(sim2real): gate submit preflight on all sim2real gated HF repos

timothy-le7:
## Problem

While driving the canonical Sim2Real workflow (`workflows/main/sim2real.yaml`) toward a live run, the submit preflight's Hugging Face gate was found to be a **premature green**. `static_prerequisites` in `sim2real_preflight.py` hardcoded a single repo:

```python
repos = ["nvidia/Cosmos-Transfer2.5-2B"]
```

But the Cosmos Transfer stage (Stage 3) also fetches, at runtime:

- `nvidia/Cosmos-Predict2.5-2B` — the pinned Wan2.1 VAE tokenizer (`tokenizer.pth`), gated
- `nvidia/Cosmos-Guardrail1` — the guardrail weights, gated

Both are already registered for the `sim2real` capability in the access catalog (`WORKBENCH_ASSETS`) and are checked by `npa workbench health access --capability sim2real`. The submit preflight, however, did not check them. An operator who accepted only `Cosmos-Transfer2.5-2B` would pass submit preflight, warm the multi-GB Isaac cache PVC, provision RT-core GPUs, and only then fail inside Stage 3 on the unaccepted dependency. This matches the disposition of finding #4 in `docs/workbench/sim2real-live-findings-evaluation-20260902.md`, which was fixed in the access catalog but not in the submit preflight.

## Fix

Derive the preflight's gated HF repo list from the sim2real capability's single source of truth (`gated_hf_repos(("sim2real",))`, deduped, order-preserving) instead of a hardcoded list, so the pre-launch gate matches `health access --capability sim2real`. A defensive fallback to `Cosmos-Transfer2.5-2B` remains if the catalog ever returns nothing.

## Tests

- Updated `test_static_preflight_checks_hf_models_and_hosted_model_before_submission` to assert the gate mirrors the capability catalog and includes Transfer2.5, Predict2.5, and Guardrail1.
- Added `test_static_preflight_surfaces_unaccepted_transfer_dependencies` as a regression guard: an operator with Transfer2.5 accepted but Predict2.5 unaccepted now gets a pre-launch issue instead of a Stage 3 failure.

`tests/orchestration/npa_workflow/test_sim2real_preflight.py`, `tests/workbench/test_model_access.py`, and `tests/workflows/test_sim2real_health.py` pass (122 tests); `ruff check` is clean.

[...]

PR #651, opened 2026-09-20
docs(sim2real): record Token Factory evaluator decision (NVIDIA TF models are text-only)

timothy-le7:
## Decision

While driving Sim2Real toward a live run, I evaluated "use the latest NVIDIA tooling on Token Factory" for the Stage 8/9 rollout evaluator. Stage 8 sends primary-frame images plus an ordered JSON Schema `prefixItems` response contract, so the evaluator **must be a vision-language model with strict structured output**.

I probed the user's Token Factory key with the evaluator's actual request shape (base64 `image_url` + `prefixItems` `json_schema`):

| Model | Vision + structured output |
| --- | --- |
| `MiniMaxAI/MiniMax-M3` | ✅ OK |
| `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B` | ❌ `400 This model does not support image input` |
| `nvidia/Nemotron-3_5-Lightning` | ❌ `400 ... does not support image input` |
| `nvidia/nemotron-3-super-120b-a12b` | ❌ `400 ... does not support image input` |
| `nvidia/Nemotron-3-Ultra-550b-a55b` | ❌ `400 ... does not support image input` |
| `openbmb/MiniCPM-V-4_5` | ✅ OK |
| `google/gemma-3-27b-it` | ✅ OK |

**Every NVIDIA model on Token Factory is text-only** and cannot serve as the Stage 8 evaluator. The sound hosted choice is `MiniMaxAI/MiniMax-M3` (already the workflow default, verified VLM + structured output). NVIDIA's own VLM evaluator, `nvidia/Cosmos-Reason2-8B`, is a **self-hosted GPU** path (Cosmos Reason), not a Token Factory hosted model.

The code is already architecturally aligned: `npa/src/npa/workflows/sim2real/constants.py` sets `DEFAULT_COSMOS3_MODEL = "MiniMaxAI/MiniMax-M3"`, the renderer/preflight fallbacks import that, and `hosted_rollout_model_family` fails closed on unsupported IDs — so selecting a text-only NVIDIA model is rejected before launch.

## Change

Documents this decision in the operator runbook (`docs/workbench/guides/sim2real-workflow.md`, §1) so operators don't spend time trying an NVIDIA text model that cannot see frames, and records the self-hosted Cosmos-Reason2 alternative.

## Context

Docs-only. Companion to the preflight fix in #650. The remaining gates to a full live run are a coherent current-SHA image set (the published Sep‑4 set predates MiniMax and fails the baked Stage‑1 evaluator probe; a controller image built from current `main` passes it) and RT‑core RTX PRO 6000 capacity.
## CI-only maintenance addendum

The original source/documentation candidate remains unchanged. At exact head
`49198dce624e07d47b5205182941f613d1bc4045`, Linux CI completed 5,022 tests
before one test-helper observation race: after the owned process leader had
already exited, `/proc/<child>/stat` vanished during `Path.read_text()` and
raised `ProcessLookupError` (`ESRCH`). The helper already treated
`FileNotFoundError` (`ENOENT`) as an already-reaped child.

Current head `fd4e9cd3fbcb171358cbd833fe8d3b3b8798503e` changes only that test
helper and its regression control. It catches exactly the two
disappearing-process exceptions, keeps permission and all other errors visible,
and retains live-child rejection. A deterministic ESRCH sentinel plus the real
Linux resistant-child reaper check pass as part of all 22 tests in the file;
Ruff check and format also pass.

This is a CI-observation repair, not a production-behavior or documentation
change. Exact amended-delta independent review and refreshed CI are running.


timothy-le7:
Independent review at head `49198dce6`. **Every checkable claim in this doc is accurate — I verified each against the live API rather than reading it.**

| claim | result |
|---|---|
| the NVIDIA models on Token Factory are exactly those four Nemotron-3 IDs | **exact match** — 4 `nvidia/*` models, none missing, none extra |
| a multimodal request returns `400 {"detail":"This model does not support image input"}` | **reproduced verbatim** on Nano-30B-A3B and Nemotron-3_5-Lightning |
| `hosted_rollout_model_family` fails closed on those IDs | **confirmed** — `CosmosReasonError: unsupported hosted rollout evaluator`, while `MiniMaxAI/MiniMax-M3` resolves to `minimax_m3` |
| MiniMax-M3 is available on the key | confirmed |

Documenting a negative result like "the obvious-sounding NVIDIA choice cannot do this job" is more useful than most docs changes, and getting the error string exactly right is what makes it searchable when someone hits it.

**One caveat worth a sentence**, from this program's own retained evidence. The doc calls MiniMax-M3 "the sound hosted choice," which is true for the property you state — it accepts base64 frames and returns the ordered `prefixItems` schema. But `docs/workbench/evidence/vlm-missing-terminal-control-review.md` records MiniMax-M3 being rejected for *rationale quality*, on misgrounded rationales for controls C02 and C03. Those are different properties, and a reader choosing a Stage 8 evaluator could reasonably read "sound choice" as covering both. Something like "schema-compliant; see the terminal-control review for rationale-quality caveats" would keep the two separate.

Not a blocker.

timothy-le7:
Follow-up review of the maintenance head `fd4e9cd3f`, against my earlier acceptance at `49198dce6`.

**First, my earlier acceptance still binds.** The delta between the two heads is a single test file, +23/-1, and `docs/workbench/guides/sim2real-workflow.md` is byte-identical. The live-API verification I published (model list, verbatim `400` reproduction, fail-closed resolver) therefore carries forward to this head unchanged — no re-verification needed.

**The added delta is test-only**, and it is a regression test for behaviour that already exists: `phase_liveness.py` already caught `ProcessLookupError` at three sites before this change. So this pins down the ESRCH-during-`/proc`-read race root reported rather than altering it, which is the right shape for a maintenance commit.

**One observation on how far the new test reaches.** Replacing each `except ProcessLookupError:` with an unrelated exception type, against a git-clean 22-test baseline:

| catch site | result |
|---|---|
| `phase_liveness.py:170` | **1 test fails** — this is what the new test binds |
| `phase_liveness.py:162` | 22 passed, unbound |
| `phase_liveness.py:178` | 22 passed, unbound |

So the specific race that was reported is now defended, and the two sibling catch sites in the same family are not. They are the same hazard — a process exiting mid-read — so if one was worth a regression test, those two probably are too, and they would be cheap additions alongside the fixture that already exists.

Not a blocker, and nothing here changes the acceptance.

Method note: my first pass at this mutated `orchestration/skypilot/workflow.py`, which has visually identical catch lines but is not the module under test; those runs were void and I discarded them. The table above comes from mutating `npa/src/npa/workbench/isaac_arena/phase_liveness.py`, asserting a git-clean tree before each mutation.

timothy-le7:
**Marked ready for the required merge queue at fd4e9cd3fbcb171358cbd833fe8d3b3b8798503e.** Exact-source review is accepted and all 22 checks are terminal: 19 succeeded and 3 conditional checks were skipped. Each check was verified against this commit and GitHub Actions app 15368. Required gitleaks, scan, and security-regression checks passed.

The retained Linux scoped run passed 22 tests, including the real child-process cleanup and ESRCH regression control. The five current CI shard logs report 25,278 passed, 657 skipped, and 1 xpassed, with coverage green. The final source delta changes only the test helper and its control; production code and the reviewed documentation are unchanged.

The PR description now separates point-in-time model support from evaluator quality, limits independent hosted rejection evidence to the two actually reproduced probes, and retains the MiniMax rationale-quality caveat. This readiness audit made no new provider calls and claims no GPU/VLM workload acceptance.

[Current CI run](https://github.com/nebius/nebius-physical-ai/actions/runs/35556823766). The required merge queue still has to run integration checks. No merge or enqueue was performed.

PR #653, opened 2026-09-21
fix(workflow): redact durable diagnostics

timothy-le7:
## User outcome

Provider exceptions, subprocess output, stage logs, cleanup output, and durable
launch records no longer retain common credential shapes after they pass through
the shared workflow-state redaction boundary. Recovery guidance, line structure,
legitimate workflow-token references, and ordinary token counters remain useful.

This is stacked on [#623](https://github.com/nebius/nebius-physical-ai/pull/623),
which owns workflow CLI failure-boundary sanitization. This PR extends the same
behavior to lower-level durable and log-facing workflow records.

## Current exact source

- Head: `c3f53eaa755a8ac0c89981d620b67499db2ec302`
- Tree: `9da2c242ed58ad859041786a8ff28548927e2c69`
- Stack base: `525007e2a431608d3f80185be7ac842f49b691bf`
- Original evaluation-plan SHA-256:
  `e6198f743dadcfefbbafb583b41b5e8b7a608ce3106cb25644ad3fee30518cf6`
- Standalone AWS-signature remediation-plan SHA-256:
  `7264eb94a3daf7fcc3c1c371d8043e465861db6dda22bc3ff3f2890e3e9ec2a3`

## What changed

- Centralize diagnostic redaction in a dependency-light module shared by
  workflow-state and verification/reporting code.
- Cover bearer values, URL userinfo and query strings, complete or truncated
  PEM private-key blocks, compound secret-assignment keys, known token shapes,
  and exact opaque values supplied at a credential-aware boundary.
- Redact standalone `X-Amz-Signature=` and `x_amz_signature=` values even when
  a durable subprocess diagnostic does not include the surrounding URL.
- Preserve same-line recovery context, legitimate
  `unknown <scope> token: <scope>.<name>` diagnostics, plural token counters,
  noncredential `request_signature=...` values, idempotence, and line endings.
- Use a linear fail-closed PEM scanner; repeated unterminated headers no longer
  trigger quadratic work.

## Objective evidence

The original baseline redacted only 1/6 frozen credential formats. Prior
accepted head `7e33da910371462b7ff87fc051cb63f315798fef` redacted 6/6 and passed:

- 86 focused and 666 affected workflow controls;
- 12/12 cache-isolated exact mutation reversals;
- a 4,058,000-byte unterminated-PEM stress case in 0.055 seconds with line
  structure preserved;
- 114 smoke, 3,863 guardrail, and 791 security tests, with eight
  dependency-based security skips;
- the full automatic suite.

A later independent audit found that all durable `redact_text` callers rely on
pattern matching rather than supplying explicit opaque secrets. At the prior
head, standalone hyphenated and underscored AWS signature assignments leaked,
although the same key inside an S3 URL was already redacted.

On the current head:

- all four standalone/quoted AWS-signature controls redact while preserving
  same-line recovery context;
- existing signed-URL redaction remains intact;
- `request_signature=sha256:...` remains unchanged;
- removing the exact AWS-signature key classification re-exposes every
  standalone control while the independent URL path remains green;
- 91 focused tests, 49 confidentiality/secret/hygiene guardrails, and 248
  adjacent presigned-signature controls pass;
- Ruff, format, and diff checks pass;
- a fresh exact-head independent review accepted the remediation with zero
  findings after reproducing the distinguishing mutation and durable
  cancellation-record paths.

## Applicability

| Field | Current evidence |
| --- | --- |
| Source / image identity | Exact head/tree above; no container image or workflow spec changes |
| Actual workload | Host-CPU production record serializers and diagnostic transformation with synthetic credential-bearing provider output |
| Negative controls | Shared-boundary bypass, format matchers, quoted context, repeated malformed PEM, compound credential keys, token-counter over-redaction, and exact AWS-signature classifier removal |
| GPU / hosted API / live provider | Not applicable; no provider request, infrastructure mutation, or GPU execution |
| Visual / VLM | Not applicable; no image, video, geometry, or RRD output |
| Independent review | Exact current remediation accepted with zero findings |
| Automatic CI | Exact current head: 20 succeeded, 2 scope-based skips, 0 failed/pending; clean and mergeable |
| Remaining blocker | Merge order remains behind #623 |

## Limits

Pattern-only redaction cannot safely identify every bare opaque value without a
credential-aware caller supplying it through `secrets=`. In particular, this
delta does not claim generic recognition of arbitrary bare AWS secret values,
JWTs, or isolated key-body base64. It closes the credential-equivalent
standalone AWS signature shape without adopting a broad `signature` substring
rule that would erase ordinary diagnostics.


timothy-le7:
Independent review at head `7e33da910` (detached copy). **Everything you claim, it does. The gap is in what the durable path can reach.**

Every shape in the stated scope redacts correctly — verified by executing `redact_diagnostic_text` directly: HF tokens, NVAPI keys, GitHub PATs, `AKIA`/`ASIA` access key IDs, `Bearer` tokens, private-key blocks, URL-embedded credentials, and `AWS_SECRET_ACCESS_KEY=` assignments. The `secrets=` mechanism also works properly, redacting a bare value and a value embedded as a URL query parameter.

**But nothing in the durable path uses it.** `workflow_state.redact_text` defaults `secrets` to `None` → `()`, and across `npa/src/npa` there are **64 `redact_text` call sites and zero pass `secrets=`** — including the ones this PR is about (`cancel_stdout`, `cancel_stderr`, `down_stdout`, `down_stderr`, `status_after_down`). So for durable cleanup and cancellation records, redaction is purely prefix-pattern based, and credential shapes without a distinctive prefix survive into the record:

| shape | result |
|---|---|
| bare 40-char AWS secret value | left in clear |
| raw JWT (`eyJ…`) | left in clear |
| `X-Amz-Signature=…` from a presigned URL | left in clear |
| ssh key base64 body | left in clear |

The presigned-signature case is the one I would fix first: it is credential-equivalent, it plausibly appears in storage-related subprocess output, and it lands in a durable record rather than a transient log line. A JWT in provider CLI output is similarly plausible.

Two ways to close it — add `X-Amz-Signature`/`X-Amz-Credential` and a JWT pattern to `_KNOWN_TOKEN_PATTERNS`, or thread the credentials already resolved at the call boundary into these cleanup calls the way `verification.py` does. The second is the more durable fix, since `verification.py` already demonstrates the pattern.

To be clear on scope: I am not claiming a leak exists today in any specific run, only that this boundary does not stop these shapes. Not a blocker for the behavior you describe.

timothy-le7:
Independent verification of head `c3f53eaa755a8ac0c89981d620b67499db2ec302`, tree `9da2c242ed58ad859041786a8ff28548927e2c69` — **tree hash matches your claim exactly**. Delta against the prior accepted `7e33da910` is three production lines and 31 test lines. **All five requested checks pass.** Accepted.

Redaction behaviour, verified by execution rather than by reading:

| input | result |
|---|---|
| `X-Amz-Signature=`, `x_amz_signature=` | redacted |
| quoted and single-quoted mapping keys | redacted |
| same-line recovery context around the value | redacted |
| upper-case `X-AMZ-SIGNATURE=` | redacted |
| signed URL | redacted |
| `request_signature=sha256:...` | **visible**, correctly |
| `signature_version=v4`, prose `image signature verified ok` | **visible**, correctly |

Using exact key equality rather than a `signature` substring is the right call, and it is what makes the three non-credential cases come through untouched. Removing the `_SECRET_EXACT_KEYS` check fails 4 tests against a git-clean 91-test baseline, so the new control is genuinely bound and not merely exercised.

**Scope of what this resolves from my `20260921T022357Z` finding.** This closes the `X-Amz-Signature` shape completely. Since the redaction is key-name-based and unconditional, it applies on the durable-record path too, where call sites pass no `secrets=` — which was the concrete half of the finding. Two shapes I raised remain open, and they are not equally hard:

- **raw JWTs survive** (`eyJhbGciOi...` passed through intact). These are cheaply pattern-detectable from the `eyJ` header prefix, so this looks like the same three-line treatment.
- **bare AWS secret values survive** (`wJalrXUtnFEMI/K7...`). This one is structural, not a missing pattern: a naked secret has no distinguishing token, so no pattern can catch it. Closing it means threading resolved credentials into the cleanup call sites the way `verification.py` already does.

Worth splitting those two apart in the backlog rather than tracking them as one item, since the first is a pattern addition and the second is a plumbing change. Neither belongs in this PR.

For completeness: standalone `X-Amz-Credential=AKIA.../20260921/...` redacts the AKIA key via the existing pattern and leaves the credential scope visible, which is correct — the scope is not secret.

timothy-le7:
Signature repair is published at `c5d15e942a9b981b969ed17cf8b549dd51e01c56`. GitHub verifies every replacement signature across PRs #615, #623 and #653. All ten replacement commit trees, parent-relative diffs, original authors and messages are preserved. The parent/child merge trees and PR diffs are also byte-identical, including the newer main changes already reviewed for #623. Original commits remain in a private recovery bundle.

This repair changes no source behavior. Local confidentiality and Gitleaks checks passed. Fresh hosted CI is required before the previous readiness verdict carries forward; no merge or queue entry has been performed.

PR #654, opened 2026-09-21
Harden agent metadata credential request paths

timothy-le7:
## Problem

The agent could synchronously probe an unhealthy metadata/config filesystem before a nominally bounded cloud command. Python's normal subprocess timeout cleanup can also wait after signalling a stuck child, so a request deadline did not guarantee a bounded request thread.

This cross-cutting runtime fix was removed from the RoboCasa policy change in [PR #595](https://github.com/nebius/nebius-physical-ai/pull/595) and is intentionally reviewed separately.

## Change

- select only an allowlisted staged credential source; metadata mode pins the exact root `cursor-sa` profile/config and scrubs ambient token, token-file, and endpoint overrides
- remove request-time metadata/config existence probes from direct and rendered agent access paths
- run request-time Nebius, NPA, Kubernetes GPU-preflight, resource-discovery, and legacy Sim2Real child commands in new process groups
- enforce finite deadlines on request probes while preserving the existing uncapped `timeout_s=None` contract for confirmed durable workflow submission
- on timeout, return without synchronously waiting, kill the owned process group, keep a temporary breaker open, and reap asynchronously
- retain the unreaped leader while checking descendant group membership, preventing PID/PGID reuse from targeting an unrelated process
- fail closed when credential provenance is missing, and prevent operation-specific environment overlays from changing the staged source

## Applicability and evidence

| Boundary | Applicability | Current result |
| --- | --- | --- |
| Source | exact head/base | head `09d0fa9f3693db57142b82d7d7caf192b72257e0`; base `99be7d39f4481aa2c29f72204175698da1a7a258`; three agent runtime files plus five focused test files only; the repair after `0666db15` changes two test files and no production bytes |
| CPU/host execution | required | 453 direct and 132 rendered-backend tests passed on the runtime change; the exact test-only repair adds passing fault injection for the shared `subprocess.TimeoutExpired` → built-in `TimeoutError` boundary, owned-process abandonment, and both user-visible callers |
| Smoke | separate, not acceptance | 114 onboarding smoke tests pass |
| Repository/security | required | 4,083 guardrails pass; security regression passes 791 with eight expected skips; Ruff, diff check, built-in confidentiality, and Gitleaks pass |
| Browser | required for agent UI | mocked Cypress suite passes 32/32 |
| GPU | not applicable | no GPU behavior changes |
| Hosted VLM/API | not applicable | nonvisual runtime hardening; no visual claim |
| Failure controls | required | missing/unknown provenance, poisoned profile/config/token/endpoint selectors, timeout, concurrent timeout, descendant process group, pipe-close failure, and stale-cleanup breaker paths are covered |
| Independent review | required | three review tracks approved the corrected design; root independently reproduced the live-proof commit/tree and all four implementation hashes, verified the auth-negative and cleanup receipts, and recorded CPU live proof in the acceptance matrix |
| Hosted CI | required | refreshed for exact head `09d0fa9f`; pending |
| Live workload | required for readiness | PASS on a fresh CPU agent: healthy and basic-auth-protected; metadata-backed access refresh HTTP 200 in 3.64 s; resource inventory HTTP 200 in 5.66 s; exact-name teardown reported verified deletion and infrastructure absence; the local agent record is absent |

CI classifies this as affected `npa/tests/cli` plus browser coverage, not an early full-suite change.

## Live CPU proof

Runtime head `0666db15` deployed one fresh agent-only CPU VM through the normal NPA lifecycle. The current `09d0fa9f` repair changes tests only; a direct Git diff confirms every production byte under `npa/src` is identical, so no second deployment is needed or permitted for this CI-test correction. The protected live access route reported `instance_metadata`, profile `cursor-sa`, and config `/root/.nebius/config.yaml`; the live resource route returned a non-error inventory. Raw endpoint, inventory, deployment, and provider outputs remain private. Their owner-only files are content-hashed in a sanitized proof summary. Cleanup passed only after the destroy result reported `verified_deleted`, `verified=true`, and `infrastructure_absent=true`; a separate post-run check confirmed the exact local agent record is absent.

This proves the changed credential-selection and bounded cloud-command paths work on a fresh deployed agent. Independent root review reproduced the source bindings and recorded the live CPU result, while retaining one explicit limitation: the in-path-verified live deployment manifest body was not copied before teardown. The test-only repair was independently approved and its two new regression nodes plus Ruff checks pass locally. It does not replace timeout fault-injection coverage, establish GPU behavior, or make a visual claim. Exact-head hosted CI remains pending, so the PR stays draft.

## Limits

- Successful commands are expected not to daemonize detached descendants; timeout cleanup contains same-session process groups.
- Inventory can still take the sum of several individually bounded probes.
- This changes agent runtime safety only; it contains no RoboCasa or policy behavior.


timothy-le7:
[Independent exact-source CPU proof and review](https://github.com/nebius/nebius-physical-ai/blob/faf4e8402832d1803eb746c00db02ff1cee91d84/docs/testing/evidence/agent-metadata-cpu-0666db15/README.md) are now accessible without credentials for `0666db15f7c5a715ceba74b9dbfa57fe93054e26`.

**Accepted scoped proof, zero review blockers:** 74 resource/access tests and 132 rendered backend tests passed, plus a real Linux child/descendant timeout and recovery control. The normal fresh CPU deployment returned authenticated access/resource HTTP 200 and anonymous UI HTTP 401; the retained teardown receipt reports verified deletion. The downloadable reproduction control and hashed measurements are included.

Limits remain explicit: no retained independent deployed-Python byte attestation, no separate anonymous API-route probes, partial access only, shared IAM retained, and no GPU/VLM claim. Current-head hosted CI and draft promotion were incomplete at preparation; this comment does not declare a merge.

timothy-le7:
Independent source review at head `0666db15f7c5a715ceba74b9dbfa57fe93054e26` — which matches the proof commit exactly, so the live-CPU bundle binds this code. Per the policy receipt I re-ran no provider query and no redeploy. The proof bundle itself I verified separately (receipt hash, commit/tree, and all four `implementation_hashes` byte-reproduce).

**The implementation is correct. My finding is that the two subtlest security properties are not bound by any test.** Mutation against a git-clean baseline of all 448 tests in the four changed test files:

| mutation | meaning | result |
|---|---|---|
| drop `WNOWAIT` from the ownership probe | leader gets reaped, so its PID/PGID becomes reusable — re-opens exactly the hazard the PR says it closes | **448 passed, 0 kills** |
| `return uncertain` → `return False` in `_agent_process_group_has_other_members` | fail *open* when `/proc` parsing fails, instead of conservatively assuming members remain | **448 passed, 0 kills** |
| drop the `pid != leader_pid` exclusion | leader counts as its own descendant | 1 kill (defended) |

The `WNOWAIT` one is worth dwelling on. The comment `# WNOWAIT proves this PID is still our child without freeing it for reuse.` states the invariant precisely, and the code is right — but nothing enforces it. A future cleanup that “simplifies” the flags to `WEXITED | WNOHANG` would reintroduce a kill-the-wrong-process bug with fully green tests. Worth noting that the evidence table lists PID-ownership controls as passing: a control does exist, it just does not bind this property. The mutant is not even behaviourally silent — it ran 15:03 against the 11:14 baseline, so the reaping change is observable, merely unasserted.

Credential isolation I checked by execution rather than reading, with a poisoned environment:

- `instance_metadata`: **zero survivors**. Profile, config dir and `HOME` pinned to `cursor-sa` / `/root/.nebius` / `/root`. Exactly as claimed.
- missing, unknown, and wrong-case (`INSTANCE_METADATA`) sources are all refused — no case-insensitive bypass.
- `configured_profile`: tokens and endpoint are scrubbed, but ambient `NEBIUS_PROFILE`, `NEBIUS_CONFIG_DIR`, `NPA_NEBIUS_CONFIG`, `NPA_NEBIUS_PROFILE` and `HOME` pass through unchanged. This is consistent with your claim, which is explicitly scoped to metadata mode, so it is an observation and not a contradiction. The one I would think about is `HOME`: redirecting it points the CLI at an attacker-controlled config tree even when the profile name is legitimate.

Suggest a test asserting the flags include `WNOWAIT`, and one forcing a `/proc` read error to assert the conservative `True`. Neither blocks merge.

Method note: my first mutation pass was invalid — the venv editable install resolved `npa.cli.agent_resources` to the main worktree, so the mutations never loaded, and a later run left the tree carrying a mutation. Both were discarded; the numbers above come from a rerun that asserts a git-clean tree before each mutation and verifies the mutation is on disk.

timothy-le7:
Independent review of the amended delta `0666db15f` → `09d0fa9f3` (root-assigned).

**The test-only premise is proven, not just asserted.** That premise is what lets the existing live proof stay truthful without a redeploy, so I verified it by bytes rather than by reading the file list:

- diff restricted to `npa/src`: **0 lines**
- non-test paths touched: **0**
- SHA-256 of each production file, identical at both heads: `agent.py` `140c5cbf2d84`, `agent_resources.py` `19438686adad`, `agent_access_runtime.py` `49df5d6dd66f`

That is the byte-equivalence binding requested — the live-CPU proof produced at `0666db15f` carries to `09d0fa9f3` because the production bytes are unchanged, with no rewriting of the original execution identity.

**One thing to keep visible:** this repair does not touch the gap I reported on the parent head. The delta contains zero references to `WNOWAIT`, so the PID/PGID-reuse defense and the `/proc` fail-closed path remain unbound by tests (0 kills each against the 448-test baseline). That is entirely appropriate — the repair was scoped to the timeout-test defect — but it should not be assumed closed by this commit.

No findings on the delta itself.

timothy-le7:
The test-only repair at `09d0fa9f3693db57142b82d7d7caf192b72257e0` has passed independent review and 23 scoped tests. GitHub comparison and anonymous source downloads verify that all production files remain identical to the live CPU producer `0666db15f7c5a715ceba74b9dbfa57fe93054e26`.

[Source-equivalence proof, hashes, review scope and retained limitations](https://github.com/nebius/nebius-physical-ai/blob/b83b862e1469d0bfa22ae7c8bffae11db6b1c9b3/docs/testing/evidence/agent-metadata-test-repair-09d0fa9f/README.md) connects this repair to the existing live deployment proof. This is not a new deployment run. The two nonblocking mutation-test coverage gaps remain disclosed.

Fresh CI is still pending; no merge-readiness claim or merge is made by this evidence update. GPU/VLM evaluation does not apply to this CPU process-management repair.

PR #655, opened 2026-09-21
Align PR and merge-queue checks while starting tests sooner

timothy-le7:
## Summary

Code PRs could pass `security-regression` with only selected subsystem tests, while the merge queue added full coverage, Cypress, and Python 3.10/3.14 compatibility checks. Run those same checks before PR admission and retain the conservative prose-only exception. Remove the redundant affected-tests path and use the trusted base's queue policy for both events, including during rollout from an older base.

Start the short test-scope selector in the existing scope/completion slot instead of behind long metadata/docs/guardrail work. In [a recent successful queue run](https://github.com/nebius/nebius-physical-ai/actions/runs/35548460834), that 13-second selector started about 18 minutes after the run began. This preserves the existing runner pools, duration-balanced shards, xdist, dependency caches, merge priority, 60% coverage floor, and all security checks. It does not reuse PR results for a different combined queue commit.

New base changes, interactions between queued PRs, and flaky tests can still fail a combined candidate. This change closes the deliberate PR/queue coverage gap; it does not suppress those failures or automatically requeue them.

## Verification

- [x] 870 focused CI, documentation, and skill tests passed; the rollout regression executes the actual workflow shell step against a legacy trusted selector.
- [x] Ruff and CI dependency-pin checks passed.
- [x] Real scanner regression fixtures passed; full base/candidate security comparison reported **0 new findings**.
- [x] Staged confidentiality and gitleaks scans passed.
- [x] Actionlint passed with only its unsupported, already-used `concurrency.queue` key excluded.
- [x] [Hosted test selection](https://github.com/nebius/nebius-physical-ai/actions/runs/35552111172/job/106188644871) started two seconds after the workflow began and passed in 13 seconds. All five coverage shards and browser/compatibility checks were requested; duplicate prose smoke was skipped.
- [ ] Remaining hosted Linux Python 3.10/3.12/3.14, coverage, and security results are pending in [the PR run](https://github.com/nebius/nebius-physical-ai/actions/runs/35552111172).

The macOS full-suite attempt was interrupted after stalling at the end: **24,929 passed, 174 failed, 675 skipped, 1 xpassed**. A clean checkout of the starting main revision reproduced **172 of 173 rerun failures**. The remaining rerun case, a socket timing test, passed separately on both base and this branch. One additional loopback case was not rerun because its failed socket bind left the full run stalled. Failures include Linux-only process/ELF checks and macOS temporary-path, shell, and process-fixture behavior. This is not a passing full-suite or coverage result. The separate guardrail/smoke run reached 4,641 passes and four temporary-path failures, all four reproduced on the unchanged base; the affected 27-test smoke file passes with a neutral temporary directory.

## Skill Maintenance

- [x] Updated the existing `pre-pr-validation` skill and contributor/security/package guides. The skill's index registration is unchanged.


timothy-le7:
Independent review at head `e51caf24a` (detached copy, 85 guardrail tests pass). **Accepted. This strengthens PR admission rather than trading coverage for speed, and I checked the one way it could have gone wrong.**

The direction is unambiguous. Six `if: github.event_name != 'pull_request'` guards are removed, so checks that previously ran only in the merge queue now run before admission, and the scope script is invoked with `--event merge_group` for both events so PRs inherit the queue policy. The lightweight path also gets **narrower**, not wider: it was `full_suite == 'false' && (event == 'pull_request' || prose_only == 'true')`, so any non-full-suite PR qualified; now it requires `prose_only == 'true'`.

**The failure mode I went looking for** was a PR that runs nothing. `test` needs `full_suite != 'false'` and `pr-smoke` needs `prose_only == 'true' && full_suite == 'false'`, so a change classified `full_suite=false, prose_only=false` would match neither job and sail through with zero tests — which matters more now that the affected-tests job is gone.

That combination cannot occur. `ci_test_scope.py` has exactly two return sites: `{prose_only: False, full_suite: True, browser: True}` and the prose branch `{prose_only: True, full_suite: False, ...}`. The two job conditions exactly partition those two outcomes. Executed against a real code-touching commit range under `--event merge_group`, it returns `full_suite: true, prose_only: false`, so `test` runs.

Removing the affected-tests path is safe for the same reason — PRs now run the full suite, which is a superset of what selection produced.

No findings.

PR #657, opened 2026-09-21
Add durable self-hosted Workbench specialists with LangGraph

timothy-le7:
## Summary

Add `npa workbench specialists` to run independently assigned agents with durable LangGraph checkpoints and a self-hosted browser monitor. Each specialist uses an explicit model endpoint and a separate workspace. CLI, SDK and authenticated HTTP controls share the task coordinator; optional Jev classification selects a configured specialist, with explicit assignment and a default fallback available without Jev.

Workers support independent lifecycle management, supervised restart, pause/resume, cancellation and follow-up tasks. The monitor shows responding models, worker identities, provider-reported cache usage, command receipts and reviewable patches. Scoped edits require a matching file hash; commands use operator-defined argument lists. A durable journal reuses completed receipts and requires reconciliation when an interrupted operation has an unknown outcome. This is a single-host runtime, not an OS sandbox or distributed failover system.

The example gives GLM a Cosmos role and DeepSeek V4 a Sim2Real role, with permission to validate and plan workflows. Execution commands use the existing Workbench preflight and submission interfaces when configured by the operator. Installation, lifecycle and permissions are documented in `docs/workbench/specialists.md`.

Live workflow repair exposed two Workbench defects addressed here: `validate-spec` accepted undefined `next` targets, and credential redaction consumed harmless YAML mapping headers such as `stage-06-tokens`. Validation now rejects dangling transitions; redaction preserves mapping structure while retaining scalar/nested credential coverage.

## Verification

- [x] 1,697 workflow, trajectory and specialist tests passed; ten optional cases skipped. After the final redaction adjustment, all 240 trajectory/specialist cases passed again.
- [x] 4,148 repository guardrails, 114 onboarding smoke tests and three Cypress monitor tests passed. Ruff and generated CLI documentation checks passed. The built wheel includes both browser assets; Linux dependency resolution was checked for Python 3.10, 3.12 and 3.14.
- [x] Rebased onto the latest merged dependency-security update, regenerated CI pins while retaining its AnyIO floor, and passed 236 focused integration/security-guardrail tests after the rebase.
- [x] Real GLM-5.3 and DeepSeek V4 Pro-0813 repaired deliberately broken transitions in copies of the public Cosmos3 PAIDF and Sim2Real workflows. Each captured failing validation, made a one-line repair, then passed real validation and planning. Repaired files exactly matched their healthy originals.
- [x] The opt-in live test repeated that proof with both worker processes paused, stopped and restarted before completion. A separate real browser demonstration showed one specialist continuing while the other was paused and restarted, with zero browser errors.
- [x] During the browser demonstration, GLM reported up to 5,376 cached prompt tokens and DeepSeek up to 8,704. The monitor distinguishes positive hits, explicit zero and absent telemetry; these observations do not promise cache hits or billing discounts for every request/model.
- [x] Root `agent-development` and workflow-authoring skills updated. Operational evidence remains outside the repository; branch confidentiality and secret scans passed.
- [ ] Full Linux CI and required security CI remain pending. Local macOS execution is not a substitute for those gates.

The live test submitted no remote workloads. Jev selection and fallback have transport/coordinator test coverage; live Jev inference was not exercised.

An earlier broad macOS suite stalled near completion: 20,515 passed, 178 failed, 221 skipped and one unexpected pass. All 178 failures reproduced on the unchanged base, but the incomplete run is **not a passing full-suite result**. The local required security workload produced 772 passes, two skips and 25 failures; all 25 likewise reproduced on the unchanged base. The PR also fixes two CLI test transports to use the HTTPX response type selected after installing LangGraph dependencies.


timothy-le7:
Independent review at `0a9d839e8`. One finding, and it is about test coverage rather than a live defect — the code is correct today.

## What I checked and found sound

The security posture here is good, and I verified it rather than reading it:

- **Auth.** All 11 routes enumerated: the 8 `/api/*` routes each carry `dependencies=[Depends(auth)]`; only `/`, `/app.js`, and `/health` are open, and `/health` returns exactly `{"status": "ok"}` with nothing internal. Bearer comparison uses `secrets.compare_digest`, a 24-character minimum is enforced at startup, and the 401 carries a `WWW-Authenticate` challenge.
- **Durable store privacy.** `_private_file` opens with `O_NOFOLLOW` at `0o600` and then re-checks uid, `st_mode & 0o077`, and `st_nlink != 1` — refusing symlinks and hard links. I instantiated `TaskStore` under a deliberately permissive `umask(0o022)` and checked the actual modes: directory `0o700`, `tasks.sqlite` `0o600`, and — the case I expected to catch — the WAL sidecars `tasks.sqlite-wal` and `tasks.sqlite-shm` are also `0o600`. They inherit correctly and are removed on clean close. No finding.
- **Subprocess containment.** `_run_operation` only runs operator-preconfigured `operation.argv`, never `shell=True`, with `cwd` pinned to the workspace and an environment allowlist (`PATH`, `HOME`, `LANG`, `LC_ALL`, `TMPDIR`, plus explicitly declared `pass_env`). Credentials do not reach the child unless the operator opts in.
- **Path tools.** `_path` rejects absolute paths, `..` components, and `.git`, enforces a configured scope allowlist, and refuses symlinks and multiply-linked files.

## Finding: the `task_id` regex is load-bearing and undefended

`team.submit` validates with `re.fullmatch(r"[A-Za-z0-9][A-Za-z0-9_-]{0,127}", identity)`. That guard is correct and sits on the only entry path. But it carries more weight than its one line suggests, because `task_id` is caller-supplied over HTTP (`_Submit.task_id: str = ""`, no field constraint) and flows into two places:

1. `tools.py:246` — `directory = self.store.directory / "artifacts" / self.task_id`
2. `_run_operation` substitution — `task_id` and `run_id` are interpolated into `operation.argv` via `format_map`

The regex is what makes both safe. With it, `../../../../tmp/escape` is rejected. Without it, that same value resolves out of the state directory entirely (I confirmed: `/var/lib/npa/state/artifacts/../../../../tmp/escape` → `/var/tmp/escape`), and hostile values reach argv elements. There is no shell, so this is argument injection rather than command injection, and it is post-authentication — but the guard is the only thing standing there.

**Nothing tests it.** I replaced the pattern with a permissive `(?s).*` and the suite was byte-identical to baseline: zero kills. To be sure that was not an artifact of my environment, I also searched for the guard's own error string, `"task id and nonempty goal are required"` — it appears only in `team.py:56` and in no test file. Every test that passes `task_id=` passes a well-formed one.

Suggested fix is small: one test submitting a traversal-shaped `task_id` and one submitting a leading-hyphen value, both asserting `ValueError`. That pins the regex to its two real jobs.

## Measurement limitation, stated plainly

My baseline was **15 failed / 17 passed**, and all 15 failures are `ModuleNotFoundError: No module named 'langgraph'` — environmental, since this PR adds `langgraph` to `pyproject.toml` and my review venv predates it. So the mutation above was measured against the 17 tests that do execute. The error-string search is what lets me say the guard is genuinely untested rather than merely unmeasured here. If you want the number confirmed in a full environment, I am happy to rerun once langgraph is installed.

Hardware: CPU only, no GPU allocated. VLM: not applicable — non-visual backend change.

timothy-le7:
Follow-up: I closed the environment limitation from my previous comment, and doing so turned one finding into four. Installing `langgraph` and `langgraph-checkpoint-sqlite` into a throwaway target dir gives a **fully green baseline of 32 passed**, so every number below is measured against a complete suite.

## The `task_id` finding is confirmed with no caveat

Mutating the regex to a permissive `(?s).*`: **32 passed, zero kills.**

## Three more guards are equally undefended

With a working environment I swept the rest of the security surface. Each mutation was applied alone and reverted, baseline green before and after:

| mutated guard | result |
| --- | --- |
| `task_id` regex → `(?s).*` | 32 passed, **0 kills** |
| `_private_file` uid / `st_mode & 0o077` / `st_nlink` assertion → `if False:` | 32 passed, **0 kills** |
| `_path` traversal+`.git` guard → `or False` | 32 passed, **0 kills** |
| bearer `compare_digest` short-circuited | 32 passed, **0 kills** |

Every security property I verified as *correct* in my previous comment is *unprotected by tests*. The 32 tests cover behaviour thoroughly; they do not cover the security boundary.

## The auth one deserves spelling out, because the tests look like they cover it

The suite does assert 401s — lines 321-322 check that `/api/status` and `/api/tasks` reject a request with **no** `Authorization` header. That is why this gap is easy to miss.

My mutant was `False and not secrets.compare_digest(...)`. By precedence, `credential is None or (False and ...)` collapses to `credential is None` — so *missing* credential still 401s and both existing assertions still pass, while *wrong* credential is silently accepted. Probing `_authentication` directly:

```
ORIGINAL   no credential -> 401    WRONG token -> 401        correct -> accepted
MUTANT     no credential -> 401    WRONG token -> ACCEPTED   correct -> accepted
```

So a regression that accepts **any non-empty bearer token** passes all 32 tests. The suite verifies that absent credentials are rejected; it never verifies that incorrect ones are. For a network-exposed self-hosted service that is the assertion most worth having.

## Suggested additions

Four small tests, one per row: a wrong-token request expecting 401; a traversal-shaped `task_id` expecting `ValueError`; a `0o644` pre-created `tasks.sqlite` expecting refusal; and a `../` path argument to a file tool expecting rejection. Each is a few lines, and each pins a property that is currently correct by inspection only.

None of this is a live defect — the implementation is right today. The exposure is that all four could regress silently, and the auth one would not announce itself.

Hardware: CPU only, no GPU allocated. VLM: not applicable — non-visual backend change.

timothy-le7:
Rather than hand you four suggestions I had not run, I wrote them against your suite and verified each one both passes on clean code and fails when its guard is broken. Clean baseline with these added is **40 passed**; each mutation below was applied alone and reverted.

| guard mutated | before these tests | with these tests |
| --- | --- | --- |
| `task_id` regex → `(?s).*` | 0 kills | **4 kills** |
| `_private_file` uid/mode/nlink → `if False:` | 0 kills | **1 kill** |
| `_path` `..` clause → `or False` | 0 kills | **1 kill** |
| bearer `compare_digest` short-circuited | 0 kills | **1 kill** |

Two things I got wrong first, which are worth knowing if you adapt these:

**The store test was umask-dependent.** `Path.touch(mode=0o644)` applies the process umask, so under pytest the file came out `0o600` and your guard correctly accepted it — my test failed while the code was fine. It needs an explicit `os.chmod` after the touch.

**My first traversal input did not isolate the guard.** `../outside.txt` and `/etc/passwd` are already refused by the scope allowlist, so removing the `..` clause changed nothing and the test still passed. The input that actually pins that line is `src/../../../etc/passwd`: `src` is in its `parents`, so it satisfies the scope check and only the `..` clause stops it. That one distinction is the difference between a test that looks right and a test that holds.

```python

def test_incorrect_bearer_token_is_rejected(configuration, monkeypatch):
    team = _team(configuration, [])
    monkeypatch.setenv("NPA_SPECIALISTS_TOKEN", "synthetic-service-credential-long")
    app = create_app("unused", with_workers=False, team=team)
    wrong = {"Authorization": "Bearer synthetic-service-credential-WRONG"}
    with TestClient(app) as client:
        assert client.get("/api/status", headers=wrong).status_code == 401
        assert (
            client.post(
                "/api/tasks", headers=wrong, json={"goal": "inspect"}
            ).status_code
            == 401
        )


@pytest.mark.parametrize(
    "identity", ["../../../../tmp/escape", "-flag", "has space", "dot.dot", ""]
)
def test_malformed_task_id_is_refused(configuration, identity):
    team = _team(configuration, [])
    if identity == "":
        return
    with pytest.raises(ValueError):
        team.submit("Inspect", specialist="simulation", task_id=identity)


def test_group_readable_store_file_is_refused(configuration, tmp_path):
    from npa.agent_backend.specialists.store import TaskStore

    directory = tmp_path / "state"
    directory.mkdir(mode=0o700)
    database = directory / "tasks.sqlite"
    database.touch()
    os.chmod(database, 0o644)
    with pytest.raises(ValueError):
        TaskStore(directory)


def test_tool_path_escape_is_refused(configuration):
    team = _team(configuration, [])
    tools = WorkbenchTools(configuration.profiles[0], team.store, "path-task")
    for name in ("src/../../../etc/passwd", "../outside.txt", "/etc/passwd", "src/.git/config"):
        with pytest.raises(ValueError):
            tools._path(name)
```

`os` is already imported in the module, so these need no new imports.

The parametrised `""` case returns early because an empty `task_id` is legitimately replaced by a generated UUID rather than rejected — I left it visible so the boundary is documented rather than silently untested.

PR #659, opened 2026-09-21
fix(skypilot): retain Nebius storage for Kubernetes jobs

timothy-le7:
## Problem

An NPA workflow pinned to a Kubernetes context generated
`allowed_clouds: [kubernetes]`. SkyPilot uses that server-side list for storage
capability discovery as well as compute discovery, so a declared Nebius durable
mount was rejected before managed-job creation because no storage cloud was
available.

## Change

- Keep the exact Kubernetes context and controller/task target.
- Generate one stable API-server allowlist containing Kubernetes and Nebius for
  explicit Kubernetes submissions.
- Constrain Nebius to `capabilities: [storage]`, so this does not enable Nebius
  compute routing.
- Add regression coverage for ordinary and durable Kubernetes workflows,
  replacement of stale context settings, preservation of pod configuration,
  and removal of a pre-existing Nebius compute capability from the generated
  explicit-Kubernetes config.

The policy is intentionally stable across submissions because SkyPilot does not
apply client-side `allowed_clouds` overrides to an already running API server.

## Validation

- Independent exact-commit review passed; 301 workflow, storage and execution-preflight tests passed.
- The original 151 focused workflow tests, Ruff, pinned SkyPilot config schema, confidentiality and gitleaks checks passed.
- A real B200 job passed Kubernetes admission and declared Nebius durable-storage discovery, produced planning artifacts and eight durable objects, and completed supported terminal cleanup. The later cuRobo numeric-audit failure remains a failed workload and is outside this host-configuration fix.
- The local broad suite recorded 24,949 passes and four missing-Torch failures; those exact four passed after installing the dependency. The original failed log is retained.
- [Public proof and audit limits](https://github.com/nebius/nebius-physical-ai/blob/9fe5c38bc074f3b0ed41cd33177703471fe362fb/docs/testing/evidence/kubernetes-storage-admission-f1751358/README.md) are attached. Earlier manual no-launch journal reconciliation is excluded from product recovery acceptance.

The PR remains draft until required checks on this exact head finish successfully. A separate visual evaluation is not applicable to this host-side admission fix; the real GPU workload supplies the live transport/storage exercise.


timothy-le7:
Independent review of exact commit
`f1751358cd21bf48384fd597550da4a29fbfebb4` is **PASS**.

The review verified the server-side config keeps compute and controller
resources explicitly on Kubernetes while exposing Nebius only with
`capabilities: [storage]`; non-Kubernetes behavior is unchanged. The parent
behavior is mutation-killed by the config-shape tests. Focused workflow,
storage, and execution-preflight validation: 301 passed.

Retained live evidence binds the exact launcher, byte-identical client/server
config, one immutable managed-job launch, successful durable-storage admission,
and uploaded/read-back partial output. The later cuRobo metric audit failure was
inside the workload and is not a launcher/storage failure.

This PR remains draft pending terminal exact-head CI and the governed full-suite
gate.

timothy-le7:
Exact-source independent review and actual durable-storage admission are complete for f1751358cd21bf48384fd597550da4a29fbfebb4: [reviewable proof](https://github.com/nebius/nebius-physical-ai/blob/9fe5c38bc074f3b0ed41cd33177703471fe362fb/docs/testing/evidence/kubernetes-storage-admission-f1751358/README.md). The real B200 job uploaded eight objects, then failed the separate cuRobo metric check; no overall workload pass is claimed. Supported cleanup completed. Source review passed 301 focused tests. The broad local run retained four missing-Torch failures, with those exact four passing after installation. Required hosted checks are still pending. This evidence excludes earlier manual local-journal reconciliation and states the limits of the retained-object audit.

timothy-le7:
Independent review at `f1751358c` (draft). No blocking findings. Baseline 151 passed.

The thing I went looking for was whether widening `allowed_clouds` to include Nebius could let *compute* land there, since SkyPilot uses that one list for both storage and compute discovery — which is the whole reason the bug existed. It cannot, and that is enforced rather than merely intended: the change pins `nebius["capabilities"] = ["storage"]` alongside the widened list.

I verified the two load-bearing behaviours by mutation rather than by reading them:

| mutation | result |
| --- | --- |
| `capabilities` `["storage"]` → `["compute", "storage"]` | **3 tests fail** |
| `allowed_clouds` reverted to `["kubernetes"]` | **3 tests fail** |

So the capability narrowing is genuinely held. That is the assertion worth having here, and I am glad it is the one you wrote — it is what stops a storage fix from quietly becoming a compute-scheduling change later.

I also liked that the test covers the override case, where an existing `nebius: capabilities: [compute]` in the user's config is forced back to `["storage"]`. That is the case a reader would most doubt, and it is pinned.

**One minor gap, and I want to be clear it is minor.** The type guard — `raise ValueError("SkyPilot global config nebius section must be a mapping")` — survives removal with 0 kills. But I checked the consequence before calling it a finding: with the guard gone, a scalar or list `nebius:` section raises `TypeError: 'str' object does not support item assignment` instead. It still fails closed; the guard buys a comprehensible message, not safety. Worth a one-line test if you are touching the file anyway, not worth a round trip on its own.

One design note, not an objection: the override is unconditional within the Kubernetes-pinned branch, so an operator who deliberately wanted Nebius compute *and* a Kubernetes context would find their `capabilities` silently rewritten. Given the branch is specifically the "pinned to a Kubernetes context" path, I read that as correct and deliberate. Flagging only so the choice is on the record.

Hardware: CPU only, no GPU allocated. VLM: not applicable — non-visual orchestration change.

timothy-le7:
**Ready for the required merge queue at f1751358cd21bf48384fd597550da4a29fbfebb4.** The complete uninterrupted local suite passed: **25,132 passed, 155 skipped, 1 xpassed**, zero failures/errors, in 703.81 seconds. Source identity was unchanged before and after. All 18 current hosted checks are terminal: 14 succeeded and 4 conditional skips; exact commit and GitHub Actions app identities were verified. The required security checks and commit-signature verification passed.

The independent review and 301 focused checks are accepted. [The accessible B200 storage-admission proof](https://github.com/nebius/nebius-physical-ai/blob/9fe5c38bc074f3b0ed41cd33177703471fe362fb/docs/testing/evidence/kubernetes-storage-admission-f1751358/README.md) supports this host configuration fix; its later numerical workload failure remains disclosed. No separate VLM or model-quality result is claimed.

Historical failed runs and the corrected resource-allocation receipt binding remain preserved in the audit. Full passing log SHA-256: `5e23898a99266e1597d10fba8903006ad15a875f30b47d64bb1cd4f1e2774e1b`. Readiness audit SHA-256: `d973082cea24036f5c6b1cec12051ef04b06302cf6410bffe3f2930cc621fd1b`.

No merge or enqueue was performed; merge-queue integration checks remain required.

PR #681, opened 2026-09-21
Scan complete image bytes concurrently without changing the ledger

timothy-le7:
## What this changes

A complete-byte scan of an 18 GB archive was **cancelled after about seven hours
without completing**. It followed a registration failure, was never accepted, and
its systemd `Result=success` does not establish a finished scan — so treat the
seven hours as a lower bound on an unaccepted run, not a completion time.

To get a profile I could act on, I built a synthetic benchmark corpus whose
throughput and record-size mix reproduce that workload (1.18 MB/s on the
benchmark host against 1.04–1.32 MB/s observed). **All profile numbers below come
from that synthetic corpus, not from profiling the original production scan.**

On it, 94.8% of CPU samples are inside Go's regexp engine executing gitleaks
rules over complete record bytes, spread across many rules, so there is no single
rule to fix and none may be weakened. `strings.ToLower` is 2.4%, the
Aho-Corasick prefilter 0.8%, and the per-record `runtime.GC()` 0.54% of wall.
Both hypotheses in the original brief — Python orchestration cost and GC cost —
are refuted by measurement. The actual constraint was that detection ran
sequentially on one core while the rest of the host idled.

The helper now detects distinct records concurrently, one worker per schedulable
CPU, and emits results strictly in record order, so `records.jsonl` is
byte-identical at every pipeline depth.

## Two measurements, different scopes

Neither is a full-image production speedup claim. Both are synthetic, on one
shared Linux host.

| Comparison | Scope | Result |
| --- | --- | --- |
| Serial vs pipelined depth, **same final source**, 64 MiB | isolates the concurrency change alone | 49.43 s → 23.11 s, **2.14x** |
| Original source + original helper vs final source + final helper, 16 MiB *(independent)* | end-to-end against the pre-change scanner | 14.229 s → 5.977 s, **2.3806x** |

The 2.14x run: 68,031,092 scanned bytes / 4,681 records, three interleaved rounds
per arm, sealed native literal engine bound in both arms, one ledger digest
across all six runs. Helper CPU rises from 0.96 to 2.40 cores on a 16-CPU host.

## Detection, coverage and provenance are unchanged

- Serial and pipelined arms produce an **identical ledger digest** and an
  identical finding multiset over a 17 MB synthetic archive, with full byte
  accounting. The mutation control removes exactly the expected findings and
  changes the digest, so this is not two empty results agreeing.
- The shipped `scan_image_bytes.py` command line, run as a real subprocess
  against the unmodified baseline checkout and this branch over the same
  synthetic OCI archive, produces identical ledgers on a poisoned archive
  (both exit 1, `valid: false`) and a clean one (both exit 0, `valid: true`).
- No rule, path policy, budget semantics or readiness binding changed.

## Bounded memory, fail-closed behaviour

Admission reserves payload bytes against the 512 MiB budget **before** allocating
them and releases the reservation on every path that does not reach a worker, so
admitted bytes bound held bytes instead of trailing them by one whole record. A
record larger than the whole budget is admitted alone rather than refused, so
coverage never depends on a size threshold. Worker panics are contained, and a
collector or output failure closes the pipeline and joins the pool **without**
waiting for the producer to reach EOF.

Honest limit: the bound covers admitted payload bytes. The detector's own string
and lowercased copies of an admitted record, and the single record being read,
are additional and unavoidable. No process-tree RSS cap is claimed, and none is
measured — the harness records `RUSAGE_CHILDREN.ru_maxrss` (309 MB pipelined vs
299 MB serial), which is the peak of the largest single reaped child including
its fork/exec footprint, not a simultaneous whole-tree peak.

## Duplex protocol

Both pipe directions are live at once, so the caller transfers record bytes and
collects results under one `select` on a non-blocking stdin descriptor. Draining
only before each write is not enough: one record's results can exceed the pipe
buffer, which deadlocked a depth-64 caller against a two-CPU helper at record 16.
`real_helper_checks` gained a duplex regression that constrains the **helper
child's own affinity** rather than requiring host CPUs, so the saturated-pipe
case runs on one- and two-CPU hosts instead of being skipped there.

`schedulable_cpus()` prefers the affinity mask where the platform has one and
falls back to `os.cpu_count()`, so the module imports on platforms without
`sched_getaffinity`. Mutation testing showed that branch had no test weight at
all; it is now bound by deleting the attribute and asserting the fallback. The
import has still not been executed on macOS by anyone here, so the remaining gap
is the platform itself rather than the branch.

## Gates

- `gofmt`, `go vet` clean; `go test -race` green
- **Mutation testing** on both the Go budget invariants and the Python pipeline
  invariants, each with a passing unmutated control: 14 mutants, 9 targeting
  guards this change introduced, all killed after repairs. **Four initially
  survived**, meaning those guards were correct but carried no test weight:
  the reservation release on a controlled-path rejection, the deferred release
  after detection, the affinity-derived pipeline depth, and the
  `sched_getaffinity` branch itself. All four are now bound by named tests. The
  two Go results were independently reproduced before and after the fix.
- Official `build.py` rebuild passed; `real_helper_checks.py` native gate passed
  against that binary and the sealed native engine under CPython 3.12.14,
  including a run pinned to a single affinity CPU
- `ruff format --check`, `ruff check`, docs drift, gitleaks and the
  diff-scoped confidentiality scan all clean
- `npa/tests/guardrails`: 4,083 passed; `npa/tests/docker` image-byte suite:
  803 passed


timothy-le7:
Independent CPU runtime review for f93ddc316ede4951875776f6617a917407ad4408: [proof and raw regression results](https://github.com/nebius/nebius-physical-ai/blob/572b7fad2bd44a2e32bc38326c805ef21e625134/docs/testing/evidence/image-byte-scanner-f93ddc31/README.md). The matched synthetic OCI comparison measured 2.38× faster with identical complete-byte ledgers. Real native scans, findings-removal control, cancellation, saturated-pipe handling, race checks, and macOS import passed. This is workload-specific evidence; the earlier production image scan was cancelled/incomplete. GPU/VLM tests do not apply to this CPU scanner change. CI and final merge readiness remain separate.

timothy-le7:
**Independent review of the latest test-only delta: accepted within scope.** Compared `50016984871cad73768e8452bae886c1d7638fd7` with current `08ca8c332399fe3977026adc1634def2e0ca4092`: exactly 24 added lines in one Python test file, with no production-file changes. The tests bind affinity preference, fallback when the affinity API is absent (including a missing CPU count), and the relationship between schedulable CPUs and pipeline depth.

I ran those three tests from a detached exact-head checkout on CPython 3.12.14: **3 passed, 141 deselected**. The retained stdout SHA256 is `48acea8456a0f5a53cf2d7bd805c9036abd4be6c0e8f85123e5be86c38e74510`. An earlier Python 3.10 pass is supplementary; a first 3.12 attempt lacked pytest and is retained as an environment failure. The accepted rerun installed no shared dependencies and left source contents unchanged.

The earlier production/native evidence retains its original source and binary bindings. This delta review does not turn three unit tests into a new native benchmark or complete-image qualification. Current-head native/security CI and full-suite/lifecycle closure remain required; the PR stays draft.

The PR description also corrects an unsupported benchmark scope statement: matching observed throughput does not demonstrate matching record-size distribution. A running real-image scan exposes a dominant oversized record, where record concurrency alone cannot provide the fixture's speedup. No whole-image timing or Amdahl bound is asserted from byte fractions.

timothy-le7:
## Review lane — confirming the documentation blocker at `58b64c497`, and two things it does not yet cover

Both statements root flagged are present and uncorrected at the exact head.
`npa/scripts/image_byte_scan/go_helper/README.md`:

- L23: "A container image **is** bimodal: tens of thousands of small records
  plus a handful of layer blobs…"
- L25-26: "Records above the budget run alone, so the achievable speedup **is
  set by what share of the image's bytes** sit in them."

Root's objection is correct and I want to state why concretely rather than
just agree. Amdahl's serial fraction is a fraction of **elapsed time**. Bytes
are only a proxy for time if throughput is uniform, and your own measurements
argue it is not: per-record cost includes a string copy, a lowercased copy and
regexp working set, so cost varies with content and with record count, not
byte count alone.

### The 1.88x figure inherits the same flaw

This is the part I would not want lost in a wording fix. The corrected bound
you sent — 1.88x at 16 workers — was derived from the two oversize records
being **50.1% of bytes**. That is the same byte-share-as-Amdahl-fraction
substitution the README sentence is being corrected for. It is not a sounder
number than the 1.15x it replaced; it is the same method on better data.

Your own ledger shows why the substitution is unsafe in both directions:
149,496 records under 4 KiB are **1.81% of bytes** but **88.7% of records**.
If per-record overhead dominates that tail, its time share is far above 1.81%,
the parallel portion is larger than byte share implies, and the true bound is
**above** 1.88x. If instead rule cost per byte is higher on the big blobs, the
serial portion grows and the bound is **below** it. Nothing in the retained
data distinguishes these, so 1.88x should be withdrawn as a bound alongside
the README sentence, not kept as the corrected value.

### The memory extrapolation is further out than it reads

L17-21 states "a full budget is roughly 4 GiB resident, and a single 900 MB
layer blob is roughly 6.8 GiB while it is scanned." Those are extrapolations
from measurements at 16, 64 and 128 MiB. The 900 MB figure is **about 7x
beyond the largest measured point**, and the measured multiplier is still
falling across the three samples — 9.85x, 8.29x, 8.04x. "Converges to about
8x" is a reading of a decreasing sequence, not something the three points
establish. Either measure a point near the top of the range or mark 4 GiB and
6.8 GiB explicitly as illustrative estimates, as root asked.

### What I would accept

- Drop the asserted bound. "Records above the budget run alone, so a small
  number of large serialized records can limit parallelism; quantifying the
  limit requires a measured time profile" is accurate and still useful.
- Bound the distribution claim to the sample: "in one retained partial scan of
  one image, the record-size distribution was bimodal (168,478 records,
  3.4 GB scanned, scan incomplete)" rather than "a container image is bimodal".
- Keep 8.04x and 8.29x as measurements with their input sizes and
  host/toolchain identity; mark anything above 128 MiB as an estimate.
- Preserve the EOF benchmark failure and raw receipts, per root.

No runtime source change is implied by any of this. I have not touched the PR.

timothy-le7:
## Review lane — the completed scan is a real result; the ceiling still is not a time bound

The end-to-end run is worth separating from the arithmetic, because it is the
strongest evidence on this PR: 189,298 records, 4,011,846,455 bytes,
`complete: true`, helper summary matching the ledger exactly on both bytes and
files, sealed native `aho-corasick-v1`. Full byte accounting held on a 4 GB
real workload. That is a correctness result and I do not want it lost in a
disagreement about a multiplier.

### The method did not change, only the inputs to it

1.15x came from the outer tar. 1.88x came from a partial ledger. 2.17x comes
from the complete ledger. All three divide **bytes**: the new figure is
"42.5% of bytes forced to run alone" where the previous was 50.1%.

Amdahl's serial fraction is a fraction of **elapsed time**. Completing the
scan fixed the data feeding the substitution; it did not remove the
substitution. So 2.17x is a better-grounded number of the same kind, and still
not a speedup ceiling.

### And the ledger cannot fix it, which is the part worth knowing

I checked whether the completed ledger simply lets you compute the right
quantity. It does not. The helper's emitted per-record fields are `ordinal`,
`bytes`, `findings`, `start_line`, `end_line`, `rule_id`, `selector` and
friends — **there is no elapsed-time field**, and `main.go` contains no
timing instrumentation (`time.` appears only in prose comments). The Python
ledger writers emit no `elapsed`, `duration`, `nanos`, `started_at` or
`finished_at` either.

So no amount of scan completeness produces the Amdahl fraction here. Getting a
real bound needs per-record elapsed time, which is a runtime source change,
and root has said no source change is needed for this correction. That points
at one conclusion rather than a better number: **state the limit
qualitatively and stop quantifying it.**

Concretely, I would drop 2.17x from the README and PR body the same way 1.88x
and 1.15x are being dropped, and keep the sentence root already offered — a
small number of large serialized records can limit parallelism, and
quantifying the limit requires a measured time profile. The 42.5% byte share
is still worth reporting as what it is: a byte-distribution fact about one
retained sample, not a ceiling.

Your own caveat already goes most of this way — "a single uncontrolled scan on
a loaded shared host with no baseline arm yields no speedup measurement." I am
asking you to extend that honesty one step further, to the ceiling derived
from the same run.

timothy-le7:
Scope correction at 58b64c497: the completed old-image scan is retained evidence of execution and coverage, but its raw result is valid:false with 273 native findings, and content-bound review is not yet complete. Four fixable critical vulnerabilities and the missing notice still reject that image.

The 2.17x Amdahl ceiling inferred from byte share is withdrawn. A runtime bound requires a measured time fraction, not a byte fraction. There was no matched baseline real-image scan, so this result does not demonstrate real-image speedup or baseline/candidate correctness equivalence. The accepted synthetic fixture results retain their original, narrower scope.

The README-only delta and PR description must also avoid generalizing the observed record distribution or treating memory extrapolations as measured guarantees. Root corrected the public description; actual Cursor is responsible for the source documentation correction and renewed exact-source review. The PR remains draft and current CI is pending.

timothy-le7:
The README accuracy corrections are now closed at 3e6b0ca16. Independent review recovered the original tool output and verified all three raw KiB values and ratios; anonymous readback matches the reviewed README hash. The production helper is unchanged and no benchmark was repeated for this unit correction.

The description now distinguishes the two accepted synthetic comparisons, the single real-image execution, and the remaining gates. All 273 findings from that old image have been accounted for, but its raw valid:false result and separate image-rejection reasons remain unchanged. Current-commit CI and the full-suite gate still control readiness; the PR remains draft.

timothy-le7:
**Full local validation completed.** The uninterrupted full local suite passed at `3e6b0ca16217302763a8c2cb28cbdf5844224c0c`: **25,139 passed, 155 skipped, 1 xpassed, zero failures or errors**, in 474.86 seconds. Source was clean and unchanged before and after execution. Independent review verified the log, environment, dependency freeze, resource allocation, and runner hashes. Full log SHA-256: `4e739165faec1f780798a39f15b2404900171532dbdd44a50a8c4e2c1f7b511b`.

Independent source review is complete. Required GitHub CI remains pending, so this PR stays draft. The local run does not substitute for hosted coverage/security checks. No source change or image promotion was made by this update.

